PR #2, opened 2026-04-03
feat: v1 database schema — team, api_key, sandbox, snapshot, activity

pavitrabhalla:
## Summary
Fresh v1 database schema. 5 tables, goose migration with UP/DOWN, sqlc queries.

- `team` — org-level grouping
- `api_key` — scoped auth keys per team
- `sandbox` — core resource, `sandbox_status` enum (starting, active, pausing, idle, deleted)
- `snapshot` — point-in-time snapshots, saved/unsaved visibility for dashboard
- `activity` — unified event log (lifecycle, exec, file ops) with filterable columns

All singular table names. All queries in SQL (sqlc generated).

Closes sa-2xd


db/migrations/004_control_plane_tables.sql
```diff
@@ -0,0 +1,113 @@
+-- 004_control_plane_tables.sql
+-- Control plane v1 schema: team, api_key (v2), sandbox, snapshot, activity.
```
pavitrabhalla: why does this have v2 here? We are pre-launch. Let's keep versions as v1.

sqlc.yaml
```diff
@@ -16,3 +16,6 @@ sql:
             go_type: "github.com/google/uuid.UUID"
           - db_type: "timestamptz"
             go_type: "time.Time"
+        rename:
```
pavitrabhalla: I don't think we need to migrate any legacy api keys. this break is acceptable

PR #20, opened 2026-04-06
feat: sandbox network security — nftables, egress proxy, rate limiting

meAmitPatil:
## Summary

Comprehensive sandbox network security and API hardening:

- **nftables firewall** — replaced all iptables with nftables (both namespace and host level). Per-sandbox egress filtering with 4 IP sets (predefined allow/deny + user allow/deny), ESTABLISHED/RELATED conntrack, private range blocking, MSS clamping, atomic rule replacement
- **TCP egress proxy** — per-sandbox userspace proxy with HTTP Host header and TLS SNI inspection for domain-based filtering. DNS rebinding protection (blocks hostnames resolving to private IPs)
- **Per-sandbox connection limiter** — application-level atomic counters for TCP connection limits
- **Network config API** — `PUT /sandboxes/:id/network` for runtime egress rule updates. Also supported at creation time via `POST /sandboxes` with `network` field. Rules persisted in DB (`network_config` jsonb column) to survive pause/resume
- **Synchronous sandbox creation** — `POST /sandboxes` now blocks until VM is running, returns `status: active`. No more client polling
- **Simplified create API** — removed `vcpu_count` and `memory_mib` from create request. Resources come from the template (1 vCPU, 512 MiB)
- **Rate limiting** — in-memory per-IP token bucket (20 req/s, 40 burst) with standard headers (`RateLimit-Limit`, `RateLimit-Remaining`, `Retry-After`)
- **Security headers** — `X-Content-Type-Options: nosniff`, `X-Frame-Options: DENY`, `Strict-Transport-Security`
- **Proto + gRPC** — `SandboxNetworkEgressConfig`, `UpdateSandboxNetwork` RPC
- **OpenAPI spec** — updated with `NetworkConfig` schema and `PUT /sandboxes/:id/network`
- **Template memory** — reduced default from 1024 MiB to 512 MiB
- **CD fixes** — Deploy VMD now injects updated boxd into rootfs. CD Migrate workflow has environment selector (staging/production)

## Tested on staging

- Security headers present on all responses ✓
- Rate limiting headers present and decrementing ✓
- Auth rejection (no key, bad key, empty key) ✓
- Sandbox create with just `{"name": "..."}` → active (synchronous) ✓
- Exec inside sandbox (`echo hello`) ✓
- Internet access from sandbox (`curl httpbin.org → 200`) ✓
- `pip install pandas numpy` inside sandbox ✓
- `pip install flask` + Flask test client inside sandbox ✓
- File upload + download ✓
- Network config at creation ✓
- Network update on running sandbox → 204 ✓
- Network validation (domain in deny_out → 400) ✓
- Network on missing sandbox → 404 ✓
- Network on paused sandbox → 409 ✓
- Invalid UUID → 400 ✓
- Pause / Resume working ✓
- Smoke tests pass (10/10) ✓


pavitrabhalla (commented):
## Review

Comprehensive security hardening PR with good coverage. Several issues need attention before merging — two are critical/blocking, the rest are high/medium.

---

### CRITICAL — Blocking

**1. EgressProxy never started — TCP domain filtering is entirely non-functional**

`cmd/vmd/main.go` never instantiates `EgressProxy`, never calls `SetEgressProxy`, and never calls `Start()`. The gRPC adapter has a nil-guard (`if a.mgr.egressProxy != nil`) that silently no-ops. The nftables rules redirect port 80/443 to proxy ports 19080/19443, but nothing listens there — result is all HTTP/HTTPS traffic gets connection-refused. The `allow_out` domain feature is silently broken end-to-end.

**2. Network config NOT reapplied on sandbox resume**

`ResumeSandbox` calls `VMD.ResumeInstance` but never reads `sandbox.NetworkConfig` from the DB and never calls `VMD.UpdateSandboxNetwork`. After a pause/resume cycle, the nftables firewall is in a fresh namespace — all per-sandbox rules are gone. The PR description says "Rules persisted in DB to survive pause/resume" but the reapplication logic is missing.

---

### HIGH

**3. Wildcard `*` in `allow_out` bypasses all deny rules** (`proxy.go`)

`matchDomain` has `if strings.EqualFold(pattern, "*") { return true }` and domain checks run before CIDR deny checks. A request with `{"allow_out": ["*"], "deny_out": ["0.0.0.0/0"]}` allows all TCP egress through the proxy regardless of `deny_out`.

**4. `deny_out` entries not validated in `CreateSandbox`** (`handlers.go`)

`UpdateSandboxNetwork` validates that `deny_out` entries are valid CIDRs, but `CreateSandbox` does not. Non-CIDR strings in `deny_out` at creation time silently pass through to VMD where they're dropped via `continue`, leaving the deny rule absent.

**5. Rate limiter cleanup goroutine leaks** (`ratelimit.go`)

```go
go func() {
    ticker := time.NewTicker(cfg.CleanupInterval)
    for range ticker.C { limiter.cleanup(cfg.MaxAge) }
}()
```

No shutdown channel. `newRouter()` is called on each integration test — each call spawns a new goroutine that runs forever. Will cause test flakiness and memory growth.

**6. Rate limiting is per-IP, not per-API-key** (`ratelimit.go`)

The API is authenticated per API key, but rate limiting uses `c.ClientIP()`. Behind GCP's load balancer, `X-Forwarded-For` can collapse many customers to a single IP, making the limit effectively shared across all customers. Should be per-API-key (post-auth).

**7. DB writes after synchronous VMD call use request context** (`handlers.go`)

If the HTTP client disconnects during the (potentially long) synchronous VM boot, `c.Request.Context()` gets cancelled. All subsequent DB writes (`UpdateSandboxStatus(active)`, etc.) fail, leaving the sandbox stuck in `starting` state in the DB even though the VM is running. Use a background context for cleanup writes.

**8. `RemoveVM` in HostFirewall silently discards GetRules errors** (`hostfw.go`)

```go
fwdRules, _ := hfw.conn.GetRules(hfw.table, hfw.fwdChain)
natRules, _ := hfw.conn.GetRules(hfw.table, hfw.natChain)
```

If `GetRules` fails, rules won't be found and won't be deleted. The VM's FORWARD/NAT rules persist in the kernel permanently after the sandbox is destroyed — this is both a resource leak and a security issue.

---

### MEDIUM

**9. IPv6 traffic unfiltered** (`firewall.go`)

The table uses `TableFamilyINet` (IPv4+IPv6) but IP sets use `nftables.TypeIPAddr` (IPv4-only), and `cidrsToElements` explicitly skips IPv6 with `continue`. IPv6 egress from sandboxes bypasses all CIDR-based filtering.

**10. `addCIDRsToSet` drops more-specific CIDRs when `0.0.0.0/0` is present** (`firewall.go`)

```go
if slices.Contains(cidrs, AllTrafficCIDR) {
    // only adds the full-range element, discards the rest
}
```

`["8.8.8.8/32", "0.0.0.0/0"]` in `deny_out` would add only the all-IP range and silently drop `8.8.8.8/32`. The all-range subsumes it, so harmless for deny-all, but if mixed with other logic this gets fragile.

**11. `prefixEnd` integer overflow for `255.255.255.255/32`** (`firewall.go`)

```go
end := (ip | ^mask) + 1  // 0xFFFFFFFF + 1 overflows uint32 to 0
```

Creates a malformed nftables interval set element. May cause a kernel error or silently wrong rule.

**12. Silent breaking change: `vcpu_count`/`memory_mib` removed without 400 on unknown fields**

Clients sending `{"vcpu_count": 2}` get a 201 with their resource request silently dropped. Should return 400 if unrecognized fields are present, or at minimum document the removal prominently.

**13. Non-atomic rootfs mutation in `deploy-vmd.yml`**

Mount → copy boxd → umount is not atomic. A failure mid-sequence leaves the rootfs in an inconsistent state with no rollback. Consider writing to a staging copy and atomically swapping.

**14. `nsExecGo` namespace pollution on restore failure** (`nsexec.go`)

If `netns.Set(hostNS)` fails during the deferred namespace restore, the OS thread returns to the goroutine pool still pointing at the sandbox's network namespace. Subsequent goroutines scheduled on that thread run in the wrong namespace.

---

### Low / Nits

- `RateLimit-Reset` header always returns `"1"` instead of the actual token refill time (`ratelimit.go`)
- `netNamespacesDir` constant declared but never used (`nsexec.go`)
- `_ = ipAddress` comment is incorrect — `UpdateSandboxHost` is never called in `CreateSandbox` (`handlers.go`)
- `soOriginalDst = 80` is an undocumented magic number; prefer `syscall.SO_ORIGINAL_DST` where available (`proxy.go`)
- `deploy-api.yml` staging now always runs before production dispatch — intentional but undocumented behavior change

pavitrabhalla:
Just one feedback - can we use `PATCH /sandboxes/:id` for updating network rules on a created sandbox?

meAmitPatil:
> Just one feedback - can we use `PATCH /sandboxes/:id` for updating network rules on a created sandbox?

@pavitrabhalla pushed the change.

pavitrabhalla (approved):

PR #21, opened 2026-04-07
feat: sandbox API v2 — internet toggle, hard timeouts, immutable metadata

meAmitPatil:
## Summary

Three additive sandbox-create features plus a hardened reaper, fully tested on staging.

1. **`network` is the only egress control** — `allow_internet_access` was removed in favour of a single, unambiguous surface: `network.allow_out` / `network.deny_out`.

2. **`timeout_seconds` hard lifetime cap** — optional int on create, capped at 7 days. When the timeout elapses, the reaper **pauses** the sandbox (snapshot preserved, can be resumed). Already-idle sandboxes are skipped. Measured from `created_at`, not `last_activity_at`.

3. **Mutable metadata tags** — `map[string]string` on create, stored as `jsonb NOT NULL DEFAULT '{}'`, filterable via `?metadata.<key>=<value>` (AND semantics, jsonb `@>` containment). Caps: 64 keys, 256 B/key, 2 KB/value, 16 KB total. Reserved namespace: keys starting with `superserve.` or `_superserve` rejected. Composite GIN index on `(team_id, metadata)` via `btree_gin`. **Updatable after creation via `PATCH /sandboxes/:id`**.

4. **`timeout_seconds` and `network` in sandbox response** — both fields now returned on GET/list/create so callers can see what was configured.

5. **Hardened reaper with saga compensation** — replaced `ListExpiredSandboxes` + separate `DestroySandbox` with a single `ClaimExpiredSandboxes` CTE that atomically marks rows `pausing` with `FOR UPDATE SKIP LOCKED`. Multi-replica Cloud Run reapers no longer double-process. The pause is a distributed transaction across VMD + Postgres, so the reaper uses a saga:
   - VMD pause fails → revert DB to `active`, retry next tick.
   - Post-VMD DB write fails → resume the VM (compensation) and revert DB to `active`.
   - Compensation itself fails → mark `failed` with a single high-signal log line including `snapshot_path`, `mem_path`, original cause, and reason for on-call triage.

## Migrations

- `20260407000001_sandbox_timeout.sql` — adds `timeout_seconds int`, partial btree index for reaper
- `20260408000001_sandbox_metadata.sql` — adds `metadata jsonb NOT NULL DEFAULT '{}'`, `btree_gin` extension, composite GIN index, NULL→{} trigger

Note: staging DB already has both migrations applied.

## Test plan

- [x] `go build ./internal/api/... ./internal/db/... ./cmd/controlplane/...` — clean
- [x] `go test ./internal/api/...` — all tests green including reaper saga tests
- [x] `go test -tags=integration -c ./internal/integration/...` — compiles
- [x] Unit tests: `TestValidateMetadata`, `TestParseMetadataFilter`, `TestCreateSandbox_WithMetadata`, `TestCreateSandbox_EmptyMetadataIsObjectNotNull`, `TestCreateSandbox_MetadataValidationRejected`
- [x] Reaper tests: nothing-expired, VMD success (full pause flow), VMD fail (revert + loop continues), DB error, batch size, context cancel mid-batch, immediate-run-on-startup
- [x] Integration tests: metadata round-trip, empty defaults to `{}`, adversarial table, list filter AND semantics, team isolation, filter validation

### Staging verification

| Test | Result |
|---|---|
| Create with metadata + timeout_seconds | ✅ |
| GET returns timeout_seconds + metadata | ✅ |
| Metadata filter — single and multiple keys | ✅ |
| Unknown fields rejected | ✅ |
| timeout_seconds validation (0, >7 days) | ✅ |
| Reserved metadata key rejected | ✅ |
| PATCH metadata round-trip | ✅ |
| PATCH on idle sandbox → 409 (network only) | ✅ |
| PATCH network rules + GET shows them | ✅ |
| Pause + resume preserves network rules | ✅ |
| Reaper pauses sandbox after timeout | ✅ |


pavitrabhalla:
Two things I want to check here:

1. DB: composite GIN index with btree_gin extension required
The metadata filter uses metadata @> $filter with a composite GIN(team_id, metadata) index. This requires enabling the btree_gin Postgres extension. Confirm this extension is available in your Supabase project tier.

2. the snapshot leak and VMD memory-map gaps documented in SNAPSHOT_ARCHITECTURE.md are known, in-production, and explicitly punted. As sandboxes accumulate, NVMe disk on the bare-metal host will fill with orphaned snapshot files, and any VMD deploy will break resume for all paused sandboxes. These should be tracked as follow-up work before you scale up sandbox volume.

pavitrabhalla (approved):

PR #22, opened 2026-04-07
feat: edge proxy for direct sandbox access via subdomain routing

meAmitPatil:
## Why

Sandboxes need direct HTTP/WebSocket access for the things users actually run inside them: web apps (Next.js, Streamlit), file upload/download via boxd's `/files` endpoint, PTY and exec streams, and any custom service bound to a port. Routing all of this through the control plane would couple every user request to the API tier, add latency, and turn the control plane into a bottleneck for streaming workloads. The edge proxy gives each sandbox a unique subdomain (`{port}-{id}.sandbox.superserve.ai`) and forwards directly from bare metal to the VM, leaving the control plane out of the data path entirely.

## Summary

- Adds `cmd/proxy/` — new edge proxy binary routing `{port}-{instanceID}.sandbox.superserve.ai` directly to boxd inside the VM, bypassing the control plane on the hot path
- Adds `internal/proxy/` — host parser, TTL-cached VMD resolver, transport cache keyed by VM lifecycle, retry dial for boxd startup window
- Adds `internal/vm/local_http.go` — VMD exposes `GET 127.0.0.1:9090/instances/{id}` for proxy → instanceID resolution without gRPC
- Adds `deploy/proxy.service` — hardened systemd unit (`DynamicUser`, `NoNewPrivileges`, `ProtectSystem=strict`)
- Updates deploy workflow to build and deploy the proxy binary alongside VMD
- Adds `scripts/setup-proxy-infra.sh` — one-time GCP HTTPS LB + Certificate Manager wildcard cert setup

## Architecture

```
Client → GCP HTTPS LB (wildcard cert *.sandbox.superserve.ai)
       → proxy :5007 on bare metal
       → boxd :49983 on VM private IP
```

Host label `{port}-{instanceID}` is parsed, instanceID is resolved to VM IP via VMD local HTTP (cached 500ms), then reverse-proxied with a lifecycle-keyed transport (detects VM restarts via `StartedAt` nanoseconds).

## Security

- Domain suffix enforcement — rejects requests not matching `*.sandbox.superserve.ai`
- Instance ID charset validation — `^[a-zA-Z0-9][a-zA-Z0-9-]{0,63}$`
- Port blocklist — rejects ports < 1024 (blocks SSH, HTTP, HTTPS on the VM)
- All X-Forwarded headers stripped before forwarding
- Negative cache (1s TTL) for unknown instance IDs to absorb amplification
- Local HTTP bound strictly to `127.0.0.1`, enforced in code regardless of caller

## Key design decisions

- **Separate binary** — proxy crash doesn't take down VMD (VMD restart kills all sandboxes)
- **`Resolver` interface** — current `VMDResolver` (localhost HTTP, ~0.2ms cached), future `RedisResolver` is a one-line swap in `main.go`
- **Lifecycle-keyed transport** — stale TCP connections to restarted VMs are closed automatically
- **`FlushInterval: -1`** — streaming/PTY output flushed immediately, no buffering
- **Retry dial on `ECONNREFUSED` only** — handles boxd startup window without masking real errors

## Deployed & tested on staging

All tests run against `*.sandbox.superserve.ai` (GCP HTTPS LB + wildcard TLS cert, DNS live).

**Security validation:**
| Test | Expected | Result |
|------|----------|--------|
| Port 22 blocked | 403 | ✓ |
| Port 80 blocked | 403 | ✓ |
| Port 443 blocked | 403 | ✓ |
| Unknown sandbox | 404 | ✓ |
| Bad host format | 400 | ✓ |
| Health probe (mux-level) | 200 | ✓ |

**Production traffic:**
| Test | Result |
|------|--------|
| File read proxy | 200, correct content |
| Write + read round-trip | Data integrity confirmed |
| 1.4 MB upload / download | 0.46s / 0.33s |
| Connect-RPC streaming (5 lines × 100ms) | Chunks arrived incrementally, 401ms spread — `FlushInterval:-1` verified |
| 50 concurrent connect-rpc RPCs | 50/50 pass, 218ms total |
| 20 concurrent HTTP requests | 20/20 pass, 181ms total |
| Sandbox delete → proxy invalidates | 404 on next request |
| Avg round-trip latency | ~138ms (GCP LB + proxy hop) |


pavitrabhalla:
**Security Concerns**
1. No authentication on the proxy (probably a known issue)
Any request to {port}-{instanceID}.sandbox.superserve.ai is proxied directly to the sandbox with zero auth check. Instance IDs are UUIDs, which are hard to enumerate but not secret. If a sandbox ID leaks anywhere (logs, referrer headers, error messages, a customer’s code that accidentally shares it), anyone can access that sandbox’s open ports. There’s no mechanism to verify the requesting user owns the sandbox. For a product where sandboxes run arbitrary user code and may hold sensitive data or credentials, unauthenticated port access is a significant exposure.

2. Port blocklist instead of allowlist
Ports < 1024 are blocked. But any service bound on port ≥ 1024 inside the VM is accessible — not just user apps, but any internal tooling, debug endpoints, or unintended listeners. An allowlist (only boxd’s port 49983, plus ports explicitly opened by the user at sandbox create time) would be tighter. Worth discussing whether this threat model matches what’s actually running in VMs.

3. Cache eviction is a full nuke — creates a VMD amplification window
When the resolver cache hits maxCacheSize (10,000 entries), the entire cache is wiped with r.cache = make(map[string]cacheEntry). This means an attacker sending 10,001 requests with distinct (invalid) instance IDs flushes all cached negative entries, forcing the proxy to re-query VMD for all live sandboxes on the next request wave. The negative cache that protects VMD becomes worthless under this pattern. Should be an LRU eviction of the oldest entry, not a full wipe.

4. DisableKeepAlives: true makes IdleConnTimeout meaningless — and has a performance cost
The transport is configured with DisableKeepAlives: true, which closes the TCP connection after every HTTP request. IdleConnTimeout: 610s is then effectively dead config since connections are never idle — they’re closed immediately. More importantly, this means every short HTTP request to a sandbox pays a full TCP handshake. The lifecycle-keyed transport cache already handles the “stale connection on VM restart” problem correctly; keep-alives could be left enabled and the lifecycle key rotation closes old transports on restart. For streaming/WebSocket connections this doesn’t matter (they’re long-lived), but it’s unnecessary overhead for high-frequency HTTP.

5. Host header reflects the public sandbox URL into boxd
The proxy sets req.Host = r.Host (the sandbox subdomain) before forwarding. This is intentional to prevent VM IP leakage in redirects, but it also means user applications inside the VM receive the public sandbox URL as the Host header and may trust it for routing or CSRF decisions. If boxd or any user app does Host-based routing or Host validation, this is the actual attack surface. Lower severity but worth documenting for boxd authors.

6. /health endpoint bypasses all proxy validation
The mux intercepts /health before the proxy handler, so it responds 200 to any request at /health regardless of the Host header. This is intentional for GCP health probes, but it means the proxy will happily confirm it’s alive to anyone who hits it directly (e.g., by reaching the bare-metal IP). Not a serious issue but worth being explicit that this is a deliberate choice.

**Scalability Concerns**
1. Architecture is fundamentally single-host
The proxy queries 127.0.0.1:9090 (VMD on the same machine). VMD’s m.vms map only contains sandboxes running on that host. There’s no cross-host routing — a sandbox on host B is invisible to the proxy on host A. The Resolver interface exists precisely to allow a RedisResolver swap, but that work doesn’t exist yet. Until it does, adding a second bare-metal host means sandbox port access is broken for sandboxes not on the host the LB happens to route to.

2. The GCP LB backend is an unmanaged instance group
The setup script creates an instance-groups unmanaged — instances are added manually, not auto-scaled. Adding host 2 requires manually running gcloud compute instance-groups unmanaged add-instances. There’s no auto-scaling trigger or health-check-driven replacement. Fine for now, but needs to be part of the multi-host plan.

3. Transport sweeper holds the global mutex over all 10,000 entries every 5 minutes
The sweep() method acquires c.mu.Lock() and iterates the entire cache. During this window, every concurrent proxy request that needs a transport is blocked. With 10,000 active sandboxes this could cause a measurable latency spike every 5 minutes. Should be done in batches or use a finer-grained lock.

4. No per-sandbox or per-IP connection limits
A single client or a single sandbox can receive unlimited concurrent connections through the proxy. There’s rate limiting on the control plane API (from PR #20), but the proxy completely bypasses it. A sandbox that becomes a traffic target, or an abusive user who knows a valid sandbox ID, can exhaust file descriptors on the host (LimitNOFILE=65536 is the only cap, and it’s per-process, not per-sandbox).

5. Proxy is a single point of failure for all sandbox port traffic
A proxy crash or OOM makes all sandbox ports unreachable even though sandboxes themselves are fine. The PR correctly separates the proxy binary from VMD so crashes don’t kill sandboxes, but there’s no redundancy on the proxy tier itself (no second proxy process, no watchdog beyond systemd’s Restart=always with a 2s delay).

pavitrabhalla (commented):
## Review

Good PR overall — the core security invariants are all correctly in place: domain suffix check before routing, instance ID charset validation before it touches the VMD URL, privileged port block, and loopback-only binding of the VMD HTTP server. The threat model in the PR description matches the implementation. Tested on staging is a nice signal.

Seven inline comments posted. Summarizing by severity:

**Major**
- `resolver.go:82` — Cache stampede on miss: concurrent goroutines all call `fetch()` in parallel on a cache miss. Use `singleflight.Group` to collapse concurrent misses into one in-flight HTTP call to VMD.
- `proxy.go:344` — `clientAddr` is broken under GCP HTTPS LB. RemoteAddr is the LB's egress IP, not the client's — the per-IP limiter currently applies to everyone equally (i.e. it does nothing useful). Needs a decision: trust `X-Forwarded-For` with validation, or drop the per-IP limiter.

**Minor**
- `proxy.service:3` — Missing `After=vmd.service`. Proxy can start before VMD's local HTTP listener is ready, causing 503s on boot.
- `local_http.go:91` — Sub-paths under `/instances/` silently treated as instance IDs.
- `connlimit.go:30` — `release()` can go negative if misused by future callers; easy defensive fix.
- `deploy-vmd.yml` — 2-second `sleep` before `is-active` check may be too short for a slow proxy start.

**Nit**
- `resolver.go:148` — `evictOldest()` is O(N) under lock; random eviction is O(1) and sufficient.

The singleflight and clientAddr issues are the ones I'd address before merge.

pavitrabhalla (approved):
Approved. The last comment was posted from an AI review. Take a look before you merge.

PR #27, opened 2026-04-10
feat: environment variables at sandbox creation

meAmitPatil:
## Summary

- Accept `env_vars` map in `POST /sandboxes` request
- Pass through VMD gRPC → boxd's new `POST /init` endpoint
- boxd stores env vars in memory, injects into every process (terminal, exec)
- Per-request env vars in exec override sandbox-level ones
- Env vars survive pause/resume via VM snapshot

## Staging verification

- Create with `env_vars: {MY_VAR: hello, DEBUG: 1}` → exec `echo $MY_VAR $DEBUG` → `hello 1`
- Per-request override: exec with `env: {MY_VAR: overridden}` → `overridden`
- Pause → resume → exec → env vars still present
- Installed and ran Claude Code inside a sandbox with `ANTHROPIC_API_KEY` set via env vars

## Test plan

- [x] Unit tests pass
- [x] Create sandbox with `env_vars`, verify via exec
- [x] Per-request env vars override sandbox-level
- [x] Env vars persist across pause/resume
- [x] Claude Code works inside sandbox with API key set via env vars

### DIFF
--- a/proto/vmdpb/vmd.pb.go
+++ b/proto/vmdpb/vmd.pb.go
@@ -329,8 +329,9 @@ type CreateVMRequest struct {
 	KernelArgs     string                 `protobuf:"bytes,4,opt,name=kernel_args,json=kernelArgs,proto3" json:"kernel_args,omitempty"`               // Boot arguments for the kernel.
 	ResourceLimits *ResourceLimits        `protobuf:"bytes,5,opt,name=resource_limits,json=resourceLimits,proto3" json:"resource_limits,omitempty"`
 	NetworkConfig  *NetworkConfig         `protobuf:"bytes,6,opt,name=network_config,json=networkConfig,proto3" json:"network_config,omitempty"`
-	Metadata       map[string]string      `protobuf:"bytes,7,rep,name=metadata,proto3" json:"metadata,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"` // Arbitrary key-value pairs.
-	SandboxNetwork *SandboxNetworkConfig  `protobuf:"bytes,8,opt,name=sandbox_network,json=sandboxNetwork,proto3" json:"sandbox_network,omitempty"`                                         // Per-sandbox egress rules (optional).
+	Metadata       map[string]string      `protobuf:"bytes,7,rep,name=metadata,proto3" json:"metadata,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`              // Arbitrary key-value pairs.
+	SandboxNetwork *SandboxNetworkConfig  `protobuf:"bytes,8,opt,name=sandbox_network,json=sandboxNetwork,proto3" json:"sandbox_network,omitempty"`                                                      // Per-sandbox egress rules (optional).
+	EnvVars        map[string]string      `protobuf:"bytes,9,rep,name=env_vars,json=envVars,proto3" json:"env_vars,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"` // Environment variables injected into every process in the VM.
 	unknownFields  protoimpl.UnknownFields
 	sizeCache      protoimpl.SizeCache
 }
@@ -417,6 +418,13 @@ func (x *CreateVMRequest) GetMetadata() map[string]string {
 func (x *CreateVMRequest) GetSandboxNetwork() *SandboxNetworkConfig {
 	if x != nil {
 		return x.SandboxNetwork
+	}
+	return nil
+}
+
+func (x *CreateVMRequest) GetEnvVars() map[string]string {
+	if x != nil {
+		return x.EnvVars
 	}
 	return nil
 }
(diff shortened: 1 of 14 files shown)


pavitrabhalla (commented):
Good overall — clean layering through the stack and solid input validation. A few issues worth addressing:

**1. Silent env-var injection failure ()**

In both `CreateVM` and `ResumeVM`, the error from `postBoxdInit` is logged but swallowed:

```go
if err := postBoxdInit(ctx, inst.IP, req.GetEnvVars()); err != nil {
    a.mgr.log.Error().Err(err).Str("vm_id", inst.ID).Msg("failed to post env vars to boxd")
}
```

If the caller passed env vars and this call fails (network hiccup, boxd not ready yet, etc.), the sandbox starts successfully but processes run without the expected env vars. For use cases like injecting `ANTHROPIC_API_KEY`, this is a silent data loss that's very hard to debug. Two options:
- Return the error (fail-fast) — safest, especially on fresh `CreateVM`
- Or at minimum surface it in the API response so callers know the sandbox started but env vars were not set

**2. Key format not validated — null bytes and `=` signs allowed**

`validateEnvVars` checks length and count but not the key format. A key containing `=` (e.g. `FOO=BAR`) produces a double-`=` entry like `FOO=BAR=value` when concatenated in `environ()`, which is silently malformed. A key with a null byte could also cause `exec` to behave unexpectedly. Suggest adding:

```go
for _, c := range k {
    if c == '=' || c == 0 {
        return fmt.Errorf("env_vars key %q contains invalid character", k)
    }
}
```

POSIX recommends `[A-Z_][A-Z0-9_]*` but even just rejecting `=` and null would cover the dangerous cases.

**3. `handleInit` sets wrong Content-Type on error**

```go
http.Error(w, `{"error":"invalid JSON"}`, http.StatusBadRequest)
```

`http.Error` sets `Content-Type: text/plain`, so despite the JSON-shaped body, the header says plain text. Use `w.Header().Set("Content-Type", "application/json")` before writing the error, or just use a plain English message and let `http.Error` set the header correctly.

**Minor: new `http.Client` per `postBoxdInit` call**

`&http.Client{Timeout: 5 * time.Second}` is allocated on every call. Since this is called on every create/resume, a package-level or adapter-level client would be cleaner (and avoids any subtle connection-pool fragmentation).

pavitrabhalla (commented):
Good overall — clean layering through the stack and solid input validation. A few issues worth addressing:

**1. Silent env-var injection failure (`grpc_adapter.go`)**

In both `CreateVM` and `ResumeVM`, the error from `postBoxdInit` is logged but swallowed:

```go
if err := postBoxdInit(ctx, inst.IP, req.GetEnvVars()); err != nil {
    a.mgr.log.Error().Err(err).Str("vm_id", inst.ID).Msg("failed to post env vars to boxd")
}
```

If the caller passed env vars and this call fails (network hiccup, boxd not ready yet, etc.), the sandbox starts successfully but processes run without the expected env vars. For use cases like injecting `ANTHROPIC_API_KEY`, this is a silent data loss that's very hard to debug. Two options:
- Return the error (fail-fast) — safest, especially on fresh `CreateVM`
- Or at minimum surface it in the API response so callers know the sandbox started but env vars were not set

**2. Key format not validated — null bytes and `=` signs allowed**

`validateEnvVars` checks length and count but not the key format. A key containing `=` (e.g. `FOO=BAR`) produces a double-`=` entry like `FOO=BAR=value` when concatenated in `environ()`, which is silently malformed. A key with a null byte could also cause `exec` to behave unexpectedly. Suggest adding at minimum:

```go
for _, c := range k {
    if c == '=' || c == 0 {
        return fmt.Errorf("env_vars key %q contains invalid character", k)
    }
}
```

POSIX recommends `[A-Z_][A-Z0-9_]*` but even just rejecting `=` and null covers the dangerous cases.

**3. `handleInit` sets wrong Content-Type on error**

```go
http.Error(w, `{"error":"invalid JSON"}`, http.StatusBadRequest)
```

`http.Error` sets `Content-Type: text/plain`, so despite the JSON-shaped body the header says plain text. Either set the header explicitly before writing, or just use a plain English message and let `http.Error` handle the header.

**Minor: new `http.Client` per `postBoxdInit` call**

`&http.Client{Timeout: 5 * time.Second}` is allocated on every call. A package-level or adapter-level client would be cleaner.

proto/vmdpb/vmd.pb.go
```diff
@@ -1874,51 +1898,55 @@ var file_proto_vmd_proto_goTypes = []any{
...
+	18, // 34: superserve.vmd.v1.VMDaemon.ExecCommand:output_type -> superserve.vmd.v1.ExecCommandResponse
+	20, // 35: superserve.vmd.v1.VMDaemon.GetVMInfo:output_type -> superserve.vmd.v1.GetVMInfoResponse
+	22, // 36: superserve.vmd.v1.VMDaemon.SetupNetwork:output_type -> superserve.vmd.v1.SetupNetworkResponse
+	24, // 37: superserve.vmd.v1.VMDaemon.UpdateSandboxNetwork:output_type -> superserve.vmd.v1.UpdateSandboxNetworkResponse
+	28, // [28:38] is the sub-list for method output_type
+	18, // [18:28] is the sub-list for method input_type
+	18, // [18:18] is the sub-list for extension type_name
+	18, // [18:18] is the sub-list for extension extendee
+	0,  // [0:18] is the sub-list for field type_name
 }
```
pavitrabhalla, line 1950: @meAmitPatil Can you explain these mappings and how they work?

meAmitPatil: > @meAmitPatil Can you explain these mappings and how they work?

@pavitrabhalla they are auto-generated by protoc, these are protobuf's internal type indexes, regenerated with make generate-proto.

pavitrabhalla (approved):

PR #28, opened 2026-04-12
VM lifecycle rearchitecture: sandboxes survive VMD restarts

meAmitPatil:
## Summary

Implements the VM lifecycle rearchitecture from `docs/VM_LIFECYCLE_REARCHITECTURE.md`. Running sandboxes now survive VMD restarts, deploys, and crashes.

**Core change:** Firecracker VMs run as standalone systemd units (`firecracker@<id>.service`) instead of VMD child processes. VMD becomes a stateless supervisor that can restart without affecting VMs.

### What's in this PR

**Phase 0 — Data model**
- `host` table for multi-host scheduling
- `sandbox.host_id` NOT NULL with backfill
- `reconciler_log` audit table
- `snapshot.mem_path` column (replaces hardcoded convention)

**Phase 1 — Multi-host routing**
- `HostRegistry` resolves `host_id → VMDClient` via DB lookup + gRPC dial
- `LeastLoaded` scheduler with power-of-two-choices algorithm
- All handlers route VMD calls through `host_id`

**Phase 2 — Firecracker as systemd unit**
- `firecracker@.service` template unit — VMs outlive VMD
- BoltDB state store for durable VM metadata
- Startup reattach: reconstructs in-memory map from BoltDB + systemd (preserves paused VMs)
- Template snapshot caching: skips cold boot when base rootfs unchanged
- Async PID resolution (no hot-path dbus overhead)

**Phase 3 — Self-healing**
- Three-way reconciler (BoltDB ↔ systemd ↔ DB) with 5 drift kinds + audit log
- Rate-limited auto-fail (max 5/hour) + 60s grace period
- VMD detects dead VMs on connection error, cleans up and returns NotFound
- Handler degradation: 410 Gone instead of 500 on dead VMs
- Stateless `RestoreSnapshot` fallback for idle sandboxes after VMD restart

**Phase 4 — Host health**
- VMD → control plane heartbeat (30s interval)
- Unhealthy host detection (2 min timeout) with auto-recovery
- Internal API auth via shared token

**Performance**
- Scheduler host list cached (30s TTL)
- `CreateSandbox`: DB INSERT parallelized with VMD call, rootfs copy parallelized with netns setup
- `PauseSandbox`: 5 DB roundtrips collapsed to 2 via BeginPause + FinalizePause CTEs
- Network slot pool: 32 pre-allocated + 100 recycled slots
- Tenant-scoped `GetSnapshot` query (SQL-layer isolation)
- VMInstance field writes protected with `inst.mu` (race fix)

**Deploy pipeline**
- Split `deploy-vmd.yml` and `deploy-proxy.yml` — proxy changes no longer restart VMD
- VMD pipeline skips boxd build/inject when boxd source unchanged
- VMD pipeline installs systemd units, fc-cleanup, creates directories
- All paths standardized to `/var/lib/sandbox/` and `/etc/sandbox/`

## Staging test results

- [x] Sandbox create, exec, pause, resume, delete — full lifecycle
- [x] Sandbox survives VMD restart (4+ restarts tested)
- [x] Sandbox survives VMD deploy (4+ deploys tested)
- [x] Files and processes inside VM survive restart
- [x] Template caching — skips cold boot when rootfs unchanged
- [x] Dead VM detection — exec/pause on killed VM returns 410 Gone
- [x] Delete on dead VM — 204 (idempotent)
- [x] Reconciler marks dead VM as failed in DB
- [x] Stateless restore fallback — resume paused sandbox after VMD restart
- [x] Heartbeat endpoint working
- [x] Proxy restart doesn't affect VMD or sandboxes (same PID verified)
- [x] Concurrent creates (5 parallel, all under 1.1s)
- [x] All unit tests pass with `-race`


supabase/migrations/20260410000002_reconciler_log.sql
```diff
@@ -0,0 +1,24 @@
+-- Audit log for every reconciler action. On-call reads this table first
+-- when diagnosing "my sandbox died at 3am" — no silent actions.
+
+CREATE TABLE reconciler_log (
```
pavitrabhalla, line 4: can you explain this table? Seems like a very narrow purpose. Who / what service updates this table?

meAmitPatil: We have a reconciler which is a vmd side loop that fixes the drift between the DB, bolt db, and the actual VM. This table has its audit trail.

pavitrabhalla:
@codex

pavitrabhalla (approved):
@meAmitPatil Please review and address feedback from Codex before merging

PR #29, opened 2026-04-13
Amit/prod infra setup

meAmitPatil:


pavitrabhalla (commented):
A few things to address before merging:

**1. Base branch is `amit/sandbox-env-vars`, not `main`**

This is stacked on #27. Make sure it's not accidentally merged until that one lands, or rebase onto `main`. Would be good to note this in the PR description.

**2. Production deploys automatically on every push to `main`**

```yaml
if: github.event_name == 'push' || github.event.inputs.environment == 'production'
```

Every merge to `main` triggers production after staging passes. That's extremely aggressive — a bad commit gets auto-promoted to prod if staging happens to pass. I'd expect this condition to be `github.event.inputs.environment == 'production'` only, making prod an explicit manual step. If auto-promotion is actually the intent, it should at minimum be documented and probably require a manual approval gate on the `production` GitHub environment.

**3. `proxy.service` loses its `PROXY_DOMAIN` fallback**

The old service file had:
```
Environment=PROXY_DOMAIN=sandbox.superserve.ai
```

This PR removes it and moves `PROXY_DOMAIN` solely into `/etc/superserve/proxy.env`. The `EnvironmentFile` has a `-` prefix (optional), so on any instance where the env file doesn't exist (fresh provisioning, before the deploy script runs, or any failure mid-deploy), `PROXY_DOMAIN` will be unset and the proxy will silently misconfigure. The safe default should remain in the service file, with the env file able to override it.

**4. `proxy_domain` is not validated**

`terminal_origins` and `require_data_plane` both get regex-checked before being interpolated into the shell script. `proxy_domain` does not:

```python
proxy_domain = os.environ.get('PROXY_DOMAIN', 'sandbox.superserve.ai')
# no validation
```

If `PROXY_DOMAIN_PROD` contains a newline, it'll inject an extra line into the `PROXYENV` heredoc, potentially overwriting another env var in the file. Should add something like:
```python
if proxy_domain and not _re.fullmatch(r'[A-Za-z0-9.\-:]+', proxy_domain):
    print('ERROR: PROXY_DOMAIN contains disallowed characters', file=sys.stderr)
    sys.exit(1)
```

**5. Staging `proxy_domain` default is the production domain**

Both jobs default to `sandbox.superserve.ai` when `PROXY_DOMAIN` isn't set. The staging job's default should be the staging domain (or empty), not the production one — otherwise a misconfigured staging var silently points at prod.

**6. ~200 lines of copy-pasted deploy logic**

The production job is a near-exact duplicate of the staging job. This is fragile — any fix (like adding the `proxy_domain` validation above) must be applied twice. Consider extracting the Python script to `deploy/deploy.py` and calling it from both jobs, or using a reusable workflow.

.github/workflows/deploy-vmd.yml
```diff
@@ -277,3 +288,229 @@ jobs:
...
+                      tag = f"{inst['name']}/{inst['zone']}"
+                      print(f'[{tag}] FAILED: {e}', file=sys.stderr)
+                      failed.append(tag)
+
+          if failed:
+              print(f'Deploy failed on: {", ".join(failed)}', file=sys.stderr)
+              sys.exit(1)
+
+          print(f'Deployed to {len(instances)} instance(s). sha={sha}')
+          PYEOF
```
pavitrabhalla: can you extract this into a script file?

PR #30, opened 2026-04-14
remove half-implemented auto-resume, rename idle to paused, add orphan snapshot GC on pause

meAmitPatil:
## Summary

Removes the half-finished auto-resume-on-traffic plumbing, renames the `idle` sandbox state to `paused`, and aligns the snapshot DB row with VMD's one-file-per-sandbox disk layout via an UPSERT in `FinalizePause`. Also hardens the resume path with a boxd readiness gate so exec-after-resume no longer races.

## What changed

**Removed (auto-resume-on-traffic):**
- `AutoWake()` middleware, `wakeIdleSandbox`, `updateLastActivityAsync`, `sandboxFromContext`.
- `last_activity_at` column and related index (migration `20260420000001`).
- `UpdateLastActivity` / `ListIdleSandboxes` queries.
- Exec handlers now require the sandbox to be `active` — paused sandboxes must be resumed explicitly via `POST /sandboxes/:id/resume`.

**Kept (timeout pause + manual pause/resume):**
- Reaper loop and `timeout_seconds` behavior — after the timeout elapses, a sandbox transitions `active → pausing → paused`; `POST /resume` restores it.
- Manual `POST /pause` / `POST /resume` endpoints.

**Renamed `idle` → `paused`** (migration `20260420000002`): enum value, Go const `SandboxStatusPaused`, API response string, OpenAPI spec, reconciler drift keys.

**FinalizePause: upsert live snapshot in place** (migration `20260421000001` adds a partial unique index on `snapshot(sandbox_id) WHERE saved = false`). VMD writes each pause to the same on-disk path per sandbox and overwrites the file; the DB row is updated in place to match. One live snapshot row per sandbox at all times. `saved=true` rows (templates, user-named snapshots later) are unaffected by the partial index.

**Resume readiness gate:** `ResumeVM` and `RestoreVMSnapshot` now wait for the guest agent to accept connections before returning (10s upper bound, typically <500ms). Closes the race where exec fired immediately after resume could get `connection reset by peer`.

**vmd `DeleteSnapshot` gRPC** with path-traversal protection — paths must live under `<SnapshotDir>/<vm_id>/`. Not currently called from the control plane; available for future janitor/operator use.

## Edge cases handled

- Re-pause: upserts the existing row; `sandbox.snapshot_id` does not rotate.
- First pause on a fresh sandbox: inserts a new row.
- Sandbox deleted mid-pause: `FinalizePause` returns 0 rows, caller maps to `ErrSandboxGone` (410).
- Exec on paused sandbox: 409 with message pointing at `/resume`.
- Resume on VMD that restarted (in-memory map empty): falls back to stateless `RestoreSnapshot` path.
- Historical duplicate live snapshot rows (from older pause flows): migration collapses them to the latest per sandbox before adding the unique index.

## Test plan

- [x] `go test ./...` green.
- [x] `go vet` and `GOOS=linux go build ./...` clean.
- [x] Staging end-to-end: 5 consecutive pause/resume/exec cycles on one sandbox; `snapshot_id` constant across pauses; exec immediately after resume succeeds with no artificial wait.
- [x] State-machine edges: resume-active → 409, pause-paused → 409, exec-paused → 409, delete-while-paused → 204.
- [x] Disk check on staging host: exactly one `vmstate.snap`/`mem.snap` pair per sandbox, no subdirs or orphans.


pavitrabhalla (approved):
## Code Review

Overall this is a solid, well-structured cleanup. The state machine simplification (explicit resume instead of auto-wake) is the right call, the UPSERT in `FinalizePause` is correct, and the readiness gate closes a real race. A few items worth flagging:

---

### Issues

**`from_snapshot` removed without deprecation** — `CreateSandboxRequest.from_snapshot` is dropped from the OpenAPI spec and handler with no migration path. Any client currently using this field will silently get a validation error. Worth either keeping it for one release cycle with a deprecation warning, or at minimum documenting it as a breaking change in the changelog.

**Orphaned snapshot files on re-pause** — `FinalizePause` UPSERTs the DB row in place, but the old snapshot files from the previous pause remain on disk. `DeleteSnapshot` gRPC is wired but not called anywhere. Disk accumulates old `vmstate.snap`/`mem.snap` files across multiple pause cycles. This needs a follow-up GC task (ideally calling `DeleteSnapshot` after the UPSERT confirms the new paths).

---

### Minor

- `waitForBoxd` after `RestoreVMSnapshot` only verifies boxd is alive at that instant — a crash between the check and the first exec still produces a `connection reset`. Acceptable for now but worth noting if this surfaces in prod.
- The partial unique index cleanup migration correctly deduplicates historical rows before adding the constraint. Good defensive migration.
- Path-traversal guards in `assertUnderVMSnapshotDir` are thorough and well-tested.

---

### Positives

- All CI green. State-machine logic is clean and correct.
- UPSERT in `FinalizePause` correctly handles re-pause, first pause, and mid-pause deletes atomically.
- Test coverage for path traversal is comprehensive.
- Reconciler and proto changes are correctly aligned with the rename.

Approving — the snapshot GC follow-up should be tracked separately.

PR #33, opened 2026-04-20
ci: add prod deploy jobs to deploy-vmd and deploy-proxy

meAmitPatil:
Adds prod deploy jobs to the VMD and Proxy CD workflows, matching the pattern in deploy-api.


.github/workflows/deploy-proxy.yml
```diff
@@ -170,3 +183,165 @@ jobs:
...
+        env:
+          GCP_PROJECT: ${{ vars.GCP_PROJECT }}
+          VMD_LABEL: ${{ vars.VMD_LABEL }}
+          VMD_INSTALL_DIR: ${{ vars.VMD_INSTALL_DIR }}
+          SHA: ${{ github.sha }}
+          SANDBOX_ACCESS_TOKEN_SEED: ${{ secrets.SANDBOX_ACCESS_TOKEN_SEED_PROD }}
+          PROXY_ALLOWED_ORIGINS: ${{ vars.PROXY_ALLOWED_ORIGINS_PROD }}
+          REQUIRE_DATA_PLANE: ${{ vars.REQUIRE_DATA_PLANE_PROD }}
+        run: |
+          python3 - <<'PYEOF'
```
pavitrabhalla: Can you extract this script out into its own file?

.github/workflows/deploy-vmd.yml
```diff
@@ -233,3 +239,205 @@ jobs:
...
+      - name: Discover and deploy to VMD instances
+        env:
+          GCP_PROJECT: ${{ vars.GCP_PROJECT }}
+          VMD_LABEL: ${{ vars.VMD_LABEL }}
+          VMD_SERVICE: ${{ vars.VMD_SERVICE }}
+          VMD_INSTALL_DIR: ${{ vars.VMD_INSTALL_DIR }}
+          SHA: ${{ github.sha }}
+          BOXD_CHANGED: ${{ steps.boxd-changed.outputs.changed }}
+        run: |
+          python3 - <<'PYEOF'
```
pavitrabhalla: Same here

.github/workflows/deploy-proxy.yml
meAmitPatil: Okay will do

pavitrabhalla (approved):
## Code Review

Clean refactor — pulling the inline heredoc Python into proper script files is the right call, and the staging-gate pattern is consistent with deploy-api. A few things to flag:

---

### Issues

**Unvalidated `install_dir` and `service` interpolated into shell** — `VMD_INSTALL_DIR` and `VMD_SERVICE` are pulled from GH Actions vars and interpolated directly into the f-string deploy scripts (e.g. `sudo mv ... {install_dir}/proxy`, `sudo systemctl restart {service}`). Unlike `access_seed`/`terminal_origins`/`require_data_plane`, these have no allowlist or regex check. A malicious or misconfigured var value would execute arbitrary shell on every VMD host in parallel. These are repo-level vars (not user-controlled), so the blast radius is limited, but it's worth adding basic sanity checks (`/` prefix, no spaces/semicolons) in the same style as the seed validation.

**No manual approval gate on production** — On every push to `main`, production auto-deploys after staging passes. The `sleep 3 && is-active` health check is the only gate. If a service passes that check but has a deeper issue (e.g. a goroutine panic that doesn't surface for a few seconds), production gets the bad binary. This is the same pattern as deploy-api so it's a conscious choice — just documenting the risk.

---

### Minor

- `git diff --name-only HEAD~1 HEAD` for `boxd-changed` breaks on the initial commit (empty history). Fine in practice but worth noting.
- Staging uses `${{ vars.PROXY_ALLOWED_ORIGINS_STAGING }}` (repo-level, suffixed) while `GCP_PROJECT` uses environment-scoped vars (no suffix). Mixing the two conventions is a footgun if someone sets up a new environment — worth picking one approach.
- `VMD_LABEL` and `VMD_SERVICE` docstring marks them as `required` but the Python has `.get()` defaults (`component=vmd`, `superserve-vmd`). Defaults are useful, but the docstring should say optional or the `.get()` should be `["VMD_LABEL"]` to fail loudly.

---

### Positives

- Extracting the scripts makes them diffable, testable locally, and not subject to YAML quoting hell.
- `needs: [deploy-staging]` correctly enforces staging-first on both push and manual dispatch.
- `workflow_dispatch` with `environment` choice for on-demand production deploy is a nice escape hatch.
- ThreadPoolExecutor error aggregation is preserved faithfully from the original.

Approving — the `install_dir`/`service` input validation is worth a quick follow-up.

PR #34, opened 2026-04-21
exec: auto-resume paused sandboxes

meAmitPatil:
## Summary

`POST /exec` and `POST /exec/stream` on a paused sandbox now auto-resume it, run the command, and leave it active. Callers no longer need to issue `POST /resume` first.

## What changed

- Extract `resumePausedSandbox` — shared by explicit `/resume` and the auto-resume path.
- `loadActiveOrResumeSandbox`: `active` → use; `paused` → auto-resume; anything else → 409.
- Atomic `BeginResume` (new `resuming` enum value, migration `20260421000002`) serializes concurrent resumes at the DB — losers get 409.
- Serialize network reapply before DB activate so "active ⇒ rules applied" holds by construction. Failure after VMD resume destroys the VM + reverts DB to paused.

## Test plan

- [x] Unit + integration + race all green.
- [x] Tested on staging.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -249,8 +249,8 @@ paths:
       summary: Execute a command inside a sandbox
       description: |
         Runs a command inside the sandbox and waits for it to finish, returning
-        stdout, stderr, and exit code. Paused sandboxes are rejected with 409;
-        resume explicitly via POST /sandboxes/{sandbox_id}/resume first.
+        stdout, stderr, and exit code. A paused sandbox is resumed automatically
+        before the command runs and stays active afterward.
       security:
         - apiKey: []
       requestBody:
@@ -286,8 +286,8 @@ paths:
       description: |
         Runs a command and streams stdout/stderr chunks as Server-Sent Events.
         Each event is a JSON payload. The final event includes `exit_code` and
-        `finished: true`. Paused sandboxes are rejected with 409; resume
-        explicitly via POST /sandboxes/{sandbox_id}/resume first.
+        `finished: true`. A paused sandbox is resumed automatically before the
+        command runs and stays active afterward.
       security:
         - apiKey: []
       requestBody:
--- a/db/queries/sandboxes.sql
+++ b/db/queries/sandboxes.sql
@@ -88,6 +88,24 @@ SET status = 'pausing', updated_at = now()
 WHERE id = $1 AND team_id = $2 AND destroyed_at IS NULL AND status = 'active'
 RETURNING *;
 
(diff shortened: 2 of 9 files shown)


pavitrabhalla (approved):
## Code Review

Solid implementation. The `BeginResume` atomic claim pattern is the right approach for serializing concurrent auto-resumes, and the compensation logic (`destroyAndRevert`) is ordered correctly — network failure and activate failure both destroy the VM before reverting status. Test coverage is exceptional, especially `TestExecSandbox_ConcurrentResumeSerializes` which validates the race-winning contract end-to-end.

---

### Issues

**`resuming` status missing from OpenAPI spec** — `openapi.yaml` still only enumerates `active` and `paused`. A client polling `GET /sandboxes/:id` during a slow resume will see `status: "resuming"` which is undocumented. Either add `resuming` to the enum (as a transient state with a note that clients should treat it as "not yet active") or document it in the status field description.

**No reaper for stuck `resuming` sandboxes** — If the control plane process crashes after `BeginResume` but before `ActivateSandbox` / `RevertResumeToPaused`, the sandbox is stuck in `resuming` indefinitely. Same gap exists for `pausing` and is presumably acceptable until a general stuck-transient-state reaper is added — just flagging so it's tracked together.

---

### Minor

- `ActivateSandbox` is unguarded on status (just `WHERE ... AND destroyed_at IS NULL`), which is correct for `resuming → active` — confirmed by reading the query. No change needed, just noting it's intentionally open.
- `postCtx` timeout covers both `reapplyNetworkConfig` and `ActivateSandbox` under a single `vmdTimeout`. Technically a network reapply that takes nearly `vmdTimeout` could starve the `ActivateSandbox` call. In practice both are sub-second; fine to keep as-is unless you see latency issues in prod.
- The early `sandbox.Status != SandboxStatusPaused` guard in `ResumeSandbox` is now technically redundant (BeginResume returns 0 rows for non-paused states), but it gives an immediate 409 without a DB write for the active/pausing/resuming cases — worth keeping.

---

### Positives

- `RevertResumeToPaused` guarded on `status = 'resuming'` prevents clobbering a successful concurrent activate. Correct.
- `context.WithoutCancel` on both `revertCtx` and `postCtx` ensures client disconnect can't leave the VM up while the DB says paused. Correct.
- `destroyAndRevert` calls `DestroyInstance` before `revertToPaused` — right order (VM down before DB says paused, so a concurrent watcher doesn't try to resume a running VM).

Approving — the OpenAPI spec gap is the only thing worth fixing before shipping.

PR #42, opened 2026-04-29
templates: rename alias to name; free the name on soft-delete

meAmitPatil:
## Summary

Two related changes that bring the template API in line with how users expect it to work.

**1. Rename for templates `alias` → `name` everywhere.** More intuitive, matches `sandbox.name`, no more "alias is jargony" friction.

**2. Free the name slot on soft-delete.** Replace the full `UNIQUE (team_id, alias)` constraint with a partial unique index on `(team_id, name) WHERE deleted_at IS NULL`. After this, the user-visible flow is:

- `POST /templates {name: "abc"}` → 202
- `DELETE /templates/{id}` → 204
- `POST /templates {name: "abc"}` → **202** (was 409)

Soft-delete still happens internally (audit trail, foreign-key safety with sandboxes/builds). User just sees: created, deleted, created — name reusable.

## Touched

- `supabase/migrations/20260429000001_template_alias_to_name.sql` — rename column + replace constraint with partial index.
- `db/queries/templates.sql` — column references, `GetTemplateByAlias` → `GetTemplateByName`, list-filter param.
- `internal/db/{models,templates.sql}.go` — sqlc regen.
- `internal/api/handlers{,_template,_test}.go` — request/response struct fields, validation, error messages.
- `cmd/seed-templates/main.go` — `seedSpec.Name`, upsert SQL with new partial-index ON CONFLICT.
- `superserve_templates/*.json` — 6 system templates use `"name"` instead of `"alias"`.
- `api/openapi.yaml` — request/response schemas, list filter, descriptions.
- `internal/integration/integration_test.go` — test seed fixture.

## Breaking change

API field renamed without compat shim. Pre-existing template rows are migrated by `RENAME COLUMN`. SDK callers using `alias` will need to switch to `name`.

### DIFF
--- a/cmd/seed-templates/main.go
+++ b/cmd/seed-templates/main.go
@@ -4,7 +4,7 @@
 //
 // Flow per run:
 //   1. Upsert each JSON under superserve_templates/ into the template table
-//      owned by SYSTEM_TEAM_ID (idempotent on alias).
+//      owned by SYSTEM_TEAM_ID (idempotent on name).
 //   2. Decide whether to enqueue a build for each template:
 //        - Template row just inserted → enqueue.
 //        - Previous build failed / never ran → enqueue.
@@ -98,7 +98,7 @@ func main() {
 		log.Warn().Str("dir", dir).Msg("no seed files found")
 		return
 	}
-	sort.Slice(specs, func(i, j int) bool { return specs[i].Alias < specs[j].Alias })
+	sort.Slice(specs, func(i, j int) bool { return specs[i].Name < specs[j].Name })
 
 	var queued []uuid.UUID
 	var seedFailures int
@@ -106,15 +106,15 @@ func main() {
 		buildID, queuedIt, err := seedOne(ctx, pool, systemTeamID, s, forceRebuild)
 		if err != nil {
 			seedFailures++
-			log.Error().Err(err).Str("alias", s.Alias).Msg("seed failed")
+			log.Error().Err(err).Str("name", s.Name).Msg("seed failed")
 			continue
 		}
 		if !queuedIt {
-			log.Info().Str("alias", s.Alias).Msg("already up-to-date; nothing to build (pass --force-rebuild to override)")
+			log.Info().Str("name", s.Name).Msg("already up-to-date; nothing to build (pass --force-rebuild to override)")
 			continue
 		}
 		queued = append(queued, buildID)
-		log.Info().Str("alias", s.Alias).Str("build_id", buildID.String()).Msg("build queued")
+		log.Info().Str("name", s.Name).Str("build_id", buildID.String()).Msg("build queued")
 	}
 
 	if noWait || len(queued) == 0 {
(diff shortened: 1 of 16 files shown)


supabase/migrations/20260429000001_template_alias_to_name.sql
```diff
@@ -0,0 +1,11 @@
+-- Rename template.alias → template.name (more intuitive, matches sandbox.name).
+-- Replace the full UNIQUE (team_id, alias) with a partial unique index that
+-- only enforces uniqueness on non-soft-deleted rows, so a deleted template's
+-- name is reusable.
+
+ALTER TABLE template RENAME COLUMN alias TO name;
+
+ALTER TABLE template DROP CONSTRAINT template_alias_unique_per_team;
+
```
pavitrabhalla, line 9: Verify this constraint name matches what's actually in production before deploying. If it differs the migration will fail with an error at deploy time. Worth a quick check against the live schema (`\d template` or prior migrations) before merging.

cmd/seed-templates/main.go
```diff
@@ -193,9 +193,9 @@ func seedOne(ctx context.Context, pool *pgxpool.Pool, teamID uuid.UUID, s seedSp
 	// Upsert and capture the resulting status so we can decide whether to
 	// enqueue a build in one follow-up query.
 	const upsertQ = `
-		INSERT INTO template (team_id, alias, build_spec, vcpu, memory_mib, disk_mib)
+		INSERT INTO template (team_id, name, build_spec, vcpu, memory_mib, disk_mib)
 		VALUES ($1, $2, $3, $4, $5, $6)
-		ON CONFLICT (team_id, alias) DO UPDATE
+		ON CONFLICT (team_id, name) WHERE deleted_at IS NULL DO UPDATE
 		SET build_spec = EXCLUDED.build_spec,
```
pavitrabhalla, line 200: Subtle behavior change: this partial `ON CONFLICT` only fires when the conflicting row has `deleted_at IS NULL`. If a system template row were ever soft-deleted, re-seeding would silently insert a *new* row rather than updating the old one — leaving two rows for the same `(team_id, name)`: one deleted, one fresh. System templates should never be deleted in practice, but a short comment here noting this edge case would prevent a future reader from being confused.

pavitrabhalla (commented):
Thorough, well-executed rename. The partial index is the right approach. A couple of things before merging:

**Missing test for the headline feature.** The name-reuse-after-delete behavior — the whole point of the partial index — has no test. The PR description already has the exact happy path (`POST → DELETE → POST same name → 202`). Without a test, a future migration or constraint change could silently regress this. Please add a case in `handlers_test.go` or the integration suite.

**Name character set.** `name` currently accepts any 1–128-character string (same as the old `alias`). Since this is already a breaking change, now is a natural moment to decide consciously whether to restrict to URL-safe characters (e.g. `[a-z0-9/_-]`). Not blocking, but harder to enforce later once names are in the wild.

meAmitPatil:
@pavitrabhalla pushed review fixes.

pavitrabhalla (approved):

PR #48, opened 2026-05-05
vmd,template-builder: per-VM rootfs at the magic path so sandboxes don't share a backing file

meAmitPatil:
Every sandbox from the same template was opening the same `rootfs.ext4` file on the host: the snapshot baked the build-VM's path while vmd's per-VM mount-namespace symlink targeted a different path that was never actually used. Concurrent writes corrupted each other and surfaced as ext4 "Structure needs cleaning" / dpkg failures.

Fix: both sides now use the same magic path so vmd's mount-namespace symlink actually intercepts, and the per-VM rootfs is copied from the template's rootfs (not the global base).

## Validated end-to-end on staging

- **Per-VM isolation**: lsof on staging vmd while two concurrent sandboxes from `superserve/base` are running — each Firecracker has its own `/var/lib/sandbox/rundir/<vm_id>/rootfs.ext4` (no shared inode).
- **Original failure gone**: ran the exact snippet that previously hit `dpkg: ... 'AdduserCommon.pm': Structure needs cleaning` against staging — Mesa installer completes cleanly end-to-end.
- **Concurrent stress**: 5 sandboxes from the same template, parallel `apt-get install` of `gpg` / `jq` / `adduser` (the package that broke before) — 0 ext4 errors, 0 dpkg errors, all exit 0.
- **Custom-template path**: built a custom template with the Mesa docs' build_spec, spun 5 concurrent sandboxes from it, parallel apt-installs — all clean. Same lsof check confirmed each had its own rootfs file.

Rollout: deploy → force-rebuild templates.

### DIFF
--- a/cmd/template-builder/main.go
+++ b/cmd/template-builder/main.go
@@ -177,9 +177,9 @@ type buildConfig struct {
 
 func runBuild(ctx context.Context, cfg buildConfig) error {
 	buildVMID := "build-" + cfg.templateID
-	rootfsDir := filepath.Join(cfg.runDir, "templates", cfg.templateID)
+	rootfsDir := filepath.Join(cfg.runDir, vm.TemplatesDirName, cfg.templateID)
 	rootfsPath := filepath.Join(rootfsDir, "rootfs.ext4")
-	snapDir := filepath.Join(cfg.snapshotDir, "templates", cfg.templateID)
+	snapDir := filepath.Join(cfg.snapshotDir, vm.TemplatesDirName, cfg.templateID)
 
 	if err := os.MkdirAll(rootfsDir, 0o755); err != nil {
 		return fmt.Errorf("mkdir rootfs dir: %w", err)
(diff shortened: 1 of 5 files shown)


pavitrabhalla:
## Review

The fix is structurally correct — the root cause (both sides targeting different paths, so the per-VM symlink never intercepted) is cleanly addressed, and the `TestTemplateMagicDir_MatchesTemplateRunDir` test directly enforcing the build/restore path invariant is exactly the right thing to test here.

A few notes:

**Minor issues**

- `resolveRestoreDisk` uses `if err == nil { ... } else { ... }` which is non-idiomatic Go. Normal style would be an early return for the success case and a flat fallthrough.
- The debug log that was removed from `RestoreVMSnapshot` (`"created rootfs copy for restored VM"`) was useful observability — worth keeping or replacing inside `resolveRestoreDisk`.
- No test covers the `copyRootfs` branch of `resolveRestoreDisk` (the case where `templateRootfsForSnapshot` succeeds). A minimal test with a stub file would be straightforward to add.

**Things to watch on rollout**

- Existing snapshots have the old `path_on_host` baked in, so any sandbox from an un-rebuilt template will fail after deploy. The PR description calls this out — just making sure the deploy runbook enforces the force-rebuild before serving traffic.
- `mount --make-rprivate /` inside the build shell script: if template-builder ever runs in a container where `/` propagation flags aren't already set, this could fail before the tmpfs mount. The `&&` chain means Firecracker won't start (right failure mode), and `cmd.Stderr = os.Stderr` makes it visible — just something to keep an eye on.

Overall: approve. The concurrency validation on staging is thorough and the fix is solid.

pavitrabhalla (approved):

PR #49, opened 2026-05-06
vmd,template-builder: overlay-mode templates for fast sandbox create

meAmitPatil:
Block-level COW templates for fast sandbox create.

Today: sandbox create copies a 4 GB rootfs per VM. With reflinks unavailable on the ext4 data disk, `cp --reflink=auto` falls back to full byte copy.

This PR replaces the per-VM copy with a forked-firecracker overlay device: each template publishes a read-only `base.ext4` plus a `rootfs.delta`, and sandbox create just opens an empty per-VM overlay file. The forked engine handles dirty-block routing and delta hydration on snapshot restore.

## What's in the PR

- **Overlay-mode template build** (`template-builder`, `vmd`): build pipeline emits `base.ext4` + `rootfs.delta` + `vmstate.snap{,.overlay}` + `mem.snap` per build. Legacy templates (`base_path = NULL`) continue to use the existing `copyRootfs` path — no behaviour change for them.

- **Per-build artifact layout.** Each build writes to its own `<runDir>/templates/<tplID>/<buildID>/…` dir. Sandbox rows pin their build via 4 new columns (`snapshot_path`, `mem_path`, `base_path`, `delta_path`) so a subsequent template rebuild can't corrupt a paused sandbox. Migration `20260506000001`.

- **DB-driven build reconciler** in the controlplane supervisor. Joins `ListPinnedBuildPaths` + `ListLiveTemplateBuilds` against a new `ListBuildArtifacts` RPC every 5 min and reaps orphan per-build dirs. Mtime grace window protects in-flight builds.

- **`FailBuild` semantics.** A failed rebuild only marks the template `failed` if its current status is `pending` or `building`. A previously-ready template stays `ready` — its prior build's artifacts are still on disk and still usable by paused sandboxes.

- **Magic mount-namespace path swap** so each VM sees its own overlay file at the same baked-in path the firecracker snapshot expects.

## Deploy precondition

Requires the firecracker fork at `feat/block-overlay-cow-v1.15` (commit `0bae3adf7`) deployed to vmd hosts before merge. Stock firecracker rejects `io_engine: "Overlay"` and the snapshot side-car format.


pavitrabhalla (changes requested):
Three items worth addressing before this lands:

---

**[Blocking] FailBuild SQL: returns no rows for already-terminal builds**

`internal/db/templates.sql.go` — the new FailBuild CTE does:

```sql
SELECT t.*
FROM template t
JOIN build_done bd ON bd.template_id = t.id
```

`build_done` is empty when the build's status is already `failed` or `cancelled` (the UPDATE filter `WHERE status IN ('pending', 'building', 'snapshotting')` matches nothing). That JOIN produces zero rows → the caller gets `pgx.ErrNoRows` instead of a Template row, which is a behavior change from the old query that always returned `template.*`.

Please verify that `failBuild` in `build_supervisor.go` handles `ErrNoRows` gracefully, or change the SELECT to a LEFT JOIN so it always returns the template row regardless of whether `tpl_update` fired.

---

**[Blocking] Nil dereference of `*row.VmdBuildVmID` in failure GC path**

`internal/supervisor/build_supervisor.go` — in the new GC block inside `pollOne`:

```go
if gcErr := vmd.DeleteBuildArtifacts(gcCtx, row.TemplateID.String(), *row.VmdBuildVmID); gcErr != nil {
```

`VmdBuildVmID` is a pointer. In normal flow it's always set by the time `pollOne` runs, but a nil dereference here would panic the supervisor goroutine. Please add a nil guard before the dereference:

```go
if row.VmdBuildVmID == nil {
    rowLog.Warn().Msg("cleanup failed-build artifacts: no build VM ID")
} else if gcErr := vmd.DeleteBuildArtifacts(gcCtx, row.TemplateID.String(), *row.VmdBuildVmID); gcErr != nil {
    ...
}
```

---

**[Non-blocking, worth discussing] No host-capability guard for overlay mode**

`internal/vm/firecracker.go` — when `BasePath != ""` the code sends `io_engine: "Overlay"` to the Firecracker API socket. Stock Firecracker rejects this with an opaque error at sandbox create time, not at build time. A host that's still running the unforked binary ends up with a `ready` template that's silently unusable.

Consider: (a) a capability RPC or flag on vmd hosts so the controlplane can refuse dispatch to non-overlay hosts, or (b) at minimum a structured error in the restore path that names the missing fork as the root cause.

meAmitPatil:
@pavitrabhalla Pushed fixes.

pavitrabhalla (approved):
Blocking issues addressed. Fixed the two minor items directly on the branch: named the torn-snapshot marker string as a constant so a fork message change is a compile-time break rather than a silent runtime miss, and propagated directory fsync/close errors from fsyncBuildArtifacts instead of swallowing them. LGTM.

PR #50, opened 2026-05-07
boxd,api: exec timeout returns exit 124 instead of HTTP 500

meAmitPatil:
## Problem

`POST /sandboxes/:id/exec` with `timeout_s` returns HTTP 500 when the timeout fires, instead of a normal `200 { exit_code: 124 }`.

Root cause: boxd's `Start` RPC wrapped one context with the timeout and used it for both the child process and the streaming response. When the timer fired, the child got killed and the stream context got cancelled simultaneously — leaving boxd unable to send the End event, which propagated as an RPC error and got mapped to 500.

## Fix

- **boxd** (`cmd/boxd/main.go`): decouple cmd ctx from stream ctx. Use `time.AfterFunc` + an `atomic.Bool` to kill only the child on timeout; stream stays alive to send the End event with exit 124.
- **boxd**: extract `finalExitCode(ps, timedOut)` to unify exit-code mapping. As a bonus, `startPipes` now maps signal-killed processes to 128+signal, matching `startPTY`.
- **controlplane** (`internal/api/handlers.go`): add a `TimeoutS + 15s` backstop on the gRPC call so a wedged boxd can't hang the request goroutine forever.

### DIFF
--- a/cmd/boxd/exit_test.go
+++ b/cmd/boxd/exit_test.go
@@ -1,0 +1,83 @@
+package main
+
+import (
+	"context"
+	"os/exec"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+func TestFinalExitCode_Normal(t *testing.T) {
+	cmd := exec.Command("/bin/sh", "-c", "exit 7")
+	if err := cmd.Run(); err == nil {
+		t.Fatal("expected non-nil err for non-zero exit")
+	}
+	got := finalExitCode(cmd.ProcessState, false)
+	if got != 7 {
+		t.Errorf("finalExitCode = %d, want 7", got)
+	}
+}
+
+func TestFinalExitCode_TimedOut(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pavitrabhalla (approved):
Two follow-up fixes pushed directly to the branch:

1. **Zero-timeout backstop guard** (): The 15s backstop now only applies when `TimeoutS > 0`. Previously, commands with no timeout would be killed after 15 s unconditionally.

2. **Nil-check `ProcessState.String()`** (): Both `startPTY` and `startPipes` now guard before calling `.String()` to avoid a panic if `ProcessState` is nil.

Rest of the PR looks good — the context decoupling fix is correct and the `finalExitCode` refactor is a nice improvement. LGTM.

pavitrabhalla (approved):
Two follow-up fixes pushed directly to the branch. Rest of the PR looks good — the context decoupling fix is correct and the finalExitCode refactor is a nice improvement. LGTM.

PR #51, opened 2026-05-11
vm,boxd: timeout actually kills child tree + response carries through

meAmitPatil:
## Issue

`POST /sandboxes/:id/exec` with `timeout_s` still returns HTTP 500 `internal_error` when the timeout fires, instead of 200 with `exit_code: 124`. PR #50 fixed the controlplane and (partly) boxd; two more layers were missed.

## Fixes

**1. `internal/vm/manager.go` — vmd stream ctx grace.** `ExecCommand` was wrapping the ctx with exactly `timeout`, so the stream cancelled at the same instant boxd killed the child — racing (and losing to) boxd's `End` event. Extend the stream ctx with `+15s` grace (same value the controlplane already uses). Boxd still kills the child at the user-supplied `timeout`; only the stream gets the extra time to deliver the response. Applied to both `ExecCommand` and `ExecCommandStream`.

**2. `cmd/boxd/main.go` — kill the whole process group.** Boxd's timeout was sending SIGKILL only to the shell (`sh -c "sleep 10"`), leaving the actual `sleep` child orphaned to init. The pipe FDs stayed open, boxd's pipe-reader goroutines blocked, and the End event was delayed until the orphaned child died naturally. Now boxd starts the child in its own pgid and signals the whole group on timeout, so child trees actually die when expected.

## Result

`sleep 10` with `timeout_s=2` now returns **HTTP 200 + exit_code 124 in ~2s** instead of 500 in 2s (pre-fix-1) or 200+124 in 10s (with fix-1 only).

### DIFF
--- a/cmd/boxd/main.go
+++ b/cmd/boxd/main.go
@@ -396,10 +396,23 @@ func (s *processService) Start(ctx context.Context, req *connect.Request[pb.Star
 	}
 
 	isPTY := msg.GetPty() != nil
-	if isPTY {
-		return s.startPTY(ctx, cmd, msg, stream, &timedOut)
+	if !isPTY {
+		// Kill the whole pgid on timeout — killing the shell alone leaves
+		// orphaned children (e.g. `sleep` under `sh -c`) running.
+		if cmd.SysProcAttr == nil {
+			cmd.SysProcAttr = &syscall.SysProcAttr{}
+		}
+		cmd.SysProcAttr.Setpgid = true
+		cmd.Cancel = func() error {
+			if cmd.Process == nil {
+				return nil
+			}
+			return syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+		}
+		cmd.WaitDelay = time.Second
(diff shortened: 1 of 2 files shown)


ghost:
Ready to review this PR? Stage has broken it down into 2 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Ensure process group termination in boxd](https://stagereview.app/superserve-ai/sandbox/pull/51/chapters/1) |
| 2 | [Extend execution stream grace period](https://stagereview.app/superserve-ai/sandbox/pull/51/chapters/2) |

<a href="https://stagereview.app/superserve-ai/sandbox/pull/51">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 1f241d2738a00b7ed87d1f53ab9cf8bb6ae452fb on May 11, 2026 7:15pm UTC.</sub>

pavitrabhalla (approved):

nirnejak (approved):

PR #54, opened 2026-05-12
api: attribute activity rows to api_key.created_by for per-user WAU

meAmitPatil:
activity.actor_id was always NULL. Middleware now reads api_key.created_by and threads it into both activity log helpers so each row gets the profile id of the api-key creator. System events (reaper, build_completed) stay NULL by design.

---

<a href="https://stagereview.app/superserve-ai/sandbox/pull/54">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -99,7 +99,7 @@ const asyncTimeout = 5 * time.Second
 // context.WithoutCancel so the goroutine is not killed when the HTTP response
 // completes, but the trace/span context is kept so the async write appears in
 // the same request trace as the synchronous work.
-func (h *Handlers) logSandboxActivity(reqCtx context.Context, sandboxID, teamID uuid.UUID, category, action, status string, sandboxName *string, durationMs *int32, metadata []byte) {
+func (h *Handlers) logSandboxActivity(reqCtx context.Context, sandboxID, teamID uuid.UUID, actorID *uuid.UUID, category, action, status string, sandboxName *string, durationMs *int32, metadata []byte) {
 	asyncCtx := context.WithoutCancel(reqCtx)
 	go func() {
 		ctx, cancel := context.WithTimeout(asyncCtx, asyncTimeout)
@@ -108,6 +108,7 @@ func (h *Handlers) logSandboxActivity(reqCtx context.Context, sandboxID, teamID 
 			SandboxID:    pgtype.UUID{Bytes: sandboxID, Valid: true},
 			ResourceType: "sandbox",
 			TeamID:       teamID,
+			ActorID:      actorUUID(actorID),
 			Category:     category,
 			Action:       action,
 			Status:       &status,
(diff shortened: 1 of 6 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Extract actor ID from API keys in middleware](https://stagereview.app/superserve-ai/sandbox/pull/54/chapters/1) |
| 2 | [Update activity logging helpers to support actor attribution](https://stagereview.app/superserve-ai/sandbox/pull/54/chapters/2) |
| 3 | [Wire actor ID through sandbox and template handlers](https://stagereview.app/superserve-ai/sandbox/pull/54/chapters/3) |
| 4 | [Verify actor attribution with integration tests](https://stagereview.app/superserve-ai/sandbox/pull/54/chapters/4) |

<a href="https://stagereview.app/superserve-ai/sandbox/pull/54">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit ee68d681ebc0b1295933396d46195e23caa430d5 on May 12, 2026 9:24pm UTC.</sub>

pavitrabhalla (approved):

pavitrabhalla:
## Review

### What's Good

- **Correct NULL design for system events.** The reaper's `timeout_paused` call explicitly passes `nil`, and `build_supervisor.go`'s `logBuildCompleted` has no `ActorID` field — both intentional and matching the PR description.
- **`actorUUID` helper** cleanly centralizes the `*uuid.UUID → pgtype.UUID` conversion and handles nil safely.
- **`actorIDFromContext` is defensive** — double type-assertion guards against a bad cast without panicking.
- **The middleware change is minimal** — one extra column in an existing query, stored only when `Valid`.

---

### Issues & Suggestions

**1. Missing DB migration confirmation**

The query `SELECT id, team_id, created_by FROM api_key` requires the column to exist. `created_by` is defined in `20260403000003_api_key.sql`, so existing deployments should be fine — but worth confirming there's no environment where that migration hasn't run yet (e.g. a local dev schema).

**2. `pgtype` import in `middleware.go`**

Minor: `pgtype` is now a dependency of middleware just for the `Scan` target. An alternative is to scan into a nullable `*string` or `*[16]byte` and convert only if non-nil — keeps middleware decoupled from pgtype. Not a blocker.

**3. `&id` in `actorIDFromContext` looks like a dangling-pointer trap**

```go
id, ok := raw.(uuid.UUID)
// ...
return &id
```

This is correct Go — `id` escapes to heap — but a future reader may flag it as suspicious. A short comment would help.

**4. No test coverage for `NULL created_by`**

If `api_key.created_by IS NULL`, `createdBy.Valid` is false, `actor_id` is not set in the gin context, and `actorIDFromContext` returns `nil`. This is the right behavior, but there are no tests to codify it. Worth adding.

---

Overall this is a solid, focused change. Main things before merging: confirm the migration has run everywhere, add a note on `&id`, and consider a test for the NULL path.

PR #60, opened 2026-05-18
template-builder: flatten rootfs.delta into base.ext4 after snapshot so per-sandbox restores don't replay the delta

meAmitPatil:
Eliminates the per-sandbox ~delta-size disk write during create — the writes that saturated the host disk during 100-concurrent burst now happen once at template build time instead of N times at sandbox create time.

---

<a href="https://stagereview.app/superserve-ai/sandbox/pull/60">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/cmd/template-builder/main.go
+++ b/cmd/template-builder/main.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/superserve-ai/sandbox/internal/builder"
 	"github.com/superserve-ai/sandbox/internal/network"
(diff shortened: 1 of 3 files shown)


meAmitPatil:
Superseded by moving flatten into firecracker as a snapshot-editor subcommand (see superserve-ai/firecracker#10). New sandbox-side PR will shell out to snapshot-editor — eliminates the Go/Rust binary-format duplication that caused the CRC saga, and fixes the latent stale-side-car bug.

PR #11, opened 2026-04-03
test: unit + integration tests for API handlers, CI update

pavitrabhalla:
## Summary
Comprehensive test suite for all API handlers, open-source repo standard.

**Unit tests** (`internal/api/handlers_test.go`, `middleware_test.go`):
- Mock VMDClient for each handler
- Auth middleware: valid key, expired, revoked, missing, bad hash
- Create, delete, exec, pause, resume, file upload/download
- Team isolation: can't access other team's sandboxes
- State validation: reject operations on wrong sandbox status
- Table-driven tests with clear names

**Integration tests** (`internal/integration/integration_test.go`):
- Real Postgres (testcontainers or CI service)
- Full lifecycle: create → exec → pause → resume → delete
- Snapshot visibility (saved vs unsaved)
- Activity log validation
- Concurrent operations

**CI update**:
- Unit tests: `go test -short ./...`
- Integration tests: `go test -run Integration ./...` with Postgres service

Closes sa-n74

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #62, opened 2026-05-19
template-builder: pass flatten=true to CreateSnapshot

meAmitPatil:
## Summary

Tells the forked firecracker (superserve-ai/firecracker#11) to bake each overlay's dirty blocks into its base.ext4 and zero the side-car bitmap as part of snapshot creation. Sandboxes restored from templates built after this lands start with an empty per-VM overlay — no apply_delta replay on create.

Single new `flatten bool` arg on `vm.CreateSnapshot`. Template-builder passes `true`; the two other callers (regular snapshot + diff snapshot) pass `false`.

**Deploy order:** firecracker PR #11 must land + new binary deployed first. With old firecracker, `flatten: true` is silently ignored (the field is `#[serde(default)]`), so the worst case during rollout is templates that aren't flat yet — no breakage.

---

<a href="https://stagereview.app/superserve-ai/sandbox/pull/62">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/cmd/template-builder/main.go
+++ b/cmd/template-builder/main.go
@@ -289,7 +289,8 @@ func runBuild(ctx context.Context, cfg buildConfig) error {
 	emitUser("system", "Saving template")
 	snapPath := filepath.Join(snapDir, "vmstate.snap")
 	memPath := filepath.Join(snapDir, "mem.snap")
-	if err := vm.CreateSnapshot(socketPath, snapPath, memPath, snapDir); err != nil {
+	// flatten=true: per-sandbox restores skip apply_delta. Safe here — base isn't shared yet.
+	if err := vm.CreateSnapshot(socketPath, snapPath, memPath, snapDir, true); err != nil {
 		return fmt.Errorf("snapshot: %w", err)
 	}
 	emitInternal("system", "snapshot captured")
--- a/cmd/template-builder/main.go
+++ b/cmd/template-builder/main.go
@@ -289,8 +289,8 @@ func runBuild(ctx context.Context, cfg buildConfig) error {
 	emitUser("system", "Saving template")
 	snapPath := filepath.Join(snapDir, "vmstate.snap")
 	memPath := filepath.Join(snapDir, "mem.snap")
-	// flatten=true: per-sandbox restores skip apply_delta. Safe here — base isn't shared yet.
-	if err := vm.CreateSnapshot(socketPath, snapPath, memPath, snapDir, true); err != nil {
+	// Flatten: per-sandbox restores skip apply_delta. Safe here — base isn't shared yet.
+	if err := vm.CreateSnapshot(socketPath, snapPath, memPath, snapDir, vm.SnapshotFlatten); err != nil {
 		return fmt.Errorf("snapshot: %w", err)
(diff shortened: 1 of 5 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Add flatten field to Firecracker snapshot model](https://stagereview.app/superserve-ai/sandbox/pull/62/chapters/1) |
| 2 | [Introduce SnapshotMode and update CreateSnapshot signature](https://stagereview.app/superserve-ai/sandbox/pull/62/chapters/2) |
| 3 | [Enable flattening in template-builder and update callers](https://stagereview.app/superserve-ai/sandbox/pull/62/chapters/3) |
| 4 | [Test snapshot serialization and validation logic](https://stagereview.app/superserve-ai/sandbox/pull/62/chapters/4) |

<a href="https://stagereview.app/superserve-ai/sandbox/pull/62">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit d6335b002573298f07b29804f0ff6a2d1eae5b83 on May 20, 2026 7:03pm UTC.</sub>

pavitrabhalla:
Overall this is clean and well-scoped — the zero-value-is-safe enum design and the deploy-order note are both good. A few things worth addressing before merge:

**1. Missing guard: `flatten=true` requires non-empty `blockDeltaDir`**

The model comment says _"Requires block_delta_dir"_ but `CreateSnapshot` doesn't enforce it. A caller passing `SnapshotFlatten` with `blockDeltaDir=""` will send `{"flatten":true}` with no `block_delta_dir` and get a confusing error back from Firecracker. Suggest adding an early return in `CreateSnapshot`:

```go
if mode == SnapshotFlatten && blockDeltaDir == "" {
    return fmt.Errorf("flatten mode requires a non-empty blockDeltaDir")
}
```

**2. Data race in the test**

`capturedBody` is written in the HTTP handler goroutine and read on the test goroutine after `CreateSnapshot` returns. It works in practice (response is sent after body is read, so write happens-before return), but `go test -race` will flag it. A `sync.Mutex` around the read/write fixes it.

**3. Server readiness race in the test**

`go srv.Serve(ln)` is called immediately before `CreateSnapshot` with no guarantee the server is ready. Usually fine, but can produce flaky results. A small `net.Dial` retry loop or a ready channel would harden it.

meAmitPatil:
@pavitrabhalla Pushed Fixes.

pavitrabhalla:
**Blocking:** nothing.

**Non-blocking suggestions:**

- `TestCreateSnapshot_FlattenFieldInJSONBody` passes a non-empty `blockDeltaDir` for both cases, but real `PauseVM`/`CreateVMSnapshot` calls use an empty one. Worth adding a `SnapshotNormal + ""` case to cover that path.

- For the normal-mode test case, `json.Unmarshal` returns `false` for both "field absent" and `"flatten": false` — so the assertion wouldn't catch an accidentally dropped `omitempty` tag. A tighter check would assert `!strings.Contains(string(body), "flatten")` for that case.

pavitrabhalla (approved):

PR #64, opened 2026-05-19
vmd: rebuild missing netns on resume, validate pool slots, allocator consults kernel

meAmitPatil:
Three network reliability fixes hardening vmd against missing-ns / poisoned-pool / colliding-slot edge cases.

---

<a href="https://stagereview.app/superserve-ai/sandbox/pull/64">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/internal/network/hostfw.go
+++ b/internal/network/hostfw.go
@@ -10,6 +10,13 @@ import (
 	"github.com/rs/zerolog"
 )
 
+// hostFirewallAPI is the Manager's view of *HostFirewall, stubbed by tests.
+type hostFirewallAPI interface {
+	AddVM(vmID, vethName, hostCIDR string) error
+	RemoveVM(vmID string) error
+	Close() error
+}
+
 // HostFirewall manages host-level iptables rules for VM internet access.
 // Appends per-VM rules to the kernel's built-in chains:
 //   - filter/FORWARD: allow traffic between each VM's veth and the host interface
--- a/internal/network/manager.go
+++ b/internal/network/manager.go
@@ -85,7 +85,7 @@ type Manager struct {
 	nextSlot  int   // next new slot (used when freeSlots is empty)
 
 	// Host-level nftables firewall (FORWARD + MASQUERADE + MSS clamping).
-	hostFW *HostFirewall
+	hostFW hostFirewallAPI
 
 	// TCP egress proxy — receives per-sandbox rule updates and cleanup.
 	egressProxy *EgressProxy
--- a/internal/network/manager.go
+++ b/internal/network/manager.go
@@ -447,9 +447,7 @@ func (m *Manager) ReattachVM(vmID, namespace, hostIP, macAddress string) error {
 		m.log.Warn().Err(err).Str("vm_id", vmID).Msg("reattach: in-namespace firewall handle not restored (existing rules still enforce traffic)")
 	}
 
-	// Do the fallible step before recording success: a future retry would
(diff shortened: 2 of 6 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define host firewall interface and mock for testing](https://stagereview.app/superserve-ai/sandbox/pull/64/chapters/1) |
| 2 | [Consult kernel state during slot allocation](https://stagereview.app/superserve-ai/sandbox/pull/64/chapters/2) |
| 3 | [Validate pool slots and handle phantom namespaces](https://stagereview.app/superserve-ai/sandbox/pull/64/chapters/3) |
| 4 | [Rebuild missing network namespaces on resume](https://stagereview.app/superserve-ai/sandbox/pull/64/chapters/4) |

<a href="https://stagereview.app/superserve-ai/sandbox/pull/64">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 161d15d2e406b0847bb474deeaef42b0af076206 on May 20, 2026 12:03am UTC.</sub>

pavitrabhalla:
## Review

Good set of targeted fixes — the three-part structure (allocator skips phantoms, pool validates before handing out slots, resume rebuilds missing netns) is coherent and addresses real crash-recovery failure modes.

**Things to address before merging:**

**`cleanup()` may nil-panic in the pool phantom test.** `newTestManager()` leaves `hostFW: nil`, and `TestPoolClaim_DiscardsPhantomReturnsNil` exercises `p.cleanup(slot)` on a phantom. If `cleanup` calls `hostFW.RemoveVM`, this will panic. Either add a nil guard in `cleanup` or confirm the code path doesn't reach it for un-registered VMs.

**`EnsureVMSlot` rebuild path has no test.** The most operationally important case — ns is missing, slot is rebuilt from scratch — isn't covered. The fast-path and invalid-namespace tests are there, but the rebuild branch is the whole point of this function.

**Fast path in `EnsureVMSlot` silently skips firewall re-add.** If `devices[vmID]` exists and the ns is present, the function returns immediately without calling `hostFW.AddVM`. After a firewall daemon restart that clears rules, this path would hand back a broken entry. Worth a comment at minimum explaining why this is safe (or not).

**Minor:** `claimSlotIndex` holds `m.mu` across `os.Stat` calls. Fine for current slot counts, but worth noting if this loop grows.

The test coverage for `claimSlotIndex` (fresh, LIFO, phantom skip, exhaustion) is solid. The `netnsDir` injection via package var is pragmatic — just don't mix with `t.Parallel()`.

pavitrabhalla:
The firewall fast-path fix (fd744c04) looks right — always calling `AddVM` before the devices insert is the correct behavior and matches the `ReattachVM` ordering.

One gap left: by making the fast path call `hostFW.AddVM`, `TestEnsureVMSlot_FastPathReturnsExisting` had to be dropped (nil hostFW in the test manager). That leaves `EnsureVMSlot` with zero tests for the paths that actually do work — the rebuild case (ns missing → `setupSlot`) and the construct-new-info case (ns exists but no device entry) are both uncovered. The only remaining test is the invalid namespace early-return.

Worth adding a mock `hostFW` (even a minimal stub struct) so the existing paths can be tested without requiring real kernel network setup.

meAmitPatil:
@pavitrabhalla Added fixes, tested on staging everything works.

nirnejak (approved):

PR #67, opened 2026-05-21
uffd: treat EAGAIN as transient, log mappings on out-of-range fault

meAmitPatil:
Two UFFD handler reliability fixes observed during the burst benchmark.

**poll()-based serveLoop.** Firecracker creates the UFFD fd as non-blocking; vmd tries to clear `O_NONBLOCK` post-handshake but silently ignores fcntl errors. When the clear doesn't take effect, `unix.Read` returns EAGAIN, which the old serveLoop treated as fatal — the handler died, the guest hung on its next page fault. Switching to `poll(POLLIN, 100ms) → read` makes the read independent of the fd's blocking mode and gives a clean cancellation path via the 100ms timeout. EAGAIN remains as a defensive safety net (counter + continue).

**Diagnostics on out-of-range fault.** The "page fault address outside all regions" error currently logs only the fault address. Without knowing what mappings were loaded at handshake, we can't tell whether the kernel sent a fault outside the registered range, our `findRegion` math is wrong, or mappings are corrupted. Now logs the full mappings list, the page-aligned fault address, and the page size on this error path.

**Observability split.** `EagainRetries` is split into `IoctlEagainRetries` (UFFDIO_COPY/ZEROPAGE EAGAIN — benign kernel queue drain) and `ReadEagainRetries` (read-path EAGAIN despite positive poll — should be zero in normal operation).

### DIFF
--- a/internal/vm/uffd/handler.go
+++ b/internal/vm/uffd/handler.go
@@ -66,15 +66,16 @@ type Config struct {
 }
 
 type Stats struct {
-	FaultsServed     atomic.Uint64
-	BytesServed      atomic.Uint64
-	CopyErrors       atomic.Uint64
-	UnknownEvents    atomic.Uint64
-	RemoveEvents     atomic.Uint64
-	EagainRetries    atomic.Uint64
-	Eexist           atomic.Uint64
-	PrefetchedPages  atomic.Uint64
-	PrefetchSkipped  atomic.Uint64 // EEXIST during prefetch (page already faulted by guest)
+	FaultsServed       atomic.Uint64
+	BytesServed        atomic.Uint64
+	CopyErrors         atomic.Uint64
+	UnknownEvents      atomic.Uint64
+	RemoveEvents       atomic.Uint64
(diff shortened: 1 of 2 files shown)


ghost:
Ready to review this PR? Stage has broken it down into 3 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Split EAGAIN stats into read and ioctl variants](https://stagereview.app/superserve-ai/sandbox/pull/67/chapters/1) |
| 2 | [Implement poll-based serveLoop for reliable UFFD reads](https://stagereview.app/superserve-ai/sandbox/pull/67/chapters/2) |
| 3 | [Enhance diagnostics for out-of-range page faults](https://stagereview.app/superserve-ai/sandbox/pull/67/chapters/3) |

<a href="https://stagereview.app/superserve-ai/sandbox/pull/67">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit cfbfcc2aa3ffd769348435b5e09462c38335cecb on May 21, 2026 1:11am UTC.</sub>

nirnejak (approved):

PR #89, opened 2026-06-02
proxy: add full-duplex /exec/ws WebSocket endpoint with stdin support

meAmitPatil:
Adds a WebSocket exec endpoint so clients can run a command and stream its output back over a single connection, with stdin support — the full-duplex counterpart to the SSE `/exec/stream`.

## Changes
- **boxd**: wire stdin for non-PTY processes — `SendInputRequest.eof`, an stdin pipe on the process, and `SendInput` writing to it / closing on EOF. (Previously a no-op for non-PTY.)
- **proxy**: new `/exec/ws` bridge — upgrades the connection and bridges to `ProcessService.Start` in non-PTY mode. Structured `stdout`/`stderr`/`exit` JSON frames downstream; binary stdin + text control (`stdin`/`stdin_close`/`signal`) upstream. Added to the boxd-port allowlist.
- **spec**: `/exec/ws` documented in `api/openapi.yaml`.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -412,6 +412,43 @@ paths:
         "500":
           $ref: "#/components/responses/InternalError"
 
+  /exec/ws:
+    servers:
+      - url: https://sandbox.superserve.ai
+        description: Shared host — send the sandbox ID in `X-Superserve-Sandbox-Id`.
+      - url: https://boxd-{sandbox_id}.sandbox.superserve.ai
+        description: Per-sandbox host.
+        variables:
+          sandbox_id:
+            default: ""
+            description: The sandbox ID.
+    get:
+      operationId: execWebSocket
+      tags: [Exec]
+      summary: Run a command over a WebSocket
+      description: |
+        Runs a command over a WebSocket, streaming output back and accepting
+        stdin over one connection.
+
+        Connect with two `Sec-WebSocket-Protocol` values: `superserve.exec.v1`
+        and `token.<access-token>`.
+
+        - First frame (text): an `ExecRequest` JSON object. Omitting `timeout_s`
+          runs with no per-command timeout.
+        - Client frames: a binary frame is stdin; a text frame is JSON control —
+          `{"type":"stdin","data":"..."}`, `{"type":"stdin_close"}`, or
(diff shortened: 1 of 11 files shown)


pavitrabhalla:
@codex review?

pavitrabhalla (commented):
Good implementation overall — the protocol design is clean, auth is consistent with the terminal bridge, and test coverage is solid. A few things to address:

**P1 — stdin pipe opens for all non-PTY execs, not just WebSocket**
`startPipes` now unconditionally creates a stdin pipe. The `/exec` and `/exec/stream` paths have no mechanism to send EOF, so any command that reads stdin (e.g. `cat`, scripts calling `sys.stdin.read()`) will block until the command timeout instead of seeing immediate EOF. The pipe should only be opened in the WebSocket path, or the pipe's write end should be closed immediately after `cmd.Start()` for non-WS invocations.

**Verify `Setpgid: true` is set for non-PTY commands**
`Signal` now sends to `-pid` (the process group) for non-PTY processes. The comment asserts Setpgid is already set, but it's not visible in this diff. If it isn't, `Kill(-pid, sig)` would signal the proxy's own process group — worth confirming explicitly.

**Flaky test timing**
`sendStart()` uses `time.Sleep(100ms)` to wait for the bridge to consume the start frame, and `TestSignal_TerminatesShellCommand` uses `time.Sleep(200ms)` before signaling. Both are timing-dependent and fragile in slow CI. A small channel or sync point would be more robust.

**Minor — undocumented UTF-8 limitation on `stdin` control frame**
The `{"type":"stdin","data":"..."}` control message carries stdin as a JSON string, which silently corrupts non-UTF-8 bytes. Worth a note in the docs/struct that binary stdin must use binary frames.

api/openapi.yaml
```diff
@@ -412,6 +412,48 @@ paths:
         "500":
           $ref: "#/components/responses/InternalError"
 
+  /exec/ws:
```
pavitrabhalla: can we rename this to `/exec/connect` ?

pavitrabhalla (approved):
One API naming suggestion. Otherwise approved

PR #91, opened 2026-06-03
observability: add Sentry error tracking and logging

pavitrabhalla:
## Summary

- Adds Sentry to the controlplane via `SENTRY_DSN` env var (no-op when unset)
- Wires `sentrygin` middleware into the router so HTTP handler panics are captured before our `ErrorHandler` recovers them
- Adds `internal/sentrylog.Writer`, a zerolog `LevelWriter` that forwards `warn`/`error`/`fatal` events to Sentry with all log fields (sandbox_id, team_id, etc.) as context
- Adds `sentrylog.RecoverAndCapture` and defers it in all three background goroutine loops (reaper, host-detector, build-supervisor) that previously had no panic recovery
- Enables structured logs (`EnableLogs: true`) for the Sentry logs feature

## Test plan

- [ ] Set `SENTRY_DSN` in deployment env (value in `.env.example`)
- [ ] Verify an error event appears in the Sentry dashboard on first 5xx or logged error
- [ ] Verify no behaviour change when `SENTRY_DSN` is unset (local dev)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,7 +8,7 @@ AWS_ACCESS_KEY_ID=
 AWS_SECRET_ACCESS_KEY=
 
 # Sentry (optional — leave unset to disable error reporting)
-SENTRY_DSN=https://4497bf1fc61a2388b355489982086ac7@o4511499764760576.ingest.us.sentry.io/4511499780620288
+SENTRY_DSN=
 
 # Control plane
 API_PORT=8080
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,9 @@ S3_BUCKET=your-sandbox-checkpoints-bucket
 AWS_ACCESS_KEY_ID=
 AWS_SECRET_ACCESS_KEY=
 
+# Sentry (optional — leave unset to disable error reporting)
+SENTRY_DSN=https://4497bf1fc61a2388b355489982086ac7@o4511499764760576.ingest.us.sentry.io/4511499780620288
+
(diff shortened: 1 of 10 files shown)


pavitrabhalla:
Should just need to set appropriate SENTRY_DSN in staging and prod after merging this.

meAmitPatil:
### Code review

Found 1 issue:

1. `RecoverAndCapture` is deferred at the goroutine-function level (above the `for` loop) and returns without re-panicking, so a single panic permanently kills the loop instead of crashing the process. The reaper, host-detector, and build-supervisor goroutines have no restart wrapper, so after one panic the process keeps serving HTTP and passing health checks while sandbox timeout-reaping / host detection / build supervision are silently dead. Before this PR an unrecovered panic crashed the process and the orchestrator restarted it. Recovering per-iteration inside the loop body (or re-panicking) would preserve the documented "runs until ctx is cancelled" / "errors do not stop the loop" invariants.

https://github.com/superserve-ai/sandbox/blob/657e02fbc60e766b08f35461242e2417aa084a5b/internal/sentrylog/writer.go#L68-L77

Call site (defer is above the loop, so the whole loop dies on panic):

https://github.com/superserve-ai/sandbox/blob/657e02fbc60e766b08f35461242e2417aa084a5b/internal/api/reaper.go#L54-L57

meAmitPatil (approved):

PR #96, opened 2026-06-05
sentrylog: errors-only + richer events, and recover goroutine panics

meAmitPatil:
Cleans up the Sentry signal: stop the noise, make remaining errors debuggable, and stop benign races from showing up as errors.

## 1. Only forward errors to Sentry, and make them debuggable
- **Drop warnings** — `sentrylog.Writer` forwards `error`+ only, not `warn`. 4xx request logs, auth-failed, bad-host, heartbeat retries, etc. stay in Cloud Logging instead of flooding Sentry. (5xx = our bug → Sentry; 4xx = expected client behavior → logs.)
- **Enrich each event** so it's debuggable instead of just a title:
  - the real error in the title (`"<msg>: <error>"`), not buried in a context
  - `caller` (file:line) preserved and promoted to a tag (was being stripped)
  - high-signal fields (`component`, `sandbox_id`, `build_id`, `template_id`, `host_id`) become searchable tags
  - fingerprint `[message, caller]` so one log site stays one issue even as the error text varies

## 2. Recover panics in background goroutines
An unrecovered panic in any goroutine crashes the whole process. Background loops were already protected via `RunSafe`; added `defer sentrylog.Recover(...)` (new helper) to the proxy WS pumps (exec/terminal), the egress per-conn handler, the build worker, the netpool refill loop, the rate-limit cleanup loop, the transport sweeper, and the async activity/mark-failed writes. A panic now becomes a Sentry event instead of a crash; existing cleanup defers still run first.

## 3. Quiet benign build-supervisor errors
Two error-level logs that were firing on benign races (verified against prod logs + the DB):
- **`finalize build` = "no rows in result set"** — a concurrent finalize: two supervisor ticks/instances both finalize the same build; the winner flips it to `ready`, the loser's `UPDATE … WHERE status IN ('building','snapshotting')` matches nothing. The build succeeds (confirmed in DB: rows end `ready`). Now treated as an idempotent no-op (`pgx.ErrNoRows` → debug).
- **`count in-flight builds failed` / `list active builds failed` / `reap stale builds failed` = "closed pool"** — the supervisor ticks once against the DB pool as it closes during a deploy. Now downgraded to debug only for `puddle.ErrClosedPool` (genuine failures still log Error).

No new dependencies, no go.mod change, no behavior change beyond the above. Build/vet/tests green across proxy, network, api, vm, supervisor.

### DIFF
--- a/internal/sentrylog/writer.go
+++ b/internal/sentrylog/writer.go
@@ -40,7 +40,8 @@ func (w *Writer) WriteLevel(level zerolog.Level, p []byte) (int, error) {
 	for k, v := range fields {
 		extras[k] = v
 	}
-	for _, k := range []string{"level", "time", "message", "msg"} {
+	// caller is promoted to a tag below, so drop it here to avoid duplication.
+	for _, k := range []string{"level", "time", "message", "msg", "caller"} {
 		delete(extras, k)
 	}
 
--- a/internal/sentrylog/writer.go
+++ b/internal/sentrylog/writer.go
@@ -10,18 +10,17 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-// Writer is a zerolog LevelWriter that forwards warn and error events to
-// Sentry. Use it inside a zerolog.MultiLevelWriter alongside the primary
-// output writer. It is a no-op when Sentry has not been initialized.
+// Writer forwards error-level zerolog events to Sentry; a no-op until Sentry is
+// initialized. Use it in a zerolog.MultiLevelWriter beside the primary output.
 type Writer struct{}
 
 // Write satisfies io.Writer; actual forwarding happens in WriteLevel.
 func (w *Writer) Write(p []byte) (int, error) { return len(p), nil }
 
-// WriteLevel parses the zerolog JSON event and forwards warn/error/fatal
-// levels to Sentry as captured messages with all log fields attached as extras.
+// WriteLevel forwards error/fatal zerolog events to Sentry with their fields;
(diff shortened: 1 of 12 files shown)


internal/sentrylog/writer.go
```diff
@@ -34,48 +33,55 @@ func (w *Writer) WriteLevel(level zerolog.Level, p []byte) (int, error) {
...
 	}
+	errStr, _ := fields["error"].(string)
+	caller, _ := fields["caller"].(string)
 
 	extras := make(map[string]interface{}, len(fields))
 	for k, v := range fields {
 		extras[k] = v
 	}
-	for _, k := range []string{"level", "time", "message", "msg", "caller"} {
+	for _, k := range []string{"level", "time", "message", "msg"} {
```
pavitrabhalla: `caller` is now extracted separately and promoted to a tag, but it's no longer deleted from `extras`, so it ends up in both the tag list and the `log` context blob — minor duplication. Adding `"caller"` back to this delete list would clean that up:

```go
for _, k := range []string{"level", "time", "message", "msg", "caller"} {
```

pavitrabhalla (approved):
Two clean, well-scoped improvements. Approving. One inline nit on `caller` duplication and a note on test coverage below.

pavitrabhalla:
**Test coverage note:** the `warn`→`error` filter change in `WriteLevel` doesn't appear to have a corresponding test update. If there's an existing test for `sentrylog.Writer`, worth adding an assertion that `warn`-level events are no longer forwarded.

PR #97, opened 2026-06-05
limits: raise sandbox/template defaults + 80% quota alerting

meAmitPatil:
## Limits
- **Templates** default 10 → **50** (code default; teams with `NULL` use it).
- **Sandboxes** default 50 → **100** via migration: new teams get 100, and existing teams *at the old default* are bumped — explicit overrides (e.g. 200) and the system team's exemption are left untouched.
- Per-instance resource caps (vCPU/RAM/disk) unchanged.

## 80% quota alerting
A small **background watcher** (off the request path) that posts a Slack alert when a team crosses 80% of a resource limit, so we can reach out proactively.

- **Detection off the hot path** — a 15-min ticker, *not* inline at create. Zero added create latency.
- **Source of truth reused** — reads `team.active_sandbox_count` (trigger-maintained) and a template `COUNT`; no new tracking.
- **Once per crossing, multi-instance safe** — a `quota_alert_state` row is claimed atomically (`INSERT … ON CONFLICT DO NOTHING`), so exactly one control-plane replica sends; the row is cleared when usage drops back under threshold so a re-crossing alerts again.
- **Retry on failure** — a failed Slack post releases the claim to retry next tick.
- **Generalized over quota types** — adding a third (storage, build minutes…) is a ~10-line `quotaCheck`, not a new pipeline.
- **No-op unless configured** — only runs when `SLACK_QUOTA_ALERT_WEBHOOK` is set.

Threshold uses integer math (`used*100 < limit*pct`) to be exact at the boundary. Unit-tested (`TestComputeQuotaAlerts`): default-vs-override limits, the 80% boundary, and the zero-limit edge.

Build/vet/tests green (api, supervisor); no new deps, no go.mod change. Migrations are additive.


internal/api/quota_watcher.go
```diff
@@ -0,0 +1,179 @@
...
+	return out
+}
+
+// StartQuotaWatcher periodically alerts on teams over the threshold, off the
+// request path. Blocks until ctx is cancelled.
+func StartQuotaWatcher(ctx context.Context, queries *db.Queries, notifier QuotaNotifier) {
+	log.Info().Int("threshold_pct", quotaAlertPct).Dur("interval", quotaWatchInterval).Msg("quota watcher started")
+	ticker := time.NewTicker(quotaWatchInterval)
+	defer ticker.Stop()
+	sentrylog.RunSafe("quota-watcher", func() { quotaWatchOnce(ctx, queries, notifier) })
```
pavitrabhalla: This initial run uses the bare `ctx` (no deadline), while every tick-driven run gets `quotaWatchTimeout`. Should be consistent:

```go
initCtx, cancel := context.WithTimeout(ctx, quotaWatchTimeout)
sentrylog.RunSafe("quota-watcher", func() { quotaWatchOnce(initCtx, queries, notifier) })
cancel()
```

internal/api/quota_watcher.go
```diff
@@ -0,0 +1,179 @@
...
+				Pct:      used * 100 / limit,
+			}
+		}
+	}
+	return out
+}
+
+// StartQuotaWatcher periodically alerts on teams over the threshold, off the
+// request path. Blocks until ctx is cancelled.
+func StartQuotaWatcher(ctx context.Context, queries *db.Queries, notifier QuotaNotifier) {
```
pavitrabhalla, line 89: `StartQuotaWatcher` is launched in a goroutine but has no outer `defer sentrylog.Recover`. `RunSafe` protects each `quotaWatchOnce` call, but a panic in the `for`/`select` scaffolding would crash the process. Consistent with the pattern in #96 to add `defer sentrylog.Recover("quota-watcher-loop")` at the top of this function.

internal/api/quota_watcher.go
```diff
@@ -0,0 +1,179 @@
...
+			continue // already alerted this episode
+		}
+		if err := notifier.Notify(ctx, alert); err != nil {
+			log.Error().Err(err).Str("team", k.team.String()).Str("resource", k.typ).Msg("quota watcher: notify failed; releasing claim")
+			_ = queries.ClearQuotaAlert(ctx, db.ClearQuotaAlertParams{TeamID: k.team, QuotaType: k.typ})
+		}
+	}
+
+	// Release teams that dropped back under threshold so a re-crossing alerts again.
+	states, err := queries.ListQuotaAlertState(ctx)
```
pavitrabhalla, line 139: Full table scan every 15 min — fine at current scale. Worth bounding if the table grows (e.g. `WHERE created_at > now() - interval '24h'`), though stale rows should stay rare given the clear-on-recovery logic.

supabase/migrations/20260605000002_quota_alert_state.sql
```diff
@@ -0,0 +1,9 @@
+-- De-dup state for quota alerts: a row means this team was already alerted for
+-- this resource (cleared when usage drops back under threshold). The atomic
+-- INSERT ... ON CONFLICT also makes exactly one replica send each alert.
+CREATE TABLE quota_alert_state (
+    team_id    uuid NOT NULL REFERENCES team(id) ON DELETE CASCADE,
+    quota_type text NOT NULL,
```
pavitrabhalla, line 6: `quota_type` is untyped `text` — no DB-level validation that it matches a valid quota check name. A typo in a future `quotaCheck.name` would create orphaned rows silently. Not a problem today with two types; just keep in mind if the set grows.

pavitrabhalla (commented):
Clean design — off hot path, claim-once atomics, self-clearing. A few notes below.

PR #98, opened 2026-06-06
sandbox: a failed resume must never brick a sandbox

meAmitPatil:
## What happened

A paused sandbox (`73517ad8…`) was permanently bricked by a *resume*. Root cause, traced end-to-end through control-plane logs, the vmd host, and the DB:

1. Resume restored the VM successfully.
2. `ActivateSandbox`'s `sandbox_active_interval` insert collided with a **stale open interval** (a prior transition left one dangling) → `idx_sai_one_open_per_sandbox` unique violation.
3. The handler's error path called `DestroyInstance(force=true)` → `cleanupRunDir` → `os.RemoveAll(rundir)`, which **deleted the per-VM `overlay.ext4`** — the writable disk.
4. The sandbox stayed `paused` in the DB but its disk was gone, so it 500'd on every resume forever, with nothing flagging it.

In short: a *bookkeeping* insert failing caused the handler to *force-destroy a healthy resumed VM* and lose its data.

## The fix — three independent layers (any one stops it)

**1. Idempotent interval opens.** `ON CONFLICT (sandbox_id) WHERE ended_at IS NULL DO NOTHING` on both open sites (`ActivateSandbox` CTE + `OpenSandboxActiveInterval`). A stale interval can no longer fail a resume — this just makes the code obey the contract already documented on `OpenSandboxActiveInterval`.

**2. Non-destructive resume compensation.** Once the VM is up, a finalize failure (egress rules **or** DB activate) now re-pauses via `PauseInstance`→`FinalizePause` — preserving the overlay — instead of `DestroyInstance`. Verified against vmd internals that *no* `pauseAndRevert` path reaches `cleanupRunDir`: VM-in-map → re-pause; snapshot fails → `handleVMError` (no rundir delete) → mark failed; VM gone → mark failed. The worst case is a `failed` sandbox with its disk intact — never data loss. *Prefer leak over loss.*

**3. Orphan-interval sweep.** The reaper tick closes intervals left open on settled non-active sandboxes (5-min settle window, `end_reason` mirrors terminal state). Kills the precondition and WAU skew; off the request path.

## Latency

Hot paths (create / resume / pause) are unaffected. `ON CONFLICT` is the same index probe that already runs; create's `ActivateSandbox` is async; the sweep is a background tick. Only a *failed* resume's 500 is slightly delayed (snapshot vs. kill), bounded by `vmdTimeout`.

## Tests

- Rewrote the two resume-failure unit tests to assert **pause-not-destroy** (`DestroyInstance` calls = 0).
- New integration tests (real Postgres): resume **succeeds** with a pre-seeded stale interval (reproduces the incident), and the sweep closes a backdated orphan while sparing fresh transitions.

## Possible future work (not needed now)

Layers 1–2 make this brick impossible, so there is no immediate follow-up. *Only if* we ever actually observe overlays going missing through some other path (out-of-band cleanup, disk faults) is it worth surfacing unrecoverable sandboxes instead of silent 500s — mark `failed` on a permanent restore error, an alert-first reconciler overlay check, and the reconciler's first tests. Not worth building speculatively: auto-`failed` from a filesystem heuristic risks mass-failing healthy sandboxes, so it should land only when there's a real signal.

### DIFF
--- a/db/queries/sandbox_active_intervals.sql
+++ b/db/queries/sandbox_active_intervals.sql
@@ -1,12 +1,11 @@
 -- name: OpenSandboxActiveInterval :exec
 -- Opens an interval row when a sandbox transitions into 'active' (created or
--- resumed). The partial unique index sandbox_active_interval(sandbox_id) WHERE
--- ended_at IS NULL guarantees at most one open row per sandbox at any time;
--- in the unlikely event a prior open row was left dangling, the caller should
--- treat the unique-violation as recoverable (sandbox is already considered
--- running) rather than a hard error.
+-- resumed). A dangling open row from a prior transition is recoverable, so
+-- ON CONFLICT makes this a no-op rather than a unique-violation the caller
+-- would have to special-case.
 INSERT INTO sandbox_active_interval (sandbox_id, team_id, actor_id, started_at)
-VALUES ($1, $2, $3, now());
+VALUES ($1, $2, $3, now())
+ON CONFLICT (sandbox_id) WHERE ended_at IS NULL DO NOTHING;
 
 -- name: CloseSandboxActiveInterval :exec
 -- Closes the currently-open interval for a sandbox. The WHERE ended_at IS NULL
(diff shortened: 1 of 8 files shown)


pavitrabhalla:
**Review notes:**

**`SizeBytes: 0` hardcoded in `FinalizePause`** (handlers.go) — if this field feeds billing, quotas, or storage analytics, every reverted resume will silently record a 0-byte snapshot. The normal pause path presumably passes the real size from `PauseInstance`'s output. Either thread the actual size through or document why 0 is correct here.

**State machine: does `FinalizePause` accept `resuming` as prior state?** — when `pauseAndRevert` fires, the sandbox is in `resuming`, not `pausing`. If `FinalizePause` has a `WHERE status = 'pausing'` precondition it will fail every time here, the new snapshot path won't be recorded, and we always fall back to `revertToPaused()`. Worth a quick confirmation that the path actually works as described.

**Shared timeout across `PauseInstance` + `FinalizePause`** — both ops share the same `pctx`/`vmdTimeout` deadline. If `PauseInstance` runs long, `FinalizePause` gets a near-expired context, fails, and the new snapshot path is lost. Consider separate timeouts.

**`time.Sleep(50ms)` in integration test** — flaky under load. A short polling loop on the DB status would be more reliable.

**`ON CONFLICT DO NOTHING` on `ActivateSandbox` during creation** — in practice a new sandbox has no prior interval, but if it somehow does, the old row is silently reused. Low risk, just worth a comment noting creation is excluded from this concern.

PR #101, opened 2026-06-10
network: config-driven egress blocklist for proxy and host firewall

pavitrabhalla:
## What

Adds an optional, config-driven egress denylist for sandbox traffic, enforced at two points:

- **Egress proxy** (`internal/network/proxy.go`): connections are checked against the denylist (Host header / SNI plus the original destination IP) before per-sandbox rules are evaluated. Hostnames that re-resolve to a denylisted IP are also rejected at dial time.
- **Host firewall** (`internal/network/hostfw.go`): configured destination ports are dropped (TCP and UDP) at the top of the FORWARD chain for all `veth+` traffic. Only ports 80/443 are redirected through the egress proxy, so other ports are enforced here.

## Configuration

All list contents (feeds, pinned domains/CIDRs, ports) come from an operator-supplied file pointed at by `VMD_EGRESS_BLOCKLIST_CONFIG` — see `deploy/egress-blocklist.example.yaml` for the schema. When the env var is unset the feature is disabled and there is no behavior change.

- Feeds are plain-text (one entry per line, `#` comments): domains, IPs, `IP:port` pairs, or CIDRs.
- Refreshed on a configurable interval with lock-free atomic snapshot swaps.
- A failed refresh keeps the last good snapshot; the merged list is persisted to disk so restarts are seeded before the first fetch.
- Domain matching walks parent labels (`a.b.example.com` matches a listed `example.com`).

## Testing

- New tests in `internal/network/blocklist_test.go`: feed parsing (incl. `IP:port` and CIDR forms), parent-label matching, denylist-precedes-sandbox-allow, last-good-on-feed-failure, config validation, and state-file seeding.
- `GOOS=linux go build ./...` and `go vet` clean; the network package is Linux-only (nftables) so the full suite runs in CI.

### DIFF
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -250,8 +250,24 @@ func main() {
 
 	lc := newLifecycle(log)
 
+	// ---- Egress blocklist (optional) ----
+	// VMD_EGRESS_BLOCKLIST_CONFIG points at the operator-supplied config
+	// (feeds, pinned domains/CIDRs, blocked ports). Unset = no global
+	// blocklist.
+	var blocklist *network.Blocklist
+	var netMgrOpts []network.ManagerOption
+	if path := os.Getenv("VMD_EGRESS_BLOCKLIST_CONFIG"); path != "" {
+		blCfg, err := network.LoadBlocklistConfig(path)
+		if err != nil {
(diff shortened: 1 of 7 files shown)


pavitrabhalla:
Superseded by #102.

PR #102, opened 2026-06-10
network: config-driven egress blocklist for proxy and host firewall

pavitrabhalla:
## What

Adds an optional, config-driven egress denylist for sandbox traffic, enforced at two points:

- **Egress proxy** (`internal/network/proxy.go`): connections are checked against the denylist (Host header / SNI plus the original destination IP) before per-sandbox rules are evaluated. Hostnames that re-resolve to a denylisted IP are also rejected at dial time.
- **Host firewall** (`internal/network/hostfw.go`): configured destination ports are dropped (TCP and UDP) at the top of the FORWARD chain for all `veth+` traffic. Only ports 80/443 are redirected through the egress proxy, so other ports are enforced here.

## Configuration

All list contents (feeds, pinned domains/CIDRs, ports) come from an operator-supplied file pointed at by `VMD_EGRESS_BLOCKLIST_CONFIG` — see `deploy/egress-blocklist.example.yaml` for the schema. When the env var is unset the feature is disabled and there is no behavior change.

- Feeds are plain-text (one entry per line, `#` comments): domains, IPs, `IP:port` pairs, or CIDRs.
- Refreshed on a configurable interval with lock-free atomic snapshot swaps.
- A failed refresh keeps the last good snapshot; the merged list is persisted to disk so restarts are seeded before the first fetch.
- Domain matching walks parent labels (`a.b.example.com` matches a listed `example.com`).

## Testing

- New tests in `internal/network/blocklist_test.go`: feed parsing (incl. `IP:port` and CIDR forms), parent-label matching, denylist-precedes-sandbox-allow, last-good-on-feed-failure, config validation, and state-file seeding.
- `GOOS=linux go build ./...` and `go vet` clean; the network package is Linux-only (nftables) so the full suite runs in CI.

### DIFF
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -255,7 +255,10 @@ func main() {
 	// (feeds, pinned domains/CIDRs, blocked ports). Unset = no global
 	// blocklist.
 	var blocklist *network.Blocklist
-	var netMgrOpts []network.ManagerOption
+	// vmd is the daemon, so it owns and reconciles the shared egress port
+	// chain even when no ports are configured (so disabling the feature
+	// clears stale drops). template-builder must not pass this.
+	netMgrOpts := []network.ManagerOption{network.WithEgressPortChainOwner()}
 	if path := os.Getenv("VMD_EGRESS_BLOCKLIST_CONFIG"); path != "" {
 		blCfg, err := network.LoadBlocklistConfig(path)
 		if err != nil {
@@ -328,6 +331,11 @@ func main() {
 			log.Fatal().Err(err).Msg("failed to install host egress block table")
 		}
 		blocklist.SetCIDRSink(hostBlock.UpdateCIDRs)
+		// Seed the host drop set from the state-seeded snapshot now, before
+		// the first feed fetch (which may block up to feedFetchTimeout per
+		// feed). Otherwise seeded CIDRs go unenforced on non-proxied ports
+		// during the startup window.
+		hostBlock.UpdateCIDRs(blocklist.CIDRs())
 		lc.addCloser("host egress block", func(_ context.Context) error { return hostBlock.Close() })
 		lc.start("egress blocklist", func() error { return blocklist.Start(ctx) })
 	}
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -313,6 +313,14 @@ func main() {
 	netMgr.SetEgressProxy(egressProxy)
 	if blocklist != nil {
 		egressProxy.SetBlocklist(blocklist)
+		// Mirror IP/CIDR entries into a host-level nftables drop set so they
+		// are enforced on every port, not just the proxied web ports.
+		hostBlock, err := network.NewHostEgressBlock(log)
+		if err != nil {
+			log.Fatal().Err(err).Msg("failed to install host egress block table")
+		}
+		blocklist.SetCIDRSink(hostBlock.UpdateCIDRs)
+		lc.addCloser("host egress block", func(_ context.Context) error { return hostBlock.Close() })
 		lc.start("egress blocklist", func() error { return blocklist.Start(ctx) })
 	}
 	lc.start("egress proxy", func() error { return egressProxy.Start(ctx) })
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -264,6 +264,14 @@ func main() {
 		blocklist = network.NewBlocklist(blCfg, log)
 		netMgrOpts = append(netMgrOpts, network.WithBlockedEgressPorts(blCfg.BlockedEgressPorts))
 		log.Info().Str("path", path).Int("feeds", len(blCfg.DomainFeeds)).Int("blocked_ports", len(blCfg.BlockedEgressPorts)).Msg("egress blocklist configured")
+	} else {
+		// Feature disabled — drop any host egress-block table a previous run
+		// installed so its CIDR drops stop affecting sandbox egress. The
+		// per-VM port-drop chain self-heals: installHostFirewall always runs
+		// and ClearChain empties it when no ports are configured.
+		if err := network.RemoveHostEgressBlock(); err != nil {
+			log.Debug().Err(err).Msg("removing stale host egress block table")
+		}
 	}
 
 	// ---- Network manager + host firewall ----
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -250,8 +250,24 @@ func main() {
 
 	lc := newLifecycle(log)
 
+	// ---- Egress blocklist (optional) ----
+	// VMD_EGRESS_BLOCKLIST_CONFIG points at the operator-supplied config
+	// (feeds, pinned domains/CIDRs, blocked ports). Unset = no global
+	// blocklist.
+	var blocklist *network.Blocklist
+	var netMgrOpts []network.ManagerOption
+	if path := os.Getenv("VMD_EGRESS_BLOCKLIST_CONFIG"); path != "" {
+		blCfg, err := network.LoadBlocklistConfig(path)
+		if err != nil {
+			log.Fatal().Err(err).Str("path", path).Msg("failed to load egress blocklist config")
+		}
+		blocklist = network.NewBlocklist(blCfg, log)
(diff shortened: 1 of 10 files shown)


pavitrabhalla:
Thanks — both were real. Fixed in the latest commit:

**P1: partial feed failure dropped entries.** Correct: `refresh()` rebuilt from an empty builder and swapped in whenever *any* feed succeeded, so a single failing feed lost all its entries (the keep-previous guard only fired on total failure). Each feed's last successful fetch is now cached and reused when that feed fails; the previous snapshot is kept wholesale only when no feed yields data, fresh or cached. Regression test: `TestRefreshKeepsFailedFeedFromCache`.

**P1: IP/CIDR entries not enforced off the web ports.** Correct: `Blocked()` only ran in the egress proxy, which receives only the REDIRECTed 80/443 traffic, so a blocklisted IP was reachable on any other port. CIDRs are now mirrored into a host nftables drop table (`sandbox-egress-block`) that drops forwarded traffic to those prefixes on every port and protocol, refreshed on each blocklist fetch. It's a single host-level set (scales to large feeds) and only ever DROPs, so it can't break egress or accept otherwise-denied traffic. The example config wording was corrected to match the actual enforcement per entry type.

pavitrabhalla:
Latest commit addresses the three P2s on `321352b5d0`. (The two P1s at the top of the thread are from the earlier `a037303234` review and were already fixed in `321352b5d0` — caching for the partial-feed failure, and the host nftables drop set for IP/CIDR enforcement off the web ports.)

**P2 — seed host CIDR sink on feed outage.** Correct, and a real consequence of the new host-drop dependency. On a cold start with a state-seeded snapshot but all feeds down, `refresh()` returned before calling the sink, so the host drop set stayed empty and seeded IPs were reachable on non-proxied ports. The early-return path now pushes the current snapshot's CIDRs to the sink. Test: `TestCIDRSinkFiresWhenAllFeedsDownWithSeededState`.

**P2 — reconcile removed port drops.** Correct: drops were inserted directly into `FORWARD` and never removed, so a port deleted from `blocked_egress_ports` kept blocking after restart. Moved them into a vmd-owned chain `SANDBOX_EGRESS_PORTS`, flushed and rebuilt via `ClearChain` on each start, entered by a single jump for `veth+` traffic. Removed ports now disappear from the live ruleset.

**P2 — limit host CIDR drops to sandbox packets.** Correct: the forward-hook drop matched any forwarded IPv4 packet. It now also matches `ip saddr` in `vmIPRange` — VM packets are SNAT'd to their hostIP (within that range) before reaching the host FORWARD hook, so Docker/kube/VPN traffic on a cohabitated host to the same CIDRs is no longer collaterally dropped. This matches the existing `veth+`/`vmIPRange` assumption documented in hostfw.go.

pavitrabhalla:
Latest commit addresses the two new findings on \`565dd31260\`, plus documents the SNAT invariant.

**P1 — preserve seeded coverage when the first refresh partially fails.** Correct, and a real edge of the caching fix: \`NewBlocklist\` seeds the snapshot from the state file but not \`feedCache\`, so on the first refresh after restart, a feed that fails before it's ever cached would lose its state-seeded entries while another feed's success swapped in the shrunken snapshot. \`refresh()\` now unions the previous snapshot whenever any feed is unavailable (degraded round), so coverage never shrinks on failure — this also subsumes the all-feeds-down case. State is persisted only on a fully successful round, so the on-disk seed never shrinks either; a later clean refresh rebuilds without the union so legitimate removals still apply. Test: \`TestRefreshPreservesSeededStateOnPartialFirstFailure\`.

**P2 — remove the host table when the feature is disabled.** Correct: unsetting \`VMD_EGRESS_BLOCKLIST_CONFIG\` after a prior run left the \`sandbox-egress-block\` table dropping egress with stale CIDRs. Startup now calls \`RemoveHostEgressBlock\` on the disabled path. (The port-drop chain already self-heals — \`installHostFirewall\` always runs and \`ClearChain\` empties it when no ports are configured.)

Also documented the topology invariant the host CIDR drop relies on — the \`ip saddr vmIPRange\` match only scopes correctly because SNAT happens in-namespace, so VM packets reach the host FORWARD hook sourced from their hostIP. Cross-linked the comment between \`firewall.go\` (the SNAT rule) and \`egressblock.go\` (the saddr match) so moving SNAT to the host can't silently un-scope the drop.

The older unresolved inline threads (on \`a037303234\` / \`321352b5d0\`) are the P1s/P2s already fixed in earlier commits — GitHub keeps them pinned to their original lines.

pavitrabhalla:
Both addressed in \`a745a56\`.

**P1 — template-builder flushing vmd's port drops.** Correct and important: \`SANDBOX_EGRESS_PORTS\` is a single shared chain, and template-builder's \`NewManager\` ran \`ClearChain\` on it, wiping the daemon's drops for the duration of a build. The chain is now reconciled only by its owner — vmd passes \`WithEgressPortChainOwner()\`, and \`installHostFirewall\` skips the chain entirely otherwise, so template-builder (and any future auxiliary manager) leaves it untouched. vmd owns it unconditionally so disabling the feature still clears stale drops.

**P2 — seed host CIDRs before the first feed fetch.** Correct: \`blocklist.Start\` doesn't call the sink until the initial fetch loop returns, up to \`feedFetchTimeout\` per feed, leaving state-seeded CIDRs unenforced on non-proxied ports during that window. vmd now replays \`blocklist.CIDRs()\` into \`hostBlock.UpdateCIDRs\` immediately after wiring the sink, before \`Start\`.

pavitrabhalla:
Both addressed in \`4e59fde\`.

**P2 — web ports in \`blocked_egress_ports\` are a no-op.** Correct: \`veth+\` TCP/80 and TCP/443 are REDIRECTed to the egress proxy in PREROUTING, which runs before the FORWARD drop, so dropping those ports there does nothing and a sandbox allow rule could still permit them. Rather than silently ignore, \`LoadBlocklistConfig\` now rejects 80/443 with a message pointing operators at the domain blocklist / proxy path, which is where web traffic is actually governed. Documented in the example config too.

**P2 — overlapping CIDRs fail the nft flush.** Correct, and it would have silently disabled host CIDR enforcement: aggregated feeds routinely carry a prefix plus one it already covers (e.g. \`203.0.113.0/24\` + \`203.0.113.7/32\`), nftables rejects overlapping interval elements, and the rejected flush leaves \`blocked_v4\` at its previous/empty state. \`cidrsToElements\` now coalesces — drops any prefix contained in another, plus exact duplicates — before emitting elements. Since CIDR prefixes are always either disjoint or nested, dropping contained ones removes every overlap. This lives in the shared helper, so the per-VM user-rule path gets the same fix. Tests: \`TestCoalescePrefixes\`, \`TestCidrsToElementsCoalesces\`.

pavitrabhalla:
Fixed in \`fa98097\`.

**P2 — oversized feed silently truncated.** Correct: \`io.LimitReader(resp.Body, maxFeedBytes)\` returns the first 16 MiB with no truncation signal, so a feed grown past the cap counted as a successful fetch, overwrote \`feedCache\`, and could persist a partial state file — dropping every entry past the cutoff. \`fetchFeed\` now reads \`maxBytes+1\` and fails the feed if the result exceeds the cap, so \`refresh()\` treats it as a failed feed and falls back to the last-good copy (and doesn't persist). Test: \`TestOversizedFeedFailsAndKeepsLastGood\`. (Local-file feeds are read whole and uncapped — those are operator-owned, so the cap is intentionally HTTP-only.)

pavitrabhalla:
Fixed in \`6169a79\`.

**P2 — feed scanner errors treated as success.** Correct: \`addFeedText\` ran a \`bufio.Scanner\` (1 MiB line cap) but never checked \`scanner.Err()\`, so a feed with an over-long line stopped the scan at \`ErrTooLong\` and the partial parse counted as a successful fetch — overwriting the cache and persisting truncated state, dropping everything after the long line. \`addFeedText\` now returns the scanner error, and \`refresh()\` parses each feed into an isolated builder first: a parse error marks the round degraded, skips the feed (no cache or state update), and the existing previous-snapshot union retains last-good coverage. Tests: \`TestAddFeedTextRejectsOverlongLine\`, \`TestFeedParseErrorKeepsLastGood\`.

meAmitPatil (approved):
LGTM

PR #104, opened 2026-06-12
limits: lower default per-team caps to 10 sandboxes and 5 templates

meAmitPatil:
Drops the per-team defaults from 100 sandboxes / 50 templates to 10 / 5 to shrink the abuse blast radius. Existing teams on the old default are reduced; explicit overrides are preserved. The system team's quotas are reconciled at controlplane start. Any per-team exemptions are applied out-of-band after this lands.

### DIFF
--- a/cmd/controlplane/main.go
+++ b/cmd/controlplane/main.go
@@ -234,15 +234,19 @@ func run() error {
 	return nil
 }
 
-// reconcileSystemTeamQuota lifts the system team's max_sandboxes above any
-// realistic cap so the sandbox_quota_on_insert trigger never rejects it.
+// reconcileSystemTeamQuota lifts the system team's max_sandboxes and
+// max_templates so the quota trigger and template-count check never block it.
 func reconcileSystemTeamQuota(ctx context.Context, pool *pgxpool.Pool, systemTeamID string) {
 	if systemTeamID == "" {
 		return
 	}
 	tag, err := pool.Exec(ctx,
-		`UPDATE team SET max_sandboxes = $1
-		 WHERE id = $2 AND (max_sandboxes IS NULL OR max_sandboxes < $1)`,
+		`UPDATE team
+		 SET max_sandboxes = $1,
+		     max_templates = $1
+		 WHERE id = $2
+		   AND (max_sandboxes IS NULL OR max_sandboxes < $1
+		     OR max_templates IS NULL OR max_templates < $1)`,
 		math.MaxInt32, systemTeamID,
 	)
 	if err != nil {
--- a/internal/api/handlers_template.go
+++ b/internal/api/handlers_template.go
@@ -103,7 +103,7 @@ const (
 	defaultMaxVcpu      = 2
 	defaultMaxMemoryMib = 2048
 	defaultMaxDiskMib   = 8192
-	defaultMaxTemplates = 50
+	defaultMaxTemplates = 5
 
 	defaultVcpu     = 1
 	defaultMemoryMi = 1024
--- a/internal/api/handlers_template.go
+++ b/internal/api/handlers_template.go
@@ -103,7 +103,7 @@ const (
 	defaultMaxVcpu      = 2
 	defaultMaxMemoryMib = 2048
 	defaultMaxDiskMib   = 8192
-	defaultMaxTemplates = 5
+	defaultMaxTemplates = 10
 
 	defaultVcpu     = 1
 	defaultMemoryMi = 1024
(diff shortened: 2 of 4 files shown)


supabase/migrations/20260612000001_lower_default_limits.sql
```diff
@@ -0,0 +1,3 @@
+-- max_sandboxes default 100 -> 10. Templates default in code.
+ALTER TABLE team ALTER COLUMN max_sandboxes SET DEFAULT 10;
+UPDATE team SET max_sandboxes = 10 WHERE max_sandboxes = 100;
```
pavitrabhalla: Two things on this line:

1. **max_templates not updated.** The migration lowers `max_sandboxes` for existing rows but doesn't touch `max_templates`. If teams have `max_templates = 50` stored explicitly in the DB, the Go constant change in `handlers_template.go` won't affect them. Should add:
```sql
UPDATE team SET max_templates = 5 WHERE max_templates = 50;
```
And if `max_templates` has a DB-level `DEFAULT 50`, also `ALTER TABLE team ALTER COLUMN max_templates SET DEFAULT 5;`

2. **WHERE clause assumption.** `WHERE max_sandboxes = 100` can't distinguish the old schema default from a deliberate override set to exactly 100. Low risk given current team counts, but worth a SQL comment documenting the assumption.

supabase/migrations/20260612000001_lower_default_limits.sql
```diff
@@ -0,0 +1,3 @@
+-- max_sandboxes default 100 -> 10. Templates default in code.
+ALTER TABLE team ALTER COLUMN max_sandboxes SET DEFAULT 10;
```
pavitrabhalla: Would suggest 20 instead of 10 for the sandbox default — 10 is quite tight and may cause friction for legitimate new teams before they can request a quota bump.

cmd/controlplane/main.go
```diff
@@ -234,15 +234,19 @@ func run() error {
...
 	}
 	tag, err := pool.Exec(ctx,
-		`UPDATE team SET max_sandboxes = $1
-		 WHERE id = $2 AND (max_sandboxes IS NULL OR max_sandboxes < $1)`,
+		`UPDATE team
+		 SET max_sandboxes = $1,
+		     max_templates = $1
+		 WHERE id = $2
+		   AND (max_sandboxes IS NULL OR max_sandboxes < $1
+		     OR max_templates IS NULL OR max_templates < $1)`,
```
pavitrabhalla, line 249: The `OR` in the WHERE clause means if either field triggers, both columns get overwritten. Since both are being set to `math.MaxInt32` this is idempotent and harmless, but unconditionally updating both columns would be simpler and more honest — the function already guards on `systemTeamID == ""` and only runs at startup.

internal/api/quota_watcher_test.go
```diff
@@ -13,14 +13,14 @@ func ptrInt32(v int32) *int32 { return &v }
...
 	zeroLimit := uuid.New()    // max 0 -> must not alert or panic
 
 	rows := []db.ListTeamQuotaUsageRow{
 		{ID: atThreshold, Name: "at", ActiveSandboxCount: 80, MaxSandboxes: 100},
 		{ID: below, Name: "below", ActiveSandboxCount: 79, MaxSandboxes: 100, TemplateCount: 1},
-		{ID: tmplDefault, Name: "tdefault", MaxSandboxes: 100, TemplateCount: 40}, // MaxTemplates nil -> default
+		{ID: tmplDefault, Name: "tdefault", MaxSandboxes: 100, TemplateCount: 4}, // MaxTemplates nil -> default
 		{ID: tmplOverride, Name: "toverride", MaxSandboxes: 100, MaxTemplates: ptrInt32(200), TemplateCount: 10},
 		{ID: zeroLimit, Name: "zero", ActiveSandboxCount: 5, MaxSandboxes: 0},
 	}
```
pavitrabhalla, line 26: This test only validates the alert-threshold arithmetic (4/5 = 80%). Consider adding a test that new teams with NULL `max_sandboxes` / `max_templates` actually receive the new defaults (10 and 5 respectively) when the fallback path is exercised.

internal/api/handlers_template.go
```diff
@@ -103,7 +103,7 @@ const (
 	defaultMaxVcpu      = 2
 	defaultMaxMemoryMib = 2048
 	defaultMaxDiskMib   = 8192
-	defaultMaxTemplates = 50
+	defaultMaxTemplates = 5
```
pavitrabhalla, line 107: Would suggest 10 instead of 5 here — pairing with a sandbox default of 20 gives new teams a 1:2 template-to-sandbox ratio which feels more workable without significantly expanding the blast radius.

PR #105, opened 2026-06-14
secrets [1/6]: database schema and queries

meAmitPatil:
**Stack 1 of 6** — credential-vault + network-log feature, split from #84 for progressive review.

Schema layer: migrations (RLS-enabled), sqlc-generated queries for the credential vault, sandbox secret bindings, revocations, and net_flow. Pure foundation — no behavior change, nothing reads these tables yet.

- Base: `main`
- Builds + sqlc-in-sync + migration-order independently.
- The feature is dormant until stack 6. Full stack was verified end-to-end on staging as a unit.


db/queries/secrets.sql
```diff
@@ -0,0 +1,126 @@
...
+
+-- name: SoftDeleteSecretByName :one
+UPDATE secret
+SET deleted_at = now(), updated_at = now()
+WHERE team_id = $1 AND name = $2 AND deleted_at IS NULL
+RETURNING *;
+
+-- name: AddSandboxSecret :exec
+INSERT INTO sandbox_secret (sandbox_id, secret_id, env_key)
+VALUES ($1, $2, $3);
```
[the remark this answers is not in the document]
meAmitPatil: Addressed in #111.

db/queries/secrets.sql
```diff
@@ -0,0 +1,126 @@
...
+
+-- name: ListProxyAuditEvents :many
+-- Request rows for the unified per-sandbox network log, filtered by an optional
+-- time window (before/since); before doubles as the pagination cursor. Merged
+-- with net_flow connection rows in the handler.
+SELECT * FROM proxy_audit
+WHERE sandbox_id = sqlc.arg('sandbox_id')
+  AND (sqlc.narg('before')::timestamptz IS NULL OR ts < sqlc.narg('before')::timestamptz)
+  AND (sqlc.narg('since')::timestamptz IS NULL OR ts >= sqlc.narg('since')::timestamptz)
+ORDER BY ts DESC
```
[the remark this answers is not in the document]
meAmitPatil: fixed in #111.

db/queries/secrets.sql
```diff
@@ -0,0 +1,126 @@
...
+WHERE team_id = $1 AND name = $2 AND deleted_at IS NULL;
+
+-- name: GetSecretByID :one
+SELECT * FROM secret
+WHERE id = $1 AND team_id = $2 AND deleted_at IS NULL;
+
+-- name: GetSecretByIDForDecrypt :one
+-- Daemon decrypt path. Team-scoped; excludes soft-deleted.
+SELECT * FROM secret
+WHERE id = $1 AND team_id = $2 AND deleted_at IS NULL;
```
pavitrabhalla, line 21: GetSecretByIDForDecrypt has identical SQL to GetSecretByID. If the decrypt path needs distinct behavior in a later stack (audit logging, caching), document that intent here. Otherwise consolidate into one query to avoid confusing callers in stacks 2–5.

db/queries/secrets.sql
```diff
@@ -0,0 +1,126 @@
...
+-- time window (before/since); before doubles as the pagination cursor. Merged
+-- with net_flow connection rows in the handler.
+SELECT * FROM proxy_audit
+WHERE sandbox_id = sqlc.arg('sandbox_id')
+  AND (sqlc.narg('before')::timestamptz IS NULL OR ts < sqlc.narg('before')::timestamptz)
+  AND (sqlc.narg('since')::timestamptz IS NULL OR ts >= sqlc.narg('since')::timestamptz)
+ORDER BY ts DESC
+LIMIT sqlc.arg('row_limit');
+
+-- name: ListAuditForSecret :many
```
pavitrabhalla, line 112: Using bare $3/$4/$5 causes sqlc to generate Column3/Column4/Column5 field names in ListAuditForSecretParams, which are opaque at call sites. All other time-range queries in this PR use sqlc.arg()/sqlc.narg() — suggest doing the same here (e.g. sqlc.narg("cursor_id"), sqlc.arg("status_min"), sqlc.arg("status_max")) and replacing the $3=0 sentinel with a nullable cursor consistent with before/since.

db/queries/secrets.sql
```diff
@@ -0,0 +1,126 @@
...
+    encrypted_dek = $4,
+    kek_id = $5,
+    updated_at = now()
+WHERE id = $1 AND team_id = $2 AND deleted_at IS NULL
+RETURNING *;
+
+-- name: TouchSecretLastUsed :exec
+UPDATE secret
+SET last_used_at = now()
+WHERE id = $1;
```
pavitrabhalla, line 41: TouchSecretLastUsed has no team_id guard, unlike every other mutating secret query. If intentional on the hot path (caller already verified ownership), add a comment here so no one "fixes" it without understanding the tradeoff.

supabase/migrations/20260608000006_net_flow.sql
```diff
@@ -0,0 +1,32 @@
...
+-- for every sandbox (not just those with secrets attached). It is the
+-- connection-level counterpart to proxy_audit's request-level rows; the two
+-- union into the per-sandbox network view.
+
+CREATE TABLE IF NOT EXISTS net_flow (
+    id            bigserial PRIMARY KEY,
+    ts            timestamptz NOT NULL DEFAULT now(),
+    team_id       uuid NOT NULL,
+    sandbox_id    uuid NOT NULL,
+    -- 'http' (port 80), 'tls' (port 443), or 'other' (raw TCP).
```
pavitrabhalla, line 11: sandbox_id (and team_id) have no FK constraint. If intentional — audit rows should survive sandbox/team hard-deletes — please add a comment explaining this, otherwise someone will add the FK in a follow-up thinking it was an oversight.

supabase/migrations/20260608000001_sandbox_secrets.sql
```diff
@@ -0,0 +1,83 @@
...
+-- DEK wrapped by a Cloud KMS KEK). Plaintext is never persisted.
+--
+-- Idempotent: re-running on a database that already has these objects is a no-op.
+
+-- Team-level knobs:
+--   credential_store_kind: 'builtin' (encrypted here) or 'external' (BYO vault).
+--   unmatched_host_policy: 'passthrough' or 'deny' for hosts no credential covers.
+ALTER TABLE team
+    ADD COLUMN IF NOT EXISTS credential_store_kind   text   NOT NULL DEFAULT 'builtin'
+        CHECK (credential_store_kind IN ('builtin', 'external')),
```
pavitrabhalla, line 12: credential_store_config is nullable with no CHECK constraint. If credential_store_kind = 'external' requires config, consider: CHECK (credential_store_kind = 'builtin' OR credential_store_config IS NOT NULL).

meAmitPatil:
@pavitrabhalla not actual bugs, addressed all as additional guards in #111

PR #112, opened 2026-06-15
Some initial changes to the readme and makefile to get things running

awille-ai:
Making some changes to the makefile and README instructions to match the current state of things


Makefile
```diff
@@ -58,11 +58,32 @@ generate-proto:
...
+db-up:
+	docker run --name $(DB_CONTAINER) \
+		-e POSTGRES_USER=postgres \
+		-e POSTGRES_PASSWORD=postgres \
+		-e POSTGRES_DB=sandbox_test \
+		-p $(DB_PORT):5432 \
+		-d postgres:16
+
+db-down:
+	docker rm -f $(DB_CONTAINER)
```
pavitrabhalla: `db-reset` duplicates the `docker run` body from `db-up`. If the image, port mapping, or env vars ever change, both blocks need updating in sync.

Simpler as:
```makefile
db-reset: db-down db-up
```

This also sidesteps the `-docker rm -f` prefix (error-suppression needed because a missing container exits non-zero) — if you make `db-down` use `|| true` instead of relying on `docker rm -f` to succeed, `db-reset` composes cleanly with no special-casing needed.

Makefile
```diff
@@ -93,7 +114,7 @@ test:
 test-short:
 	go test ./... -short -count=1
 
-test-integration:
+test-integration: db-reset
```
pavitrabhalla: `docker run -d` returns as soon as the container process starts, not when Postgres is accepting connections. `go test` then runs immediately — the test harness connects once with a 30s context but has no retry loop, so a refused connection fails the whole suite.

CI avoids this via `--health-cmd "pg_isready -U postgres"` on its service container. For the Makefile path, a simple poll target between `db-reset` and the test run closes the gap:
```makefile
db-wait:
	@until docker exec $(DB_CONTAINER) pg_isready -U postgres -q; do sleep 1; done

test-integration: db-reset db-wait
	go test -tags integration ./internal/integration/ -v -count=1 -timeout 10m
```

README.md
```diff
@@ -39,22 +41,90 @@ VMD uses Firecracker to launch microVMs. Snapshot pools pre-boot VMs so sandboxe
...
+### Local development: control plane only
 
-# Apply migrations
-make migrate-up
+```bash
+export DATABASE_URL='postgres://postgres:postgres@localhost:5432/sandbox_test?sslmode=disable'
+export ALLOW_EPHEMERAL_SEED=1
 
-# Run control plane
 make run-controlplane
```
pavitrabhalla, line 97: Missing migration step. The controlplane does not apply migrations on startup — the only thing that runs `supabase/migrations/` automatically is the integration test harness (`applyMigrations` in `TestMain`). Starting the controlplane against a fresh DB here will hit "relation does not exist" immediately.

Suggest adding `make migrate-local` (or equivalent) to this section, and a corresponding Makefile target:
```makefile
migrate-local:
	for f in $$(ls supabase/migrations/*.sql | sort); do psql "$(DATABASE_URL)" -f $$f; done
```

Then:
```bash
make db-up
make db-wait        # wait for postgres to accept connections
make migrate-local  # apply supabase/migrations/ before first run
make run-controlplane
```

pavitrabhalla (approved):
Just go through my (AI) review comments once, and feel free to discard. Squash and merge when ready

PR #114, opened 2026-06-16
files: download a sandbox directory as a zip via ?format=zip

mohamedsobhi777:
## What

Adds directory→ZIP download to the sandbox data plane (boxd + edge proxy) so the console can pull a whole directory in one action. This is the boxd + proxy half of SS-66. **SDKs are unchanged.**

## Design

`GET /files?path=…` gains an opt-in `&format=zip` flag; the **server** decides:

- file → streamed as-is (unchanged bytes)
- directory + `format=zip` → streamed as a ZIP
- directory **without** the flag → still `400` (preserves the SDK's `files.read()` / `readText()` contract — they never send the flag)

The flag is what lets the console add folder download without silently changing published SDK behavior.

## Changes

**boxd (`cmd/boxd/main.go`)**
- `handleFileDownload`: on a directory with `format=zip`, stream a ZIP via `archive/zip`. The directory's contents are attacker-controlled (untrusted sandbox code) and can change mid-walk, so the walk is hardened:
  - the whole walk and every file open go through an **`os.Root`** anchored at the directory — `os.Root` errors on any path that resolves outside it, so nothing escapes the tree even via a symlink swapped in mid-walk
  - **only directories and regular files** are archived — symlinks, FIFOs, sockets and device nodes are skipped before any open (a named pipe would otherwise block the download forever)
  - files are opened with **`O_NOFOLLOW | O_NONBLOCK`** and the fd is re-`stat`'d before copying, closing the TOCTOU window (a regular-file→symlink/FIFO swap between walk and open is caught and skipped)
  - `safePath` re-applied to every descendant for the blocklist (`/proc`, `/sys`, `/dev`, `/sbin/init`, the boxd binary) — `os.Root` doesn't exclude those on its own
  - entries are relative to the directory, prefixed with its base name (archive expands to `<base>/…`); empty dirs preserved
  - per-file errors are logged and skipped; constant memory, no temp file; `Content-Disposition` filename sanitized (strips `"` / control bytes)
- **Build-tag split (enabler):** `setHostname` (the Linux-only `syscall.Sethostname`) moved into `hostname_linux.go` / `hostname_other.go`, so `go test ./cmd/boxd/` builds and runs on non-Linux dev machines. Behavior on Linux is identical.

**proxy (`internal/proxy/files.go`)**
- CORS block now sets `Access-Control-Expose-Headers: Content-Disposition` so the cross-origin console can read the archive filename (`<dir>.zip`). It is not a CORS-safelisted response header, so without this it is invisible to JS even though boxd sends it. The `format=zip` query and the streamed response already pass through unchanged (`FlushInterval: -1`).

## Test plan

- [x] `cmd/boxd/files_test.go` (new, table-driven, `-race`): file unchanged · directory-without-flag still 400 · directory→valid zip with expected relative paths · symlinks skipped · **FIFO/special files skipped (no hang)** · **escaping symlink not archived (no content leak)** · empty dir → valid zip · file+flag returned as-is · blocklisted paths rejected · filename sanitized
- [x] `internal/proxy/files_test.go`: preflight exposes `Content-Disposition` · `format=zip` forwards to boxd and the response streams back
- [x] `go test -race ./cmd/boxd/ ./internal/proxy/` green; whole module compiles for `GOOS=linux`
- [ ] End-to-end on staging after deploy: download a real sandbox directory, confirm the zip extracts to the expected tree (can't be exercised locally — no Firecracker on dev)

## Review feedback (@awille-ai)

- ✅ **Non-regular files** (blocking) — FIFOs / sockets / device nodes are now skipped before opening (`41a242b`); FIFO regression test added (it previously hung the suite in `syscall.Open`).
- ✅ **Symlink TOCTOU / no-follow** (blocking) — the walk and all opens go through `os.Root`, with `O_NOFOLLOW` / `O_NONBLOCK` and an fd re-`stat`; symlink-escape test added.

## Notes

- Pairs with the console PR **superserve-ai/superserve#221** (the consumer). The two **can deploy in any order**: the console degrades gracefully (a directory download shows a "not available yet" message; single-file downloads keep working) until this ships, after which directory→zip is live. SDKs are unchanged.


awille-ai:
Blocking edge cases for directory ZIP safety:

1. Non-regular files should be skipped before opening. `serveDirAsZip` currently treats anything that is not a directory or symlink as a file, then opens and copies it. A FIFO inside the requested directory can block the download indefinitely, and device/socket/special files should not be archived. Please restrict ZIP entries to directories and regular files only, and add tests for FIFO/special-file entries.

2. Symlink skipping is not race-safe on a live sandbox filesystem. `WalkDir` can observe a regular file, but sandbox code can replace it with a symlink before `os.Open(p)`, and `os.Open` follows symlinks. Since this feature’s safety depends on not following symlinks or escaping into forbidden paths, please use no-follow open semantics plus a post-open regular-file check, or a descriptor-relative traversal approach.

I’d treat these as blockers before enabling whole-directory downloads from untrusted sandboxes.

mohamedsobhi777:
Both addressed in `41a242b`:

**1. Non-regular files.** `serveDirAsZip` now archives only directories and regular files — symlinks, FIFOs, sockets and device nodes are skipped before any open, so a named pipe can't block the download. Added a FIFO test (it timed the suite out before the fix — blocked in `syscall.Open` — and now it's skipped and the download completes).

**2. Symlink race / no-follow.** The whole walk and every file open now go through an `os.Root` anchored at the requested directory, so no path can resolve outside it — including one reached via a symlink swapped in mid-walk (`os.Root` errors on escape). Files are opened with `O_NOFOLLOW | O_NONBLOCK` and the fd is re-`stat`'d before copying, so a regular-file→symlink (ELOOP) or →FIFO (non-regular) swap between `WalkDir` and the open is caught and skipped. Added a symlink-escape test (a symlink whose target is outside the dir is not archived and its content doesn't leak).

I went with `os.Root` rather than `openat2(RESOLVE_NO_SYMLINKS)` so the path stays cross-platform and still unit-tests on macOS. One residual: `os.Root` confines intermediate components to the tree but may still follow an *in-root* symlink for an intermediate dir under a race — that can't escape the requested directory, so it's not an information-disclosure path, and the walk-time skip handles the non-race case. The `safePath` blocklist is still re-applied per descendant (`os.Root` doesn't exclude `/proc`/device files on its own).

All green with `go test -race`.

awille-ai:
One more
In the sandbox PR, handleFileDownload still uses os.Stat(path) to decide whether the requested path is a directory, and os.Stat follows symlinks. Then serveDirAsZip calls os.OpenRoot(dirPath). Go’s docs say OpenRoot “follows symbolic links in the directory name,” so a requested path like /home/user/export where export -> /etc can make the archive root become /etc, even though the user requested a path under /home/user.

Relevant code:

superserve-ai/sandbox, cmd/boxd/main.go: os.Stat(path) in handleFileDownload
superserve-ai/sandbox, cmd/boxd/main.go: os.OpenRoot(dirPath) in serveDirAsZip
Go docs: https://pkg.go.dev/os#OpenRoot

Fix direction:

Reject top-level symlink directories for ZIP downloads.
Add a regression test: create dir/link -> outsideDir, request link?format=zip, and assert it does not archive the outside target.
For stronger race safety, avoid Lstat then OpenRoot as the only defense, since the path can change between calls. Prefer opening relative to a rooted parent, or use Linux no-follow directory-open semantics for the final component if production only needs Linux guarantees.

cmd/boxd/main.go
```diff
@@ -876,6 +886,218 @@ func handleFileDownload(w http.ResponseWriter, r *http.Request, path string) {
...
+	// component. handleFileDownload's os.Stat and a plain os.OpenRoot(dirPath)
+	// both follow a symlink at dirPath, so a request for "/home/user/export"
+	// where export -> /etc (or /, /proc) would make the archive root the symlink
+	// target — escaping the requested directory and bypassing safePath's
+	// blocklist. Anchor at the parent and reach the leaf through the parent's
+	// Root: a no-follow Lstat refuses a symlinked leaf, and Root.OpenRoot
+	// resolves it relative to the parent's open fd, so there is no
+	// Lstat-then-open TOCTOU window and a leaf swapped for an escaping symlink is
+	// rejected rather than followed.
+	parent, err := os.OpenRoot(filepath.Dir(dirPath))
```
meAmitPatil: 🔴 **High — confirmed: intermediate symlink escapes the tree + bypasses the blocklist.** Only the *leaf* (`base`) gets the no-follow `Lstat` guard. `os.OpenRoot(filepath.Dir(dirPath))` resolves the *parent* normally and follows any symlink in an intermediate component, so the `os.Root` ends up anchored outside the requested tree.

Reproduced on this branch: sandbox code makes `a -> /` (or `-> /tmp/outside`), then a request for `path=…/a/realsub&format=zip` archived `realsub/leak.txt` with outside content. With `a -> /` and `path=…/a/proc/1` the archive walks `/proc/1` — the descendant check at L1034 rebuilds paths from the literal `dirPath` string so it never matches `/proc`. The comment's "nothing … can escape the requested directory" and "blocklist enforced below the top-level argument" don't hold for intermediate components.

meAmitPatil (commented):
Symlink hardening is the point of this PR, but it's incomplete: a directory listing/archive can still be steered outside the requested tree and past safePath's /proc,/sys,/dev blocklist. Two confirmed escapes below (🔴) — both reproduced with a local test on this branch — plus a few smaller notes.

cmd/boxd/main.go
```diff
@@ -876,6 +886,218 @@ func handleFileDownload(w http.ResponseWriter, r *http.Request, path string) {
...
+// exactly as it does for ?format=zip, so no proxy change is needed.
+//
+// Unlike serveDirAsZip this never opens or streams file contents; it reports only
+// names, sizes, and modification times. So it deliberately skips the zip path's
+// symlink/TOCTOU hardening: the worst a symlink can do is have the listing follow
+// it like `ls` would, exposing entry names the sandbox occupant can already read
+// via /exec. safePath has already gated the requested path against the blocklist,
+// and the access token scopes the request to this one sandbox.
+func serveDirAsJSON(w http.ResponseWriter, dirPath string) {
+	entries, err := os.ReadDir(dirPath)
```
meAmitPatil: 🔴 **High — confirmed: no symlink guard, asymmetric with the zip path.** `handleFileDownload` reaches here via `os.Stat` (follows symlinks) and `os.ReadDir(dirPath)` follows too, so a symlinked directory `export -> /etc` (or `/proc`, `/`) is listed wholesale. Reproduced: `path=<symlink-to-outside>&format=json` returned the target dir's entries. The zip path explicitly refuses a symlinked top-level dir (L994-1005 + `TestFileDownload_Zip_TopLevelSymlinkRejected`); this path has no equivalent check and no test. The "already readable via /exec" justification understates it — it bypasses safePath's blocklist, which exists precisely so this HTTP surface can't enumerate `/proc,/sys,/dev` regardless of /exec.

cmd/boxd/main.go
```diff
@@ -876,6 +886,218 @@ func handleFileDownload(w http.ResponseWriter, r *http.Request, path string) {
...
+		if walkErr != nil {
+			// Unreadable entry mid-walk: log and skip, keep the archive going.
+			log.Printf("files: zip walk skip %q: %v", rel, walkErr)
+			return nil
+		}
+		if rel == "." {
+			return nil // the directory itself; entries carry the base prefix
+		}
+		// Enforce the blocklist on descendants, not just the top-level path.
+		if _, err := safePath(filepath.Join(dirPath, rel)); err != nil {
```
meAmitPatil: 🟠 Note (ties into the L985 escape): this re-check reconstructs paths from the literal `dirPath` string via `filepath.Join`, so it only enforces the blocklist when `dirPath` is its own real location. If an ancestor component was a symlink (the L985 issue), `filepath.Join(dirPath, rel)` still reads `…/a/proc/…` and never matches `/proc`, so the walk archives the real target unchecked.

cmd/boxd/main.go
```diff
@@ -876,6 +886,218 @@ func handleFileDownload(w http.ResponseWriter, r *http.Request, path string) {
...
+//     time-of-check/time-of-use window.
+//   - safePath is re-applied to every descendant so the blocklist (/proc, /sys,
+//     /dev, the boxd binary, ...) is enforced below the top-level argument too.
+//
+// os.OpenRoot validates the directory before the first byte, so a bad request
+// still gets a clean error. Once the archive starts streaming the HTTP status is
+// committed, so per-file errors are logged and skipped rather than aborting the
+// whole download. Entries are written as the walk proceeds — constant memory,
+// no temp file.
+func serveDirAsZip(w http.ResponseWriter, dirPath string) {
```
meAmitPatil: 🟠 Medium — no entry-count / byte / time cap, and boxd runs with `WriteTimeout: 0`. Untrusted sandbox code can stage millions of small files or a few multi-GB files; one `format=zip` then streams indefinitely and pins a proxy connection. Uploads already use `MaxBytesReader`; this download path has no analogous bound.

cmd/boxd/main.go
```diff
@@ -876,6 +886,218 @@ func handleFileDownload(w http.ResponseWriter, r *http.Request, path string) {
...
+// /files allow-list via ?format=json — the proxy forwards the query untouched,
+// exactly as it does for ?format=zip, so no proxy change is needed.
+//
+// Unlike serveDirAsZip this never opens or streams file contents; it reports only
+// names, sizes, and modification times. So it deliberately skips the zip path's
+// symlink/TOCTOU hardening: the worst a symlink can do is have the listing follow
+// it like `ls` would, exposing entry names the sandbox occupant can already read
+// via /exec. safePath has already gated the requested path against the blocklist,
+// and the access token scopes the request to this one sandbox.
+func serveDirAsJSON(w http.ResponseWriter, dirPath string) {
```
meAmitPatil, line 1000: 🟡 Cleanup: this `os.ReadDir` + per-entry `Info()`-with-degrade loop duplicates `FilesystemService.ListDir`'s logic with a slightly different field set. Worth a shared entry-collection helper so listing semantics (hidden entries, concurrent-removal degrade) don't drift between the two surfaces.

Also minor: the JSON error paths here use `http.Error`, which forces `Content-Type: text/plain` even though the body is JSON — inconsistent with the success path and `storageFullResponse`. A `writeJSONError(w, code, msg)` helper would fix the content-type once and stop the `{"error":…}` shape being hand-rolled at ~5 new sites.

meAmitPatil:
@mohamedsobhi777 Feel free to address the ones you think are important.

cmd/boxd/main.go
```diff
@@ -860,7 +899,21 @@ func handleFileDownload(w http.ResponseWriter, r *http.Request, path string) {
...
+		// A directory has no single byte stream, so it is only returned when the
+		// caller explicitly opts in to a representation via ?format=:
+		//   zip  → an archive of its contents      (serveDirAsZip)
+		//   json → a one-level listing of its names (serveDirAsJSON)
+		// Without a format flag we keep the original 400, so the SDK's
+		// files.read()/readText() (which never send one) still throw on a
+		// directory, exactly as before.
+		switch r.URL.Query().Get("format") {
+		case "zip":
+			serveDirAsZip(w, path)
```
awille-ai: 🔴 High — this directory branch is now hardened, but the adjacent single-file branch below still uses `os.Stat(path)` followed by `os.Open(path)`, both of which follow sandbox-controlled symlinks. A hostile sandbox can create something like `/home/user/export -> /dev/zero` or `/home/user/export -> /proc/...`; the lexical `safePath` check passes because the requested string is benign, but the HTTP download then bypasses the `/proc`/`/sys`/`/dev` blocklist or can stream a special file indefinitely. Please give raw file downloads the same treatment as ZIP file entries (`O_NOFOLLOW`, fd `Stat`, regular-file-only, blocklist after symlink resolution) and add tests for symlink-to-blocklist plus FIFO/device/special-file targets.

awille-ai (changes requested):
Security review focused on hostile sandbox download behavior.

The latest head addresses several earlier issues: ZIP skips non-regular entries, rejects top-level symlink directories, resolves intermediate symlink/blocklist cases, shares JSON listing behavior, and caps ZIP output. I still would not ship this as-is because the single-file download path remains symlink-following and special-file-following while the PR is adding adjacent directory download surfaces that explicitly rely on the same `safePath` blocklist model.

The fix should make file downloads follow the same policy as ZIP file entries: do not follow sandbox-controlled symlinks into `/proc`, `/sys`, `/dev`, the boxd binary, etc.; reject or skip non-regular files; and add regression tests for symlink-to-blocklist and FIFO/device/special-file single-file downloads. Ideally the blocklist/symlink policy is centralized so `Stat`, `ListDir`, JSON listing, ZIP, and raw file download do not drift again.

internal/api/handlers.go
```diff
@@ -1747,6 +1747,64 @@ func (h *Handlers) ExecSandbox(c *gin.Context) {
...
+	sandbox := h.loadActiveOrResumeSandbox(c)
+	if sandbox == nil {
+		return
+	}
+
+	path := c.Query("path")
+	if path == "" {
+		path = "/"
+	}
+	if !strings.HasPrefix(path, "/") || strings.Contains(path, "..") {
```
pavitrabhalla: `strings.Contains(path, "..")` rejects valid paths where `..` appears as a substring in a filename (e.g. `/project/hello..world`). The error message says "free of '..' segments" but the check is more restrictive than that. Safer: `filepath.Clean(path)` first, then check that no component equals `".."`. Defense-in-depth saves you here (boxd re-checks), but the rejection is misleading.

cmd/boxd/main.go
```diff
@@ -876,6 +929,314 @@ func handleFileDownload(w http.ResponseWriter, r *http.Request, path string) {
...
+	defer root.Close()
+	w.Header().Set("Content-Type", "application/zip")
+	w.Header().Set("Content-Disposition", fmt.Sprintf(`attachment; filename="%s"`, zipDownloadFilename(base)))
+
+	// Bound the archive so attacker-staged content can't stream unbounded and
+	// pin a proxy connection (boxd runs with WriteTimeout: 0): cappedWriter caps
+	// total output bytes; entryCount caps the entry count in the walk below.
+	cw := &cappedWriter{w: w, limit: maxZipBytes}
+	zw := zip.NewWriter(cw)
+	defer zw.Close()
```
pavitrabhalla, line 1250: When `maxZipBytes` is hit mid-file, `cappedWriter` returns `errZipLimit` on all subsequent writes, including the ones from `defer zw.Close()`. That means the end-of-central-directory record never gets written and the resulting byte stream is not a valid zip — most unzip clients will reject it outright. The entry-cap path (`maxZipEntries`) does not have this problem because `zw.Close()` runs normally after `fs.SkipAll`. Is a silently corrupt archive acceptable, or should the byte cap terminate the connection instead?

proto/boxdpb/boxd.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.36.11
-// 	protoc        v7.34.1
+// 	protoc        v5.29.3
```
pavitrabhalla, line 4: `protoc` version regressed from `v7.34.1` to `v5.29.3` here and in `vmd.pb.go`. The generated file now reflects whoever ran `protoc` last on a different local install. Will cause spurious re-diffs the next time someone regenerates on the other version. Worth pinning `protoc` in the Makefile/CI alongside `protoc-gen-go`.

api/openapi.yaml
```diff
@@ -412,6 +412,49 @@ paths:
...
+      operationId: listSandboxFiles
+      tags: [Files]
+      summary: List a directory inside a sandbox
+      description: |
+        Returns a one-level listing of a directory inside the sandbox. Listing
+        is metadata served through the control plane (boxd's
+        FilesystemService.ListDir), so it works on every sandbox regardless of
+        its boxd version. A paused sandbox is resumed automatically before the
+        listing and stays active afterward. `modified_unix` is 0 for entries on
+        sandboxes whose boxd predates that field.
```
meAmitPatil: Can you avoid using our internal architectural terms for e.g. `boxd`.  The desc goes in our public API Reference.

meAmitPatil (approved):
LGTM

awille-ai:
Non-blocking follow-ups from this pass — I don't see a new high-severity blocker in the current head.

1. `GET /sandboxes/{sandbox_id}/files` currently defaults `path` to `/`, and the OpenAPI description says the same. But boxd `safePath` still rejects `/` (`cannot operate on root directory`), so an omitted `path` looks like it will return 400. Either default this public API to `/home/user` (probably the product default) or update the docs/handler to avoid advertising a default that cannot succeed.

2. Directory listings are now capped in boxd, which is the right operational guardrail, but the gRPC/API surface drops the `truncated` signal and returns only `entries`. That means API clients and the console cannot tell "complete listing" from "first N entries". Not blocking for this PR if you're comfortable with the current behavior, but I'd prefer carrying `truncated` through `ListDirResponse`, the HTTP response, and OpenAPI so callers can render a "showing first N" state instead of silently presenting partial data.

awille-ai (commented):
Please make sure to rebase this, your PR is behind main by quite a bit.
I left one non-blocking PR comment covering two API-contract follow-ups:

GET /sandboxes/{sandbox_id}/files defaults path to /, but boxd rejects root, so omitted path likely returns 400.
List responses are capped, but the truncated signal is not carried through the gRPC/API/OpenAPI surface, so clients cannot tell partial from complete listings.

awille-ai (approved):

PR #115, opened 2026-06-16
Adding billing changes

awille-ai:
https://www.notion.so/rayai/Sandbox-Metrics-and-Billing-Architecture-381743ab873380518423c563c653aac8


pavitrabhalla (commented):
Overall looks solid — immutability model is correct, feature flag layering is clean, and test coverage is thorough. A few things worth addressing below.

db/queries/sandbox_active_intervals.sql
```diff
@@ -3,8 +3,26 @@
...
+    SELECT t.id, t.team_id, sqlc.arg(actor_id), now()
+    FROM target t
+    ON CONFLICT (sandbox_id) WHERE ended_at IS NULL DO NOTHING
+    RETURNING sandbox_id
+)
+INSERT INTO sandbox_compute_billing_interval (
+    sandbox_id, team_id, vcpu_count, memory_mib, started_at
+)
+SELECT t.id, t.team_id, t.vcpu_count, t.memory_mib, now()
+FROM target t
```
pavitrabhalla: **Bug:** The billing interval insert reads from `target` directly, not from `opened_active`. If a competing call already has an open active interval, `opened_active` returns 0 rows (DO NOTHING), but this insert still fires from `target` — so you can get a billing interval without a corresponding active interval opening. The billing unique index saves you from a duplicate row, but the logic is inconsistent. Fix: `FROM opened_active oa JOIN target t ON t.id = oa.sandbox_id`.

db/queries/billing.sql
```diff
@@ -0,0 +1,352 @@
...
+    usage.vcpu_seconds,
+    usage.memory_mib_seconds,
+    usage.storage_mib_seconds
+FROM usage
+ON CONFLICT (team_id, period_start, period_end) DO UPDATE
+SET vcpu_seconds = EXCLUDED.vcpu_seconds,
+    memory_mib_seconds = EXCLUDED.memory_mib_seconds,
+    storage_mib_seconds = EXCLUDED.storage_mib_seconds,
+    updated_at = now()
+WHERE team_billing_usage.finalized_at IS NULL
```
pavitrabhalla: **Silent no-op on finalized rows.** When the row exists and is finalized, the `WHERE finalized_at IS NULL` predicate prevents the DO UPDATE from running, so Postgres returns 0 rows. The caller gets `pgx.ErrNoRows` but can't distinguish "row was never created" from "row is finalized and immutable". Callers that drive billing ops should check the row count or treat `ErrNoRows` here as a distinct error so operators can surface the right signal.

supabase/migrations/20260615000001_sandbox_billing_usage.sql
```diff
@@ -0,0 +1,313 @@
...
+    finalized_at  timestamptz,
+    created_at    timestamptz NOT NULL DEFAULT now(),
+    updated_at    timestamptz NOT NULL DEFAULT now(),
+
+    PRIMARY KEY (team_id, period_start, period_end),
+    CONSTRAINT team_billing_period_valid CHECK (period_end > period_start),
+    CONSTRAINT team_billing_period_status_valid CHECK (
+        status IN ('open', 'validating', 'blocked', 'approved', 'exported', 'finalized')
+    ),
+    CONSTRAINT team_billing_period_block_fields CHECK (
```
pavitrabhalla, line 266: **Minor:** `team_billing_period_block_fields` only checks `blocked_at`, so `blocked_reason` can be NULL when `status = 'blocked'`. `BlockTeamBillingPeriod` always sets it in practice, but the schema doesn't enforce it. Either extend the constraint to include `blocked_reason IS NOT NULL` or make the column `NOT NULL DEFAULT ''`.

db/queries/billing.sql
```diff
@@ -0,0 +1,352 @@
...
+  AND team_billing_usage.exported_at IS NULL
+RETURNING *;
+
+-- name: UpsertTeamBillingUsageHour :one
```
pavitrabhalla, line 155: **Minor:** When `billing_hourly_rollups` is disabled, the `usage` CTE returns 0 rows, the INSERT produces nothing, and Go gets `pgx.ErrNoRows`. Tested and safe, but treating 'flag off' as an error is odd. Consider making this `:exec` or `:execrows` so callers don't have to silently swallow a fake error on the disabled path.

supabase/migrations/20260615000001_sandbox_billing_usage.sql
awille-ai: fixed: Blocking is restricted to pre-export states.

db/queries/billing.sql
awille-ai: I did not change the minor billing_hourly_rollups disabled path to :execrows; it is safe as-is and changing it would alter the sqlc method signature.

db/queries/sandbox_active_intervals.sql
awille-ai: Billing interval insert now reads from opened_active, so billing compute intervals only open when the active interval actually opened.

db/queries/billing.sql
awille-ai: fixed: Existing exported/finalized usage rows are returned as immutable rows instead of becoming an indistinguishable silent no-op.

pavitrabhalla (approved):

PR #117, opened 2026-06-17
adding a rollup job

awille-ai:
Adds durable hourly billing rollups with a scalable worker queue, backfill support, retry handling, and tenant-scoped billing usage aggregation. Also introduces PAYG pricing and credit tables so billing calculations can use versioned pricing data and customer credits safely.
https://linear.app/superserve/issue/SS-55

### DIFF
--- a/internal/billing/rollup.go
+++ b/internal/billing/rollup.go
@@ -1,0 +1,547 @@
+package billing
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/rs/zerolog/log"
+
+	"github.com/superserve-ai/sandbox/internal/db"
+)
+
+const (
+	defaultHourlyRollupSchedulerInterval     = 5 * time.Minute
+	defaultHourlyRollupWorkerPoll            = 10 * time.Second
+	defaultHourlyRollupLookbackHours         = 24
+	defaultHourlyRollupBackfillLookbackHours = 90 * 24
+	defaultHourlyRollupBackfillBatchHours    = 24
+	defaultHourlyRollupLeaseDuration         = 2 * time.Minute
+	defaultHourlyRollupLockDuration          = 5 * time.Minute
+	defaultHourlyRollupBatchSize             = 100
+	defaultHourlyRollupMaxAttempts           = 5
+	defaultHourlyRollupWorkers               = 1
+)
+
+type HourlyRollupConfig struct {
+	SchedulerInterval     time.Duration
+	WorkerPoll            time.Duration
+	LookbackHours         int
+	BackfillLookbackHours int
+	BackfillBatchHours    int
+	LeaseDuration         time.Duration
+	LockDuration          time.Duration
+	BatchSize             int
+	MaxAttempts           int
+	Workers               int
+}
+
+type rollupJob struct {
+	ID           uuid.UUID
+	TeamID       uuid.UUID
+	HourStart    time.Time
+	HourEnd      time.Time
+	AttemptCount int
+}
+
+func DefaultHourlyRollupConfig() HourlyRollupConfig {
+	return HourlyRollupConfig{
+		SchedulerInterval:     durationFromEnv("BILLING_HOURLY_ROLLUP_INTERVAL", defaultHourlyRollupSchedulerInterval),
+		WorkerPoll:            durationFromEnv("BILLING_HOURLY_ROLLUP_WORKER_POLL", defaultHourlyRollupWorkerPoll),
+		LookbackHours:         intFromEnv("BILLING_HOURLY_ROLLUP_LOOKBACK_HOURS", defaultHourlyRollupLookbackHours),
+		BackfillLookbackHours: intFromEnv("BILLING_HOURLY_ROLLUP_BACKFILL_LOOKBACK_HOURS", defaultHourlyRollupBackfillLookbackHours),
+		BackfillBatchHours:    intFromEnv("BILLING_HOURLY_ROLLUP_BACKFILL_BATCH_HOURS", defaultHourlyRollupBackfillBatchHours),
+		LeaseDuration:         durationFromEnv("BILLING_HOURLY_ROLLUP_LEASE_DURATION", defaultHourlyRollupLeaseDuration),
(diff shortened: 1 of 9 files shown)


internal/billing/rollup.go
```diff
@@ -0,0 +1,547 @@
...
+	if cfg.LockDuration <= 0 {
+		cfg.LockDuration = defaultHourlyRollupLockDuration
+	}
+	if cfg.BatchSize <= 0 {
+		cfg.BatchSize = defaultHourlyRollupBatchSize
+	}
+	if cfg.MaxAttempts <= 0 {
+		cfg.MaxAttempts = defaultHourlyRollupMaxAttempts
+	}
+	if cfg.Workers < 0 {
```
pavitrabhalla: `Workers` uses `< 0` here but all other fields use `<= 0`. Setting `BILLING_HOURLY_ROLLUP_WORKERS=0` will start the scheduler with no workers, silently letting the queue grow. Should be `<= 0` maybe?

internal/billing/rollup.go
```diff
@@ -0,0 +1,547 @@
...
+	for _, hourStart := range hours {
+		enqueued, err := enqueueHour(ctx, pool, hourStart, hourStart.Add(time.Hour), false)
+		if err != nil {
+			return total, err
+		}
+		total += enqueued
+	}
+
+	if err := advanceBackfillCursor(ctx, pool, hours[0], hours[len(hours)-1].Add(time.Hour)); err != nil {
+		return total, err
```
pavitrabhalla, line 329: If `advanceBackfillCursor` fails here, `total` (already-enqueued jobs) is returned without error surfacing to the caller's log. Next tick re-enqueues the same hours because the cursor didn't advance — idempotent but creates silent duplicate work. Worth a `log.Warn` on this path at minimum.

internal/billing/rollup.go
```diff
@@ -0,0 +1,547 @@
...
+		return
+	}
+	if !claimed {
+		return
+	}
+
+	total := int64(0)
+	now := time.Now().UTC()
+	for _, hourStart := range rollupHours(now, cfg.LookbackHours) {
+		enqueued, err := enqueueHour(ctx, pool, hourStart, hourStart.Add(time.Hour), true)
```
pavitrabhalla, line 125: `refreshCompleted = true` means every scheduler tick (every 5 min) resets recently-completed jobs back to `pending`. A job completed 30 seconds ago will be re-queued immediately. Consider a grace period condition in the upsert (e.g. `completed_at < now() - interval '1 hour'`) so truly-done recent hours aren't churned constantly.

internal/db/billing.sql.go
```diff
@@ -372,6 +680,92 @@ func (q *Queries) ListTeamBillingUsageHourly(ctx context.Context, arg ListTeamBi
...
+WHERE feature_enabled('billing_hourly_rollups', billing_teams.team_id)
+ORDER BY team_id
+`
+
+type ListTeamsForHourlyBillingRollupParams struct {
+	HourEnd   time.Time   `json:"hour_end"`
+	HourStart interface{} `json:"hour_start"`
+}
+
+// Teams with raw billing intervals overlapping an hour. The feature flag keeps
```
pavitrabhalla: `HourStart interface{}` loses type safety — sqlc couldn't infer the type because the arg appears in `$2 < LEAST(now(), $1)` rather than against a typed column. `ListTeamsForHourlyBillingRollup` is also unused in production (the enqueue path embeds the CTE directly). Either fix the sqlc hint or remove the generated query.

supabase/migrations/20260617000003_billing_pricing_credits.sql
```diff
@@ -0,0 +1,137 @@
...
+    CONSTRAINT pricing_rate_period_valid
+        CHECK (effective_to IS NULL OR effective_to > effective_from),
+    CONSTRAINT pricing_rate_unique_version
+        UNIQUE (plan_key, resource, unit, effective_from)
+);
+
+CREATE INDEX idx_pricing_rate_active
+    ON pricing_rate(plan_key, resource, unit, effective_from DESC)
+    WHERE effective_to IS NULL;
+
```
pavitrabhalla, line 42: `WHERE effective_to IS NULL` means this index only covers open-ended rates. `ListActivePricingRates` also matches rates where `effective_to > now()`, so those rows fall back to a seq scan. Drop the partial predicate or widen it to `WHERE effective_to IS NULL OR effective_to > now()`.

internal/integration/integration_test.go
```diff
@@ -713,6 +714,111 @@ func TestIntegration_BillingMetricsWriteFeatureFlag(t *testing.T) {
...
+	if len(rates) != 3 {
+		t.Fatalf("rates length = %d, want 3", len(rates))
+	}
+	for _, rate := range rates {
+		if rate.Resource == "vcpu" {
+			if got := numericFloat64(t, rate.PriceUsd); got != 0.000022 {
+				t.Fatalf("vcpu price = %v, want newest effective price", got)
+			}
+			return
+		}
```
pavitrabhalla, line 761: `return` here exits the test after finding vcpu, so `memory_gib` and `storage_gib` rates are never verified. Use `t.Errorf` (not `t.Fatalf`) and let the loop finish.

internal/billing/rollup.go
awille-ai: Backfill cursor-advance failures now log a warning with the affected hour range and enqueue count.

internal/billing/rollup.go
awille-ai: Completed historical jobs are not requeued every scheduler tick; current-hour jobs can still refresh.

internal/db/billing.sql.go
awille-ai: Removes the unused ListTeamsForHourlyBillingRollup sqlc query that generated interface{} params.

supabase/migrations/20260617000003_billing_pricing_credits.sql
awille-ai: Widens the pricing-rate index by removing the effective_to IS NULL partial predicate.

internal/integration/integration_test.go
awille-ai: Updates the pricing tests so they verify all expected resources instead of returning after vcpu.

internal/billing/rollup.go
awille-ai: BILLING_HOURLY_ROLLUP_WORKERS=0 now falls back to the default worker count.

pavitrabhalla (approved):

pavitrabhalla:
LGTM

PR #119, opened 2026-06-18
Expose pricing UI api

awille-ai:
this exposes the pricing for the UI to be able to get cost/MB/CPU/etc.
Pricing is used by the marketing page (unauthenticated) as well as user specific pricing used in the usage calculations, so it's exposed both authenticated, and unauthenticated


awille-ai:
Checked the current PR head for the inactive-plan review thread. The underlying concern appears addressed in the current branch: `GetTeamActivePricingPlan` joins `pricing_plan` and filters `p.active`, and `ListActivePricingRates` ranks current rates with `row_number()` so only the newest active `(resource, unit)` row is returned. I also reviewed the pricing API for billing security/scale and recommended adding cache headers: public PAYG pricing can be short-lived public/CDN cached, while authenticated team pricing should be private and vary on `Authorization`.

awille-ai:
I checked the current branch after the rebase. The review comments appear addressed:

- `ListActivePricingRates` now ranks active rates by `(resource, unit)` and returns only `rate_rank = 1`, ordered by newest `effective_from` / `created_at` / `id`, so stale duplicate active rates should not be returned.
- `GetTeamActivePricingPlan` still joins `pricing_plan` and filters `p.active`, so inactive team pricing assignments fall back to `payg` instead of selecting an inactive plan.
- The generated sqlc code matches those query definitions.

One follow-up I’d still consider before merge is adding/restoring an integration test for the inactive-team-pricing-plan fallback, since the branch currently has coverage for newest effective rates but I did not see that fallback test in the current diff.

internal/api/handlers_billing.go
```diff
@@ -0,0 +1,202 @@
...
+	teamID, err := teamIDFromContext(c)
+	if err != nil {
+		respondError(c, ErrUnauthorized)
+		return
+	}
+
+	rates, err := h.DB.ListActivePricingRatesForTeam(c.Request.Context(), db.ListActivePricingRatesForTeamParams{
+		TeamID:      teamID,
+		EffectiveAt: time.Now().UTC(),
+	})
```
awille-ai: This fixes the earlier two-read consistency issue, but it now makes effective pricing depend on the API server clock rather than the database clock used elsewhere in the billing SQL. Around `effective_from`/`effective_to` boundaries, even small app/DB clock skew could select a plan or rate that the billing calculations would not yet consider active. Could we have the SQL choose a single DB-side timestamp instead, e.g. a `selected_time AS (SELECT now() AS effective_at)` CTE used for both the plan and rate predicates? If tests need deterministic timestamps, a separate `...ForTeamAt` query would keep production on DB time while preserving testability.

awille-ai (commented):
I re-reviewed the current head. The earlier cache-header and two-read race issues look addressed, and CI is green. I left one edge-case comment around using the API server clock for effective pricing selection.

meAmitPatil (approved):
LGTM !!

PR #121, opened 2026-06-18
Runtime secret binding: attach/detach secrets on an existing sandbox

meAmitPatil:
Attach or detach a secret on a sandbox that already exists, with revocation so a running sandbox loses a removed secret.


internal/api/handlers_sandbox_secret.go
```diff
@@ -0,0 +1,285 @@
...
+	case err != nil:
+		log.Error().Err(err).Str("sandbox_id", sandboxID.String()).Msg("insert sandbox secret binding")
+		respondError(c, ErrInternal)
+		return
+	}
+
+	// A running sandbox is updated now; a paused one picks it up on resume. Fail
+	// closed: roll back the row if the proxy JWT can't be re-minted/injected.
+	if sandbox.Status == db.SandboxStatusActive {
+		meta, lerr := h.loadSecretBindingMeta(ctx, sandboxID)
```
awille-ai, line 192: if applySecretBindings fails after boxd accepted the /init update, the rollback deletes the DB row but does not insert the newly minted proxy token into revoked_proxy_token. A network timeout after apply is enough to produce this. Result: API returns 500, but the sandbox may still hold a JWT/env var that can swap the real credential until JWT expiry.

internal/api/handlers_secret.go
```diff
@@ -1081,6 +1095,65 @@ func (h *Handlers) resolveSecretBindingsForCreate(
 	return bindings, meta, nil
 }
 
+// loadSecretBindingMeta rebuilds a sandbox's binding metadata from the DB. The
+// stored proxy token is reused so re-minting a JWT doesn't rotate stand-ins; a
+// binding stored before tokens were persisted gets one minted on the fly.
+func (h *Handlers) loadSecretBindingMeta(ctx context.Context, sandboxID uuid.UUID) ([]SecretBindingMeta, error) {
```
awille-ai, line 1101: P1: Legacy bindings with NULL proxy_token cannot be safely detached
The migration leaves sandbox_secret.proxy_token nullable, and loadSecretBindingMeta mints a token on the fly without persisting it. Later, DetachSandboxSecret skips revocation when ProxyToken == nil. Result: a legacy binding that was reminted can continue using the detached credential.

internal/api/handlers_secret.go
```diff
@@ -1081,6 +1095,65 @@ func (h *Handlers) resolveSecretBindingsForCreate(
...
+			ProxyToken:       token,
+		})
+	}
+	return meta, nil
+}
+
+// applySecretBindings mints the secrets JWT for meta and injects it, with the secret
+// env vars, into a live sandbox. Empty meta injects no JWT. InjectSandboxEnv merges,
+// so it can add or update an env var but not remove one.
+func (h *Handlers) applySecretBindings(ctx context.Context, sandbox db.Sandbox, meta []SecretBindingMeta) error {
```
awille-ai, line 1147: P2: Detach does not actually remove env vars from live sandbox state
applySecretBindings calls InjectSandboxEnv, but boxd /init is additive and only merges env vars in cmd/boxd/main.go. After detach, future execs can still inherit the old stand-in env var and JWT. Revocation eventually blocks swaps, but the runtime state does not match the API contract.

internal/api/handlers_sandbox_secret.go
```diff
@@ -0,0 +1,285 @@
...
+)
+
+type attachSecretRequest struct {
+	EnvKey     string `json:"env_key"`
+	SecretName string `json:"secret_name"`
+}
+
+// AttachSandboxSecret binds a stored secret to an existing sandbox under an env var.
+// Takes effect for processes started after the call. POST /sandboxes/{id}/secrets.
+func (h *Handlers) AttachSandboxSecret(c *gin.Context) {
```
awille-ai, line 48: P2: Multi-API ordering is still not fully serialized
The attach path uses a DB advisory lock for the count-check/insert, but releases it before loadSecretBindingMeta and applySecretBindings. In multi-instance deployments, two attaches or attach-vs-resume can still apply stale JWT/env state out of order. The local lockSandboxSecrets does not protect across API instances. This affects AttachSandboxSecret and resume around internal/api/handlers.go.

cmd/secretsproxy/main.go
```diff
@@ -73,12 +73,13 @@ func run() error {
 	cachedRules := secretsproxy.NewCachedRules(httpRules, secretsproxy.RulesCacheOptions{})
 
 	revoker := secretsproxy.NewRevoker()
-	revoked, err := secretsproxy.FetchRevokedSandboxes(rootCtx, cfg.ControlPlaneURL, cfg.DaemonAuthToken)
+	revoked, revokedTokens, err := secretsproxy.FetchRevokedSandboxes(rootCtx, cfg.ControlPlaneURL, cfg.DaemonAuthToken)
```
awille-ai, line 76: P2: Revocation is polling-based, not pushed
Detached proxy tokens only reach the in-host proxy through bootstrap/reconcile, currently every 60 seconds in cmd/secretsproxy/main.go. That may be acceptable if explicitly documented as eventual revocation, but it is a real security window for active sandboxes.

awille-ai (commented):
The PR correctly scopes attach by team secret ownership, adds a DB-backed binding cap guard, persists proxy tokens for new bindings, and adds cleanup for expired revoked tokens. Tests cover happy paths and some fail-closed behavior.

PR #122, opened 2026-06-18
Race condition fix

awille-ai:
Fixes hourly billing rollup backfill to use tenant-scoped cursors so disabled teams are not skipped and concurrent scheduler workers can safely make progress without cursor races.

### DIFF
--- a/internal/billing/rollup.go
+++ b/internal/billing/rollup.go
@@ -43,6 +43,13 @@ type HourlyRollupConfig struct {
 	Workers               int
 }
 
+type teamBackfillBatch struct {
+	TeamID     uuid.UUID
+	FromHour   time.Time
+	ToHour     time.Time
+	HourStarts []time.Time
+}
+
 type rollupJob struct {
 	ID           uuid.UUID
 	TeamID       uuid.UUID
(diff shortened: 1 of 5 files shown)


internal/billing/rollup.go
```diff
@@ -302,35 +309,218 @@ func enqueueBackfill(ctx context.Context, pool *pgxpool.Pool, cfg HourlyRollupCo
...
+        FROM sandbox_storage_interval i
+        WHERE i.started_at < $2
+          AND $1 < LEAST(now(), $2)
+          AND COALESCE(i.ended_at, LEAST(now(), $2)) > $1
+    ) billing_teams
+    WHERE feature_enabled('billing_hourly_rollups', billing_teams.team_id)
+),
+upserted AS (
+    INSERT INTO billing_rollup_team_backfill_state (team_id, next_hour_start)
+    SELECT team_id, $1
```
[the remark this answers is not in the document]
awille-ai: existing enabled tenants inherit billing_rollup_backfill_state.next_hour_start, avoiding a 90-day replay after deploy. Teams that were disabled are not seeded, so when they are enabled later they can still catch up from the configured backfill window.

supabase/migrations/20260618000003_billing_rollup_team_backfill_state.sql
```diff
@@ -0,0 +1,43 @@
...
+CREATE TABLE billing_rollup_team_backfill_state (
+    team_id          uuid PRIMARY KEY REFERENCES team(id) ON DELETE CASCADE,
+    next_hour_start  timestamptz NOT NULL,
+    updated_at       timestamptz NOT NULL DEFAULT now()
+);
+
+CREATE INDEX idx_billing_rollup_team_backfill_next_hour
+    ON billing_rollup_team_backfill_state(next_hour_start, team_id);
+
+ALTER TABLE public.billing_rollup_team_backfill_state ENABLE ROW LEVEL SECURITY;
```
pavitrabhalla, line 20: RLS is enabled but no policies are defined. Is access to this table restricted entirely to the service role, or are policies planned separately?

internal/integration/integration_test.go
```diff
@@ -1513,6 +1506,36 @@ func TestIntegration_BillingRollupBackfillCursorAdvancesAfterEnqueue(t *testing.
...
+		WHERE team_id = $1
+	`, teamID).Scan(&cursor); err != nil {
+		t.Fatalf("read team backfill cursor: %v", err)
+	}
+	if !cursor.Equal(firstHourEnd) {
+		t.Fatalf("cursor = %s, want advanced to %s after successful enqueue", cursor, firstHourEnd)
+	}
+}
+
+func TestIntegration_BillingRollupBackfillCursorAdvanceToleratesConcurrentAdvance(t *testing.T) {
```
pavitrabhalla: This test exercises `advanceBackfillCursor` (global cursor). Is there a corresponding test for `advanceTeamBackfillCursor` concurrent advance?

pavitrabhalla (commented):
Are `nextBackfillHours` and `advanceBackfillCursor` still called in production? If `enqueueBackfill` now routes entirely through the per-team path, these are dead code and should be removed.

awille-ai:
Addressed the latest review feedback in the follow-up patch:

- Clarified the `billing_rollup_team_backfill_state` RLS posture in the migration: RLS is enabled with no policies intentionally, so anon/authenticated roles have no direct access and only the service-role control plane scheduler can read/advance cursor state.
- Added a dedicated concurrent-advance integration test for `advanceTeamBackfillCursor`.
- Removed the now-dead global backfill cursor helpers/exports/tests (`nextBackfillHours` / `advanceBackfillCursor`) since production backfill now routes through the per-team cursor path.

Validation to run after applying: `gofmt -w internal/billing/rollup.go internal/billing/rollup_integration_exports.go internal/integration/integration_test.go` and `DATABASE_URL="postgres://postgres:postgres@localhost:5432/sandbox_test?sslmode=disable" go test -tags integration ./internal/integration/ -v -count=1 -run "BillingRollupBackfill|BillingRollupTeamBackfillCursor" -timeout 10m`.

internal/billing/rollup.go
```diff
@@ -302,85 +309,216 @@ func enqueueBackfill(ctx context.Context, pool *pgxpool.Pool, cfg HourlyRollupCo
...
-func nextBackfillHours(ctx context.Context, pool *pgxpool.Pool, startHour, endHour time.Time, batchHours int) ([]time.Time, error) {
+func enqueueTeamHour(ctx context.Context, pool *pgxpool.Pool, teamID uuid.UUID, hourStart, hourEnd time.Time, refreshCompleted bool) (int64, error) {
 	const query = `
-INSERT INTO billing_rollup_backfill_state (name, next_hour_start)
-VALUES ('hourly', $1)
-ON CONFLICT (name) DO UPDATE
-SET next_hour_start = CASE
-        WHEN billing_rollup_backfill_state.next_hour_start < $1 THEN $1
-        ELSE billing_rollup_backfill_state.next_hour_start
+INSERT INTO billing_rollup_job (team_id, hour_start, hour_end, status, next_attempt_at)
```
pavitrabhalla, line 361: All the other billing interval queries go through db.Queries — is there a reason the ones in this file bypass sqlc rather than living in the query files?

internal/billing/rollup.go
```diff
@@ -302,35 +309,218 @@ func enqueueBackfill(ctx context.Context, pool *pgxpool.Pool, cfg HourlyRollupCo
...
+           billing_rollup_job.hour_end > now()
+           OR billing_rollup_job.completed_at < now() - interval '1 hour'
+       )
+   )`
+
+	tag, err := pool.Exec(ctx, query, teamID, hourStart, hourEnd, refreshCompleted)
+	return tag.RowsAffected(), err
+}
+
+func nextTeamBackfillBatches(ctx context.Context, pool *pgxpool.Pool, startHour, endHour time.Time, batchHours int, teamLimit int) ([]teamBackfillBatch, error) {
```
pavitrabhalla, line 418: Same as `enqueueTeamHour` — this query is inline rather than in the sqlc query files. Could move it there for consistency with the rest of the DB access layer.

internal/billing/rollup.go
```diff
@@ -302,35 +309,218 @@ func enqueueBackfill(ctx context.Context, pool *pgxpool.Pool, cfg HourlyRollupCo
...
+		}
+		for hour := batch.FromHour; hour.Before(batch.ToHour); hour = hour.Add(time.Hour) {
+			batch.HourStarts = append(batch.HourStarts, hour)
+		}
+		batches = append(batches, batch)
+	}
+	return batches, rows.Err()
+}
+
+func advanceTeamBackfillCursor(ctx context.Context, pool *pgxpool.Pool, teamID uuid.UUID, fromHour, toHour time.Time) error {
```
pavitrabhalla, line 501: Same — inline SQL, could be a sqlc query.

internal/billing/rollup.go
```diff
@@ -302,35 +309,218 @@ func enqueueBackfill(ctx context.Context, pool *pgxpool.Pool, cfg HourlyRollupCo
...
+`, teamID).Scan(&current); err != nil {
+		return err
+	}
+	if !current.Before(toHour) {
+		return nil
+	}
+	return fmt.Errorf("billing rollup team backfill cursor moved backward before advance for team %s from %s to %s; current cursor is %s", teamID, fromHour, toHour, current)
+}
+
+func teamFeatureEnabled(ctx context.Context, pool *pgxpool.Pool, teamID uuid.UUID, key string) (bool, error) {
```
pavitrabhalla, line 531: Same — inline SQL, could be a sqlc query.

awille-ai:
Follow-up on the latest comments:

- Addressed the scale issue in `nextTeamBackfillBatches`: the query now orders/limits eligible team cursors before `INSERT ... ON CONFLICT`, so each scheduler tick only writes up to `BatchSize` cursor rows instead of touching every enabled tenant with historical intervals.
- On the inline SQL comments: I agree these could eventually move into sqlc, but I’d keep the scheduler queue/claim SQL inline in this PR. These are operational worker-queue statements with `SKIP LOCKED`, lock refresh semantics, and data-modifying CTEs that are tightly coupled to the rollup worker. Moving them to sqlc would require a wider generated-code change and doesn’t change the race/scaling fix here. The customer-facing billing reads/writes remain in `db.Queries`.

Validation to run after applying: `gofmt -w internal/billing/rollup.go` and `DATABASE_URL="postgres://postgres:postgres@localhost:5432/sandbox_test?sslmode=disable" go test -tags integration ./internal/integration/ -v -count=1 -run "BillingRollupBackfill|BillingRollupTeamBackfillCursor" -timeout 10m`.

supabase/migrations/20260618000003_billing_rollup_team_backfill_state.sql
awille-ai: Yes, this table is intentionally restricted to service-role/control-plane access only. I added an explicit migration comment above the RLS enablement explaining that no policies are created on purpose: anon/authenticated roles get no direct access, while the scheduler advances cursors using the service role.

internal/integration/integration_test.go
awille-ai: Added a dedicated test for the tenant cursor path: `TestIntegration_BillingRollupTeamBackfillCursorAdvanceToleratesConcurrentAdvance`. It seeds `billing_rollup_team_backfill_state`, advances from an earlier `fromHour`, verifies the cursor reaches `toHour`, and checks the advance is idempotent once already advanced.

internal/billing/rollup.go
```diff
@@ -302,85 +309,216 @@ func enqueueBackfill(ctx context.Context, pool *pgxpool.Pool, cfg HourlyRollupCo
...
+          AND $1 < LEAST(now(), $2)
+          AND COALESCE(i.ended_at, LEAST(now(), $2)) > $1
+    ) billing_teams
+    WHERE feature_enabled('billing_hourly_rollups', billing_teams.team_id)
+),
+upserted AS (
+    INSERT INTO billing_rollup_team_backfill_state (team_id, next_hour_start)
+    SELECT team_id, $1
+    FROM candidate_teams
+    ON CONFLICT (team_id) DO UPDATE
```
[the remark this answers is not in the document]
awille-ai: Addressed. `nextTeamBackfillBatches` now builds `eligible_team_cursors`, orders/limits into `candidate_teams`, and only then performs `INSERT ... ON CONFLICT`, so each scheduler tick writes at most `BatchSize` cursor rows instead of touching all enabled tenants with historical intervals.

internal/billing/rollup.go
awille-ai: Good question. I kept these rollup scheduler/queue statements inline intentionally for this PR because they are operational worker-queue SQL rather than customer-facing billing reads/writes: they use `SKIP LOCKED`, lock refresh semantics, scheduler leases, and data-modifying CTEs tightly coupled to the rollup worker. Moving them into sqlc would require regenerating and widening the patch, but it would not materially change the race-condition fix. I’d be comfortable doing a follow-up refactor to move the queue helpers into sqlc once this correctness fix lands.

internal/billing/rollup.go
awille-ai: Same answer for this helper: it is internal scheduler cursor state, and keeping it inline keeps this PR focused on the per-team cursor correctness/scaling change. A follow-up sqlc cleanup would be reasonable, but I don’t think this PR should grow into a generated-code refactor.

internal/billing/rollup.go
awille-ai: Same rationale here: this is worker-queue/scheduler SQL, not a tenant-facing billing query. It is tightly coupled to enqueue semantics and the rollup worker. I’d keep it inline in this race-condition PR to avoid a broader sqlc regeneration/refactor, and move the queue helpers to sqlc in a separate cleanup if we want that consistency across the worker internals.

internal/billing/rollup.go
awille-ai: Same here. This is an internal feature-flag check for the worker’s cursor advancement path. It can be moved into sqlc later, but I’d avoid expanding this race-condition fix into generated query plumbing unless we want to do a dedicated DB-access consistency pass.

supabase/migrations/20260618000003_billing_rollup_team_backfill_state.sql
```diff
@@ -0,0 +1,46 @@
...
+    WHERE name = 'hourly'
+),
+candidate_teams AS (
+    SELECT DISTINCT team_id
+    FROM (
+        SELECT team_id FROM sandbox_compute_billing_interval
+        UNION
+        SELECT team_id FROM sandbox_storage_interval
+    ) billing_teams
+    WHERE feature_enabled('billing_hourly_rollups', billing_teams.team_id)
```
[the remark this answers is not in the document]
awille-ai: Addressed. The migration bootstrap now seeds only tenants that are enabled without an explicit team-level `billing_hourly_rollups` override. Teams with a `team_feature_flag` row are left unseeded because current state alone cannot prove they were processed under the old global cursor; when enabled, the scheduler will initialize them from the configured backfill start and catch up skipped hours.

pavitrabhalla (approved):

PR #123, opened 2026-06-19
reaper: anchor sandbox timeout window on current active session so resume re-arms it

meAmitPatil:
## Problem

`timeout_seconds` (the opt-in auto-pause cap) is measured from `created_at`. Once a sandbox lives past its timeout, `created_at + timeout_seconds` is permanently in the past — so **every resume is immediately re-eligible for the reaper**, which re-pauses it within one poll cycle (≤30s). The result: a sandbox that has outlived its timeout can never stay running — resuming it (e.g. opening a terminal) drops the connection ~20s later, repeatedly.

This hits interactive/terminal users whose sandboxes are resumed after the timeout has elapsed.

## Fix

Anchor the timeout window on the **start of the current active session** instead of `created_at`:

```sql
-- before:  created_at + timeout_seconds < now()
-- after:   max(open sandbox_active_interval.started_at) + timeout_seconds < now()
```

The resume path (`ActivateSandbox`) already opens a fresh `sandbox_active_interval` row with `started_at = now()` on every resume, so each resume now re-arms the full `timeout_seconds` window. The 60s grace floor is moved to the same anchor. No migration — reuses existing data.

## Why it's safe

Both pause paths atomically close the open interval (`BeginPause`, `ClaimExpiredSandboxes`), and resume opens a fresh one — so `max(open started_at)` is the last activation time:

- Pause → interval closed (`ended_at` set).
- Resume → fresh interval (`started_at = now()`); `ActivateSandbox`'s `ON CONFLICT (sandbox_id) WHERE ended_at IS NULL DO NOTHING` only matters if a stale open interval exists, which the atomic closes prevent.

### DIFF
--- a/db/queries/sandboxes.sql
+++ b/db/queries/sandboxes.sql
@@ -309,23 +309,35 @@ JOIN team t ON s.team_id = t.id
 WHERE s.id = $1 AND s.destroyed_at IS NULL;
 
 -- name: ClaimExpiredSandboxes :many
--- Atomically claims active sandboxes whose hard timeout has elapsed and marks
+-- Atomically claims active sandboxes whose timeout window has elapsed and marks
 -- them as 'pausing'. FOR UPDATE SKIP LOCKED lets concurrent reaper replicas
 -- skip rows already being processed, so multi-replica Cloud Run deployments
 -- do not double-process the same sandbox.
 --
+-- The timeout window is measured from the start of the current active session
+-- (the open sandbox_active_interval row, which is re-opened on every resume),
+-- not from created_at. So each resume re-arms the full timeout_seconds window:
+-- a sandbox resumed after pausing gets a fresh window instead of being claimed
+-- again immediately on a deadline that already elapsed.
+--
 -- Only 'active' sandboxes are claimed — paused sandboxes are already stopped,
 -- and transient states (starting, pausing) are skipped to avoid racing with
 -- in-progress operations. The 60-second grace window prevents reaping a sandbox
--- that was just created with a very short timeout before it finishes starting up.
+-- that was just resumed (or created) with a very short timeout before it settles.
 WITH expired AS (
   SELECT id, team_id, name, snapshot_id, host_id
   FROM sandbox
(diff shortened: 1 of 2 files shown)


db/queries/sandboxes.sql
```diff
@@ -309,23 +309,35 @@ JOIN team t ON s.team_id = t.id
...
   WHERE destroyed_at IS NULL
     AND timeout_seconds IS NOT NULL
     AND status = 'active'
-    AND created_at + (timeout_seconds || ' seconds')::interval < now()
-    AND created_at < now() - interval '60 seconds'
+    AND (
+      SELECT max(sai.started_at) FROM sandbox_active_interval sai
+      WHERE sai.sandbox_id = sandbox.id AND sai.ended_at IS NULL
+    ) + (timeout_seconds || ' seconds')::interval < now()
+    AND (
```
pavitrabhalla: Does sandbox creation open a `sandbox_active_interval` row? If not, this returns NULL for a sandbox that's never been paused, and the reaper silently skips it — the old `created_at` path handled that case.

db/queries/sandboxes.sql
```diff
@@ -309,23 +309,35 @@ JOIN team t ON s.team_id = t.id
...
+      SELECT max(sai.started_at) FROM sandbox_active_interval sai
+      WHERE sai.sandbox_id = sandbox.id AND sai.ended_at IS NULL
+    ) < now() - interval '60 seconds'
   ORDER BY created_at ASC
```
pavitrabhalla: Subquery runs twice. A lateral join evaluates it once — does that matter at your scale, or is PG de-duplicating it already?

pavitrabhalla (commented):
No test for the repro case — is there one planned?

pavitrabhalla (approved):
One thing worth noting: the open_sessions CTE scans all open intervals across all sandboxes (not just active ones), whereas the old correlated subquery only touched intervals for rows that passed the status = 'active' filter. At large scale this could be heavier, but it avoids N correlated lookups so it’s almost certainly a net win.

PR #127, opened 2026-06-20
sandbox: reclaim snapshot dir and failed-sandbox rundir on delete

meAmitPatil:
Fixes the two in-band cleanup leaks that #124 (rundir) didn't cover.

**Bug 5 — failed sandboxes leaked their rundir.** Deleting a `failed` sandbox skipped VM teardown entirely. It now attempts `DestroyInstance` (idempotent; cleans the rundir path-based) — best-effort, so a failed sandbox stays deletable even if its host is unreachable.

**Bug 2 — snapshot dir leaked.** `cleanupSandboxSnapshots` deleted snapshot files only via DB rows + a reachable vmd, so a missing snapshot row (pause wrote files but the row insert didn't land) or an unreachable vmd left `<SnapshotDir>/<id>/` behind forever. New `DeleteSandboxSnapshots` vmd RPC removes the whole per-sandbox snapshot dir path-based and idempotently, independent of DB rows. Guarded (leaf + non-reserved vm_id) so it can't escape the snapshot root or remove the shared template tree. Only invoked on delete — never on pause/resume/revert.

Both are fast-path fixes that lower the leak *rate*; the disk reconciler (#125 → reclamation) remains the eventual backstop. Tests: vmd-side removal + reserved-name rejection; handler-side failed-sandbox teardown + best-effort-on-error + path-based snapshot sweep.


meAmitPatil:
@claude review

PR #129, opened 2026-06-20
vmd: enable disk reclamation (flag-gated, quarantine-move + sweep)

meAmitPatil:
PR 3 of the disk-GC plan, stacked on #125. Turns Drift 6 from detect-only into reclamation — **behind a flag that defaults off**, so merging is a behavioral no-op until the detect-only numbers are validated on prod and `VMD_DISK_RECLAIM=true` is set.

When enabled, each pass quarantine-moves orphan dirs to `<root>/.trash/<date>/<uuid>` (reversible `rename`, same filesystem) up to `DiskDeleteBudget` (50), and a sweep permanently removes quarantine older than `DiskTrashRetention` (72h) — the soak window to spot and reverse a bad reclaim.

Safety stack (on top of #125's keep-set + grace):
- flag default off (`VMD_DISK_RECLAIM`); fail-closed (no DB keep-set → no-op)
- **empty-keep-set abort on the reclaim path** (the guard #125 deliberately moved off detect onto here)
- per-pass delete budget; UUID-leaf path containment in `quarantineDir`; `.trash` excluded from the scan by construction
- quarantine-move (reversible) → 72h soak → sweep `rm` (the only irreversible step, and only on aged-out trash)

Tests: quarantine moves a uuid dir / rejects non-uuid; sweep removes expired & keeps recent; reclaim respects the budget.

### DIFF
--- a/internal/vm/reconciler.go
+++ b/internal/vm/reconciler.go
@@ -824,6 +824,13 @@ func (r *Reconciler) consumeAutoFailBudget(vmID string) bool {
 // in-memory map. The VM is already gone in reality; this just cleans up
 // VMD's cache.
 func (r *Reconciler) markStale(vmID string) {
+	// Capture the namespace before deleting the record: a VM whose teardown
+	// didn't run (e.g. a vmd timeout mid-DELETE) would otherwise leak its slot.
+	var namespace string
+	if rec, err := r.mgr.state.Get(vmID); err == nil && rec != nil {
+		namespace = rec.Namespace
+	}
+
 	// Delete from BoltDB first. If this fails, keep the in-memory entry
 	// so the state stays consistent — the reconciler will retry on the
 	// next run. Deleting from the map before BoltDB would cause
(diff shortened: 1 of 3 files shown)


internal/vm/reconciler.go
```diff
@@ -471,22 +504,109 @@ func (r *Reconciler) detectDiskOrphans(ctx context.Context, snapshotTime time.Ti
 		Int("orphan_dirs", dirCount).
 		Int64("orphan_bytes", bytes).
 		Strs("sample", sample).
-		Msg("disk scan (detect-only): orphan per-sandbox dirs would be reclaimed")
+		Bool("reclaim", r.cfg.DiskReclaimEnabled).
+		Msg("disk scan: orphan per-sandbox dirs detected")
+
+	if r.cfg.DiskReclaimEnabled {
+		r.reclaimDiskOrphans(ctx, snapshotTime, orphans, onDisk)
```
awille-ai: **P2: Bound pre-reclaim sizing to the same budget/cancellation rules**

Before this move phase, the code walks every orphan directory to compute `orphan_bytes`, even though `DiskDeleteBudget` may only allow a small prefix to be quarantined. With a large backlog, each reconcile pass can spend most or all of `runTimeout` repeatedly sizing directories that will not be moved this pass; if the context is cancelled during that work, this still proceeds into `quarantineDir`, which does not check `ctx`. Consider sizing only the budgeted sample when reclaim is enabled, or checking `ctx.Err()` before starting destructive moves. A test with `DiskDeleteBudget < len(orphans)` plus a cancelled context would cover this edge case.

awille-ai (commented):
I reviewed the latest head (`ecfdf7e2`). The earlier high-risk reclaim-safety issues appear addressed in this revision, and I did not find a new high/critical security issue. I left one non-blocking scalability/cancellation edge-case comment around the reclaim pass sizing work.

awille-ai (approved):
Approved with one non-blocking P2 scalability/cancellation note. I did not find any high or critical security/correctness issues in the latest revision.

PR #140, opened 2026-06-22
phase 1 of user management

awille-ai:
Add RBAC tables.
Add audit log table.
Add seed data for roles and permissions.
Backfill existing users into their current teams.
Allow users to belong to multiple active teams while preventing duplicate active memberships within the same team.

### DIFF
--- a/supabase/migrations/20260622000001_rbac_phase1.sql
+++ b/supabase/migrations/20260622000001_rbac_phase1.sql
@@ -1,0 +1,447 @@
+-- RBAC Phase 1: schema foundation, seed data, and legacy backfill.
+--
+-- This migration adds scoped roles/permissions, membership tracking, and an
+-- append-only audit log table. It does not add application-level enforcement.
+
+CREATE TABLE IF NOT EXISTS team_memberships (
+    id         uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+    team_id    uuid NOT NULL REFERENCES team(id),
+    user_id    uuid NOT NULL REFERENCES profile(id),
+    status     text NOT NULL DEFAULT 'active',
+    created_at timestamptz NOT NULL DEFAULT now(),
+    updated_at timestamptz NOT NULL DEFAULT now(),
+
+    CONSTRAINT team_memberships_status_check
+        CHECK (status IN ('active', 'inactive', 'invited'))
+);
+
+CREATE INDEX IF NOT EXISTS idx_team_memberships_team_user
+    ON team_memberships(team_id, user_id);
+CREATE INDEX IF NOT EXISTS idx_team_memberships_user
+    ON team_memberships(user_id);
+CREATE UNIQUE INDEX IF NOT EXISTS idx_team_memberships_active_team_user_unique
+    ON team_memberships(team_id, user_id)
+    WHERE status = 'active';
+CREATE UNIQUE INDEX IF NOT EXISTS idx_team_memberships_active_user_unique
+    ON team_memberships(user_id)
+    WHERE status = 'active';
+
+CREATE TABLE IF NOT EXISTS roles (
+    id          uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+    name        text NOT NULL UNIQUE,
+    scope_type  text NOT NULL,
+    description text,
+    created_at  timestamptz NOT NULL DEFAULT now(),
+    updated_at  timestamptz NOT NULL DEFAULT now(),
+
+    CONSTRAINT roles_name_nonempty CHECK (name <> ''),
+    CONSTRAINT roles_scope_type_check CHECK (scope_type IN ('platform', 'team'))
+);
+
+ALTER TABLE roles
+    ADD CONSTRAINT roles_id_scope_unique UNIQUE (id, scope_type);
+
+CREATE TABLE IF NOT EXISTS permissions (
+    id          uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+    name        text NOT NULL UNIQUE,
+    description text,
+    created_at  timestamptz NOT NULL DEFAULT now(),
+    updated_at  timestamptz NOT NULL DEFAULT now(),
+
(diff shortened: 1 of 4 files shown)


supabase/migrations/20260622000001_rbac_phase1.sql
```diff
@@ -0,0 +1,316 @@
...
+    permission_id  uuid NOT NULL REFERENCES permissions(id) ON DELETE CASCADE,
+    created_at     timestamptz NOT NULL DEFAULT now(),
+
+    PRIMARY KEY (role_id, permission_id)
+);
+
+CREATE TABLE IF NOT EXISTS user_role_assignments (
+    id          uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+    user_id     uuid NOT NULL REFERENCES profile(id),
+    role_id     uuid NOT NULL REFERENCES roles(id) ON DELETE CASCADE,
```
pavitrabhalla: `role_id ON DELETE CASCADE` means deleting a role wipes all of its assignment history. Is RESTRICT safer here — preserve the audit trail even if the role itself goes away?

supabase/migrations/20260622000001_rbac_phase1.sql
```diff
@@ -0,0 +1,316 @@
...
+      AND a.revoked_at IS NULL
+);
+
+DO $$
+DECLARE
+    orphan_count integer;
+BEGIN
+    SELECT COUNT(*)
+    INTO orphan_count
+    FROM profile p
```
pavitrabhalla, line 427: This raises if any `profile` row has no active team membership. Does that include service accounts, blocked users, or profiles that were invited but never joined? A `WHERE p.blocked_at IS NULL` (or equivalent) would scope it to active users only and avoid a surprise migration failure.

supabase/migrations/20260622000001_rbac_phase1.sql
```diff
@@ -0,0 +1,316 @@
...
+CREATE INDEX IF NOT EXISTS idx_audit_logs_team_created_at
+    ON audit_logs(team_id, created_at DESC);
+CREATE INDEX IF NOT EXISTS idx_audit_logs_actor_created_at
+    ON audit_logs(actor_user_id, created_at DESC);
+CREATE INDEX IF NOT EXISTS idx_audit_logs_target_created_at
+    ON audit_logs(target_user_id, created_at DESC);
+CREATE INDEX IF NOT EXISTS idx_audit_logs_event_created_at
+    ON audit_logs(event_type, created_at DESC);
+
+ALTER TABLE public.team_memberships ENABLE ROW LEVEL SECURITY;
```
pavitrabhalla, line 261: RLS is enabled on all six tables but no policies are added. With no policies, all non-service-role access is blocked by default — is that the intended posture until Phase 2? Worth a comment so this doesn't read as an oversight.

supabase/migrations/20260622000001_rbac_phase1.sql
```diff
@@ -0,0 +1,447 @@
...
+);
+
+CREATE INDEX IF NOT EXISTS idx_team_memberships_team_user
+    ON team_memberships(team_id, user_id);
+CREATE INDEX IF NOT EXISTS idx_team_memberships_user
+    ON team_memberships(user_id);
+CREATE UNIQUE INDEX IF NOT EXISTS idx_team_memberships_active_team_user_unique
+    ON team_memberships(team_id, user_id)
+    WHERE status = 'active';
+CREATE UNIQUE INDEX IF NOT EXISTS idx_team_memberships_active_user_unique
```
pavitrabhalla: This enforces at most one active team per user across all teams (not just per team/user pair). The preflight DO block also refuses the migration if any legacy user belongs to multiple teams. Is this constraint permanent or a temporary measure for the current product shape? Removing it later requires a migration.

meAmitPatil:
PR #140 — RBAC Phase 1 review

Worth fixing

1. Multi-member teams end up with no owner after backfill — the practical one to worry about.
Legacy team_member.role defaults to 'member' and nothing in this repo writes any other value. The backfill heuristic (lines 392-397):
member_count = 1            -> team_owner
'owner'/'team_owner'        -> team_owner
'admin'/'team_admin'        -> team_admin
ELSE                        -> viewer
So any team with ≥2 members where everyone is 'member' (likely every multi-member team in prod) backfills to all viewer, no team_owner — nobody can administer the team. The owner/admin branches are effectively dead unless the TS SDK actually writes those strings. Action: confirm the real role values in staging/prod before shipping; if it's all 'member', the heuristic needs a real ownership signal (e.g. team.created_by / earliest joined_at), not the legacy role column.

2. Backfill role-assignment isn't safely idempotent. team_sizes and legacy_memberships read from the new team_memberships table (lines 354-373), not legacy team_member. Fine on first run (memberships were just copied 1:1), but a re-run after Phase-2 app activity would inflate member_count and auto-assign viewer to app-created memberships that deliberately have no assignment. Source both CTEs from team_member directly. Low real-world odds (migrations run once) but the rest of the file is written to be idempotent, so this is a gap in that contract.

3. Non-idempotent DDL contradicts the file's own convention. ALTER TABLE ... ADD CONSTRAINT (lines 39, 81) and all five CREATE TRIGGER statements have no guard, while everything else uses IF NOT EXISTS / CREATE OR REPLACE. Sibling migration sandbox_secret_team_guard uses DROP TRIGGER IF EXISTS + CREATE. A partial-run replay dies with "already exists." Add DROP TRIGGER IF EXISTS before each trigger; wrap the constraints in a DO/EXCEPTION WHEN duplicate_object block or inline them in CREATE TABLE.

4. active -> invited silently revokes all team roles. The revocation trigger fires on OLD.status='active' AND NEW.status <> 'active' (line 158), which catches invited, not just inactive. Status enum is ('active','inactive','invited'). If re-inviting an existing member is ever a real transition, it strips their roles. Either restrict to NEW.status = 'inactive' or confirm active -> invited can't happen.


@awille-ai Feel free to pick and fix the ones which seem right.

meAmitPatil (approved):

PR #143, opened 2026-06-23
Phase 2 of the user management

awille-ai:
Phase 2: Authorization Library
Implement permission lookup.
Implement platform admin validation.
Implement team-scoped checks.
Add unit tests for authorization logic.
This depends on 
https://github.com/superserve-ai/sandbox/pull/140, which creates all the tables.

### DIFF
--- a/internal/api/handlers_billing.go
+++ b/internal/api/handlers_billing.go
@@ -49,6 +49,27 @@ func (h *Handlers) GetBillingPricing(c *gin.Context) {
 		respondError(c, ErrUnauthorized)
 		return
 	}
+	actorID := actorIDFromContext(c)
+	if actorID == nil {
+		respondError(c, ErrUnauthorized)
+		return
+	}
+
+	authzSvc := h.authzService()
+	if authzSvc == nil {
+		respondError(c, ErrInternal)
+		return
+	}
+	allowed, err := authzSvc.CanTeam(c.Request.Context(), *actorID, teamID, "billing:read")
+	if err != nil {
+		log.Error().Err(err).Str("team_id", teamID.String()).Str("user_id", actorID.String()).Msg("RBAC billing read check failed")
+		respondError(c, ErrInternal)
+		return
+	}
+	if !allowed {
+		respondError(c, ErrForbidden)
+		return
+	}
 
 	rates, err := h.DB.ListActivePricingRatesForTeamCurrent(c.Request.Context(), teamID)
 	if err != nil {
--- a/internal/api/rbac_phase2b.go
+++ b/internal/api/rbac_phase2b.go
@@ -1082,6 +1082,7 @@ func (h *Handlers) assignRole(c *gin.Context, platform bool) {
 
 	ctx := c.Request.Context()
 	var assignment roleAssignmentRecord
+	var assignmentEmail string
 	err = h.withRBACMutation(ctx, func(q db.DBTX, svc *authz.Service) error {
 		if err := lockTeamMutation(ctx, q, teamID); err != nil {
 			return err
@@ -1124,11 +1125,13 @@ func (h *Handlers) assignRole(c *gin.Context, platform bool) {
 			return err
 		}
 		if err := q.QueryRow(ctx, `
-			SELECT ura.id, ura.user_id, ura.role_id, r.name, ura.team_id, ura.granted_by, ura.granted_at,
+			SELECT ura.id, ura.user_id, p.email, ura.role_id, r.name, ura.team_id, ura.granted_by, ura.granted_at,
 			       ura.revoked_at, ura.created_at, ura.updated_at
 			FROM user_role_assignments ura
 			JOIN roles r
 			  ON r.id = ura.role_id
+			JOIN profile p
+			  ON p.id = ura.user_id
 			WHERE ura.user_id = $1
 			  AND ura.role_id = $2
 			  AND ura.team_id = $3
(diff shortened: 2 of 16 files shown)


internal/integration/rbac_phase1_test.go
```diff
@@ -0,0 +1,601 @@
...
+			NULL,
+			lm.created_at,
+			NULL,
+			lm.created_at,
+			lm.created_at
+		FROM legacy_memberships lm
+		JOIN roles r
+		  ON r.scope_type = 'team'
+		 AND r.name = CASE
+			WHEN lm.legacy_role IN ('owner', 'team_owner') THEN 'team_owner'
```
pavitrabhalla, line 714: The CASE priority here differs from the migration: the migration checks `member_count = 1` first (single-member team always becomes `team_owner` regardless of legacy role), but here legacy role wins. The `soleAdminTeam` test case from Phase 1 was also removed — did the ordering change intentionally, or did removing that test case mask the divergence?

internal/api/rbac_phase2b.go
```diff
@@ -0,0 +1,1247 @@
...
+	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			respondErrorMsg(c, "not_found", "team or user not found", http.StatusNotFound)
+			return
+		}
+		if errors.Is(err, ErrConflict) {
+			respondError(c, ErrConflict)
+			return
+		}
+		if strings.Contains(err.Error(), "membership is not active") {
```
pavitrabhalla: `strings.Contains` on error messages is fragile — a one-word change to the `fmt.Errorf` on line ~1056 silently breaks this 409 mapping. Define `var errMembershipNotActive = errors.New(...)` and use `errors.Is` here (same issue at line 964).

internal/api/rbac_phase2b.go
```diff
@@ -0,0 +1,1247 @@
...
+	RoleName  string
+	TeamID    uuid.UUID
+	GrantedBy pgtype.UUID
+	GrantedAt time.Time
+	RevokedAt *time.Time
+	CreatedAt time.Time
+	UpdatedAt time.Time
+}
+
+func (h *Handlers) rbacService() *authz.Service {
```
pavitrabhalla, line 87: `rbacService()` here and `authzService()` in `handlers.go` both return `authz.New(h.Pool)` — identical logic, two names. Can we pick one and remove the other?

internal/api/handlers_billing.go
```diff
@@ -49,6 +49,27 @@ func (h *Handlers) GetBillingPricing(c *gin.Context) {
 		respondError(c, ErrUnauthorized)
 		return
 	}
+	actorID := actorIDFromContext(c)
+	if actorID == nil {
+		respondError(c, ErrUnauthorized)
+		return
```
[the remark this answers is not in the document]
awille-ai: I have already run a manual backfill process and checked to make sure that all data has a created_by column that is non-null.

internal/api/rbac_phase2b.go
```diff
@@ -0,0 +1,1311 @@
...
+			  AND revoked_at IS NULL
+		)
+	`, teamID, userID).Scan(&exists)
+	if err != nil {
+		return false, err
+	}
+	return exists, nil
+}
+
+func hasPrivilegedRoleGrantHistory(ctx context.Context, q db.DBTX, teamID, userID uuid.UUID) (bool, error) {
```
pavitrabhalla: No filter on `revoked_at` — this returns true for any historical `team_owner`/`team_admin` grant, including ones long since revoked. So reactivating a former admin permanently requires `roles:write`, even if their current active grants are just `viewer`. A `user_admin` can never re-add a former admin without help. Is that the intended lifetime behavior, or should this be bounded (e.g., only revoked grants within some window)?

meAmitPatil:
PR #143 — RBAC Phase 2b review

Confirmed, worth fixing

1. GET /internal/teams/{id}/members is permanently broken (always 403). listMembers hardcodes "users:read" regardless of scope (rbac_phase2b.go:482), unlike every sibling handler which branches to a platform:-prefixed string. On the platform path it calls RequirePlatformPermission(actor, "users:read"); CanPlatform rejects any non-platform: permission with ErrScopeMismatch → 403. No platform admin can ever list members. Flagged by 4 of my agents and by Codex (still open). Fix: branch to "platform:teams:read" like listAssignments does (note: platform:team_users:read doesn't exist in the seed — platform:teams:read is the read perm).

2. RequirePlatformAdminGoogleSession is defined and unit-tested but never called by any handler (authz/service.go:156, zero non-test call sites — confirmed by grep). Platform-admin mutations are gated only by the INTERNAL_API_TOKEN, a DB platform-permission check, and a caller-supplied X-Actor-User-Id. The @superserve.ai Google-session gate the function implements is not enforced at runtime. Either wire it into the internal handlers/middleware, or confirm it's deliberately deferred — right now it's a security control that exists only on paper.

3. Cascade role revocations aren't audited. When deactivateMember sets status='inactive', the Phase 1 trigger trg_team_memberships_revoke_assignments revokes all the user's team role assignments — but the Go code writes only a team_member_deactivated audit event, no per-role team_role_revoked events. For an authz system the audit log should reflect the access that was actually removed. (Same gap on addMember's active→invited downgrade.)

OpenAPI defects (Codex-flagged, still open — can break codegen)

4. Seven paths $ref #/components/responses/Forbidden, but that component is never defined (api/openapi.yaml defines only Unauthorized/TooManyRequests/InternalError/NotFound/Conflict). Unresolved ref. 5. Both POST endpoints lack a requestBody schema, and the team paths omit the 404/409 responses the handlers actually return.

meAmitPatil (approved):

PR #144, opened 2026-06-23
controlplane: quiet shutdown-time DB error noise

meAmitPatil:
Cancel background reconcilers on SIGTERM before draining HTTP so they stop querying a DB that's being torn down, and drop benign `context canceled`/`closed pool` errors from Sentry (real timeouts/connection failures still alert).

### DIFF
--- a/cmd/controlplane/main.go
+++ b/cmd/controlplane/main.go
@@ -259,6 +259,10 @@ func run() error {
 		}
 	}
 
+	// Stop background reconcilers before draining HTTP so they don't keep
+	// issuing DB queries (which time out and log as errors) while shutting down.
+	cancel()
+
 	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 15*time.Second)
 	defer shutdownCancel()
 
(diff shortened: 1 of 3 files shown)


awille-ai (approved):
Reviewed for correctness, security, scalability, edge cases, and test coverage.

The change looks sound overall. Canceling the process-lifetime context before HTTP shutdown should stop the background reconcilers/rate-limit cleanup loops before the DB pool starts tearing down, while in-flight HTTP handlers still use request-scoped contexts during the drain. The Sentry writer keeps timeout/connectivity failures visible, and CI is green across unit tests, integration tests, vet, build, sqlc, and migration-order checks.

Non-blocking caveat: the new suppression is intentionally global and string-based. That is probably fine for this PR, but if `closed pool` ever appears outside intentional shutdown, Sentry would no longer catch it. If this grows beyond shutdown-noise cleanup, I’d prefer pushing this closer to the known shutdown/log sites or matching typed errors where available.

No high or critical findings from this review.

PR #145, opened 2026-06-23
Aw/useradmin/phase3

awille-ai:
Phase 3: Protect APIs

Start with highest-risk endpoints:

Role assignment
User management
Billing writes
Billing reads
Team settings

### DIFF
--- a/internal/api/rbac_phase3.go
+++ b/internal/api/rbac_phase3.go
@@ -1,0 +1,87 @@
+package api
+
+import (
+	"fmt"
+
+	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
+	"github.com/rs/zerolog/log"
+)
+
+// customerTeamPermissionAllowed checks an active team-scoped permission without
+// writing a response. Use this for optional capability fields on otherwise
+// readable endpoints.
+func (h *Handlers) customerTeamPermissionAllowed(c *gin.Context, teamID uuid.UUID, permission string) (bool, error) {
+	if h == nil {
+		return false, fmt.Errorf("nil handlers")
+	}
+
+	// DBTX-only handler unit tests do not have the Phase 1 RBAC tables or a real pool.
+	// Production routers always provide h.Pool, so this keeps old unit tests focused
+	// on handler behavior without weakening real requests.
+	// Unit tests must set h.Pool to get real RBAC enforcement.
+	if gin.Mode() == gin.TestMode && h.Pool == nil {
+		return true, nil
+	}
+
+	actorID := actorIDFromContext(c)
+	if actorID == nil {
+		return false, nil
+	}
+
+	authzSvc := h.authzService()
+	if authzSvc == nil {
+		return false, fmt.Errorf("missing authz service")
+	}
+
+	return authzSvc.CanTeam(c.Request.Context(), *actorID, teamID, permission)
+}
+
+// requireCustomerTeamPermission gates a handler on an active team-scoped permission.
+// Callers should pass the team ID already resolved from request context.
+func (h *Handlers) requireCustomerTeamPermission(c *gin.Context, teamID uuid.UUID, permission string) bool {
+	allowed, err := h.customerTeamPermissionAllowed(c, teamID, permission)
+	if err != nil {
+		log.Error().
+			Err(err).
+			Str("team_id", teamID.String()).
+			Str("permission", permission).
+			Msg("RBAC team permission check failed")
+		respondError(c, ErrInternal)
+		return false
+	}
+	if !allowed {
+		respondError(c, ErrForbidden)
+		return false
+	}
+	return true
+}
+
+// requireTeamPermissionFromContext resolves team_id from the Gin context before
+// applying the supplied permission check.
+func (h *Handlers) requireTeamPermissionFromContext(c *gin.Context, permission string) bool {
+	teamID, err := teamIDFromContext(c)
+	if err != nil {
+		return false
+	}
+	return h.requireCustomerTeamPermission(c, teamID, permission)
+}
+
+// Sandbox access intentionally aliases settings permissions for Phase 3. Keep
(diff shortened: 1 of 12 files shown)


internal/integration/integration_test.go
```diff
@@ -63,6 +63,11 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
+	if err := resetTestSchema(ctx, testPool); err != nil {
```
awille-ai, line 66: This reset makes the integration suite destructive for whatever `DATABASE_URL` happens to point at. Since `TestMain` accepts an arbitrary `DATABASE_URL`, a local misconfiguration or CI secret mix-up would run `DROP SCHEMA public CASCADE` against that database before any migrations/tests fail. Can we add a fail-closed guard here before calling `resetTestSchema` (for example require the parsed DB name to end in `_test`/equal `sandbox_test`, or require an explicit `ALLOW_TEST_SCHEMA_RESET=1`) so this isolation step cannot wipe a non-test database by accident?

awille-ai (commented):
I found one reliability/safety issue worth addressing before merging. The RBAC guard additions themselves look consistent with the surrounding handlers from the code I reviewed.

awille-ai:
Pushed a follow-up commit (`8bab8c0`) that adds an integration-test safety guard before `TestMain` can reset the schema. It allows the default `sandbox_test` database and other `*_test` database names, or an explicit `ALLOW_TEST_SCHEMA_RESET=1` override for intentional resets.

internal/api/rbac_phase3.go
```diff
@@ -0,0 +1,101 @@
...
+// readable endpoints.
+func (h *Handlers) customerTeamPermissionAllowed(c *gin.Context, teamID uuid.UUID, permission string) (bool, error) {
+	if h == nil {
+		return false, fmt.Errorf("nil handlers")
+	}
+
+	// DBTX-only handler unit tests do not have the Phase 1 RBAC tables or a real pool.
+	// Production routers always provide h.Pool, so this keeps old unit tests focused
+	// on handler behavior without weakening real requests.
+	if gin.Mode() == gin.TestMode && h.Pool == nil {
```
pavitrabhalla, line 23: This is fine as a compat shim, but a future test author who forgets to set `h.Pool` will silently pass all RBAC checks with no indication why. Could add a comment: `// Unit tests must set h.Pool to get real RBAC enforcement.`

internal/api/rbac_phase3.go
```diff
@@ -0,0 +1,101 @@
...
+	teamID, err := teamIDFromContext(c)
+	if err != nil {
+		return false
+	}
+	return h.requireCustomerTeamPermission(c, teamID, permission)
+}
+
+// requireTeamRead/Write are small convenience wrappers for the most common
+// customer-scoped authorization checks.
+func (h *Handlers) requireTeamSandboxRead(c *gin.Context, teamID uuid.UUID) bool {
```
pavitrabhalla, line 73: `requireTeamSandboxRead` and `requireTeamSettingsRead` both check `settings:read` — they're the same function. Do sandboxes and settings need separate permissions eventually, or can one of these be removed now?

internal/api/rbac_phase3.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (h *Handlers) requireTeamSettingsRead(c *gin.Context, teamID uuid.UUID) bool {
+	return h.requireCustomerTeamPermission(c, teamID, "settings:read")
+}
+
+func (h *Handlers) requireTeamSettingsWrite(c *gin.Context, teamID uuid.UUID) bool {
+	return h.requireCustomerTeamPermission(c, teamID, "settings:write")
+}
+
+func (h *Handlers) requireTeamBillingRead(c *gin.Context, teamID uuid.UUID) bool {
```
pavitrabhalla: `requireTeamBillingRead`, `requireTeamBillingWrite`, `requireTeamRolesRead`, `requireTeamRolesWrite` are defined but not called anywhere in this PR. Placeholders for Phase 4, or dead code?

internal/integration/integration_test.go
```diff
@@ -2445,37 +2532,12 @@ func TestIntegration_ActivityLog_ActorAttributedToKeyCreator(t *testing.T) {
 }
 
 func TestIntegration_ActivityLog_ActorNullWhenKeyHasNoCreator(t *testing.T) {
```
pavitrabhalla: The test name says `ActorNullWhenKeyHasNoCreator` but the assertion is just that the request returns 403 — the original check that the activity record has a null actor is gone. Should we rename the test or restore the null attribution assertion somewhere?

meAmitPatil:
PR #145 — RBAC Phase 3 review

This is a clean PR. The guard placement is actually solid: all ~25 endpoints gated, reads use settings:read, writes use settings:write, teamID always comes from the authenticated API-key context (never path/body, so no IDOR), and guards run before any data fetch or VM resume. The scary stuff is fine. The findings are smaller:

Worth fixing

1. Compile blocker — seedTeamAndKeyWithRole is defined twice (integration_test.go:223 and :360, byte-identical). The integration package won't build, so the three new Phase 3 RBAC tests — the whole point of this PR — can't run. CI will catch it, but flagging because it nullifies the PR's own coverage. Delete the second copy.

Decide before merge (access matrix)

2. settings:* is the only permission family gating sandboxes, templates, and secrets. Per the Phase 1 seed, that means:
- billing_admin and user_admin get 403 on every sandbox/template/secret endpoint (neither holds settings:read). A billing admin can't list sandboxes to reconcile usage; a user admin can't see what exists.
- viewer can read secrets metadata, secret audit logs, and network logs (all settings:read). Confirm a viewer seeing secret/audit data is intended.

This is documented as a deliberate alias with a planned sandboxes:*/secrets:* split, so it's a known decision — just make sure the resulting matrix is the one you want, since it ships as the live behavior now.

meAmitPatil (approved):

PR #146, opened 2026-06-24
controlplane: sandbox quota — 80% owner email + count active-only (cap resume)

meAmitPatil:
Two related changes to the sandbox quota.

## 1. 80% owner email
Customer-facing email (Resend) alongside the existing internal Slack alert when a team crosses 80% of its sandbox limit.
- Per-channel de-dup (`quota_alert_state` gains `channel`) so Slack and email retry independently.
- Hysteresis (alert 80 / re-arm under 70) + 7-day cooldown to avoid inbox spam.
- Owner-only recipient; sandbox limit only. Gated on `RESEND_API_KEY` + `QUOTA_EMAIL_FROM` — no-op until set.

## 2. Quota counts active-only (paused no longer counts)
`team.active_sandbox_count` now counts only live, non-paused sandboxes (starting/active/pausing/resuming); paused frees a slot.
- Because paused frees a slot, **resume is now capped too** — resuming a paused sandbox (incl. transparent auto-resume on access) returns 429 when the team is at its active limit.
- Trigger predicate changed + counter backfilled under an exclusive lock.
- Verified end-to-end against real Postgres (integration suite incl. new `ResumeSandbox_QuotaBlocked`).

## Deploy
Apply both migrations before rolling out the build (channel PK widen; trigger swap). Email stays dark until the env vars are set.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -206,9 +206,9 @@ paths:
           description: |
             Rate-limited or quota reached. The `error.code` distinguishes:
             - `rate_limited` — request rate exceeded; retry after a short backoff.
-            - `too_many_sandboxes` — team has reached its sandbox count limit
-              (paused sandboxes count toward this cap; only DELETE frees a slot);
-              delete existing sandboxes or contact support to raise the cap.
+            - `too_many_sandboxes` — team has reached its active sandbox count
+              limit (paused sandboxes do not count; pause or delete a sandbox to
+              free a slot, or contact support to raise the cap).
           content:
             application/json:
               schema:
@@ -353,6 +353,12 @@ paths:
           $ref: "#/components/responses/Conflict"
         "401":
           $ref: "#/components/responses/Unauthorized"
+        "429":
+          description: "`too_many_sandboxes` — resuming would exceed the team's active sandbox limit; free a slot or contact support to raise it."
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
         "500":
           $ref: "#/components/responses/InternalError"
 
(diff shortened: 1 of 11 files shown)


pavitrabhalla:
## Review comments

### Important

**Cooldown: description says 7 days, code says 30**
`quotaAlertCooldown = 30 * 24 * time.Hour` — PR description says "7-day cooldown". Which is the intended value?

**No hysteresis implemented**
Description says "re-arm under 70" but `shouldRearm` is purely time-based — it clears the row after the cooldown regardless of current usage. The old behavior (re-arm when the team drops back below the alert threshold) is gone. Was the 70% floor intentionally dropped, or is this an implementation gap?

**Email subject doesn't match the trigger**
Subject is `"Verify your team on Superserve"` but this fires on quota usage, not a user-initiated verification step. A recipient who's never heard of a verification flow gets this email cold. Is "verify your team" intentional product framing, or should it read something like "You're approaching your sandbox limit"?

**No unit test for the per-channel fan-out**
The core correctness property of the channel split is that a Slack failure doesn't suppress the email claim and vice versa. `quotaWatchOnce` is tested via integration but there's no unit test exercising both notifiers together — e.g. one succeeds and one fails, verify claims are independent. This is the easiest place for a regression to hide.

---

### Minor

**Email body omits current usage** — `quotaEmailHTML` uses `a.Limit` ("we limit the number of active sandboxes to **X**") but never shows `a.Used` or `a.Pct`. Customer knows their cap but not where they currently stand.

**Migration idempotency check is fragile** — `pg_get_constraintdef(oid) LIKE '%channel%'` relies on Postgres formatting the constraint definition predictably. Checking `array_length(conkey, 1) = 3` against `pg_constraint` is more stable.

**`srv.Close()` inside loop without defer** — `TestEmailQuotaNotifierConfigErrorRetries` calls `srv.Close()` inside the for-loop body; an early `t.Errorf` would leak the server. `defer srv.Close()` or closing after the loop would be safer.

PR #149, opened 2026-06-26
vmd: resume from pause via UFFD backend, flag-gated

meAmitPatil:
Routes resume-from-pause through the UFFD memory backend (instead of the legacy File backend), behind `VMD_RESUME_UFFD` — **default off, so merging is a no-op until flipped**. Unifies resume with the create path. Also adds `snapcheck`, a page-by-page mem-snapshot comparator (unit-tested).

**Tested on staging** with the flag on: resume works and is state-preserving — a live process + its in-memory data, filesystem, and networking all survive pause→resume on `base` and `python-ml`.

### DIFF
--- a/cmd/snapcheck/main.go
+++ b/cmd/snapcheck/main.go
@@ -1,0 +1,112 @@
+// Command snapcheck compares two Firecracker guest-memory snapshot files page
+// by page and reports any pages that differ. It is the correctness gate for
+// incremental snapshots: a written page the change set misses makes the restored
+// VM's memory differ from the original — silent corruption this catches.
+package main
+
+import (
+	"bytes"
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+)
+
+func main() {
+	pageSize := flag.Int("page-size", 4096, "page size in bytes")
+	maxReport := flag.Int("max-report", 20, "max differing page offsets to print")
+	flag.Usage = func() {
+		fmt.Fprintln(os.Stderr, "usage: snapcheck [flags] <mem-a> <mem-b>")
+		flag.PrintDefaults()
+	}
+	flag.Parse()
+
+	if flag.NArg() != 2 {
+		flag.Usage()
+		os.Exit(2)
+	}
+	if *pageSize <= 0 {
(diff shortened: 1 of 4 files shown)


awille-ai (commented):
Architecture/security review focused on tenant safety, scaling, and rollout risk.

Summary: the change is appropriately flag-gated (`VMD_RESUME_UFFD=false` by default), so merging this should be operationally inert until the flag is flipped. I do not see a new obvious tenant isolation bypass in the diff: resume still reuses the instance namespace/network identity, and the UFFD path only activates when `EnsureVMSlot` returns a concrete tap device.

High severity before enabling the flag: the existing inline comment about UFFD backing-file immutability should be addressed before rollout. `RestoreSnapshotUffdInternalWithOverrides` keeps serving pages from `memPath` for the Firecracker process lifetime, while pause/resume state still points at the standard pause snapshot path. If a later pause, cleanup, or snapshot rewrite truncates or replaces that memory file while the VM has unfaulted pages, the guest can see corrupted memory or the next snapshot can become invalid. That is a data-integrity and availability risk for live sandboxes, and it can be hard to diagnose after the fact. I would make the UFFD resume backend use an immutable per-resume backing copy or move subsequent snapshots to distinct paths before enabling `VMD_RESUME_UFFD` outside a narrow test cohort.

Scaling implication: this can improve resume latency by moving memory load to demand paging, but it also shifts work into first-touch runtime I/O. Under concurrent resumes, cold-page bursts can become disk read amplification and tail-latency spikes inside tenants. Before widening rollout, I would add/verify metrics for UFFD page faults served, page fault latency, backend read errors, restore duration, and post-resume first-command latency, then roll out by host pool or small tenant cohort with a fast flag rollback.

Security note: I did not find a direct cross-tenant data exposure in this diff, but snapshot memory files contain tenant memory. The new longer-lived dependency on `memPath` means deletion, path reuse, permissions, and cleanup ordering become more security-sensitive than with eager File restore. Treat the backing file as live VM state until the resumed Firecracker process is gone.

Recommendation: proceed with changes, but do not approve or enable the flag broadly until the backing-file lifetime issue is fixed or explicitly mitigated. I am not requesting changes because the flag defaults off and I did not find a critical issue in the default behavior.

PR #150, opened 2026-06-26
vmd: add snapshot bit-identical verify (load-no-resume + re-snapshot)

meAmitPatil:
Adds the **bit-identical snapshot verify gate** — the correctness check for upcoming incremental snapshots. `VerifySnapshot` loads a paused snapshot **without resuming**, re-snapshots the frozen image, and `snapcheck` compares it to the original; any unexpected page diff = a lost/corrupted page. Exposed via vmd's local HTTP server (`POST localhost:9090/verify-snapshot/{id}`), behind `VMD_VERIFY_SNAPSHOT_ENABLED` (**default off**) and guarded to paused VMs only — dark and inert in prod by default. Stacked on #149.

**Tested on staging:** gate is sensitive and trustworthy — byte-flip control → `differing: 1`, two different VMs → `differing: 286`, and the no-resume baseline is a stable, deterministic **3-page floor** (virtio device-state restore, confirmed by byte-level diff), identical across `base`/`python-ml` and a memory workload.

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -754,7 +754,10 @@ func (m *Manager) VerifySnapshot(ctx context.Context, vmID string) (string, erro
 		tapDevice = netInfo.TAPDevice
 	}
 
-	// Throwaway Firecracker in the sandbox's existing env; stopped on the way out.
+	// Throwaway Firecracker in the sandbox's existing unit/rundir, stopped on the
+	// way out. Safe because the VM is paused (no live FC to disrupt), but it must
+	// not run concurrently with a resume of the same sandbox — fine for the
+	// debug/staging use this endpoint is gated to.
 	if _, err := m.startFirecrackerViaSystemd(ctx, vmID, socketPath, rootfsPath, inst.Config.BasePath, inst.Namespace); err != nil {
 		return "", fmt.Errorf("start firecracker for verify: %w", err)
 	}
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -702,6 +702,70 @@ func (m *Manager) restoreForResume(socketPath, snapshotPath, memPath string, net
 	)
 }
 
+// VerifySnapshot loads vmID's snapshot without resuming, re-snapshots the frozen
+// image to a temp file, and returns its path (compare with the original via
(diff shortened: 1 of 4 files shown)


internal/vm/manager.go
```diff
@@ -702,6 +707,79 @@ func (m *Manager) restoreForResume(socketPath, snapshotPath, memPath string, net
...
+
+	if err := LoadSnapshotNoResume(socketPath, snapshotPath, memPath, "eth0", tapDevice, ""); err != nil {
+		return "", err
+	}
+
+	verifyDir := filepath.Join(filepath.Dir(memPath), "verify")
+	if err := os.MkdirAll(verifyDir, 0o755); err != nil {
+		return "", fmt.Errorf("mkdir verify dir: %w", err)
+	}
+	memBPath := filepath.Join(verifyDir, "mem.snap")
```
awille-ai, line 774: **P2: Avoid reusing a fixed verify output for every run**

Every verify call for the same VM writes to the same `verify/vmstate.snap` and `verify/mem.snap` paths, and the endpoint allows concurrent or back-to-back calls. That creates two failure modes: concurrent verifies can write the same multi-GB files at once, and a later verify can overwrite `mem.snap` after an earlier caller receives the path but before `snapcheck` compares it. This also gives the debug endpoint an unbounded disk I/O amplification path when enabled. Please either serialize verification per VM and reject/return the in-flight result, or write to a unique per-run directory and clean it up after comparison.

awille-ai (commented):
Reviewed for security, scaling, and lifecycle safety.

I do not see a critical security finding in this diff: the endpoint is default-off and the local HTTP server still binds to 127.0.0.1. I am not approving because the existing unresolved thread about sharing the production `firecracker@<id>` unit with verify is a real operational blocker before this is safe to enable outside a tightly controlled staging gate: verify can race with resume and stop the VM that resume just restored.

I left one additional scaling/correctness comment about unbounded repeated verify calls and fixed output paths. Recommendation: proceed with changes after serializing per-VM verify/resume activity and making verification output either unique per run or internally compare-and-cleanup so callers do not race on `verify/mem.snap`.

PR #152, opened 2026-06-27
vmd: incremental + layered diff snapshots

meAmitPatil:
Reduces pause cost by writing incremental memory snapshots instead of a full image each time. Flag-gated by `VMD_INCREMENTAL_SNAPSHOT` (default off); no behavior change when off.

**Incremental pause:** when dirty-page tracking is armed on a UFFD resume, a pause writes a Diff (only the pages dirtied since resume, merged in place into the existing `mem.snap`) instead of a Full snapshot.

**Layered first pause:** a template-created VM arms tracking on create and records the template `mem.snap` as its base. The first pause then writes a small diff overlay (`mem.diff`) against the template instead of a full image, and resume loads layered (overlay if present, else the template base — served lazily). Later pauses merge into `mem.diff`. This requires the layered UFFD restore in the firecracker fork (superserve-ai/firecracker#13).

- `MemoryBackend.base_path` (model + swagger) threads the base through `/snapshot/load`.
- `VMInstance.BaseMemPath` (daemon-internal, persisted) tracks the layered base — no control-plane change.
- `PauseVM`: layered overlay when a base is set; else in-place diff when tracking is armed and `mem.snap` is the resume base; else Full. `shouldWriteDiff` is unit-tested.
- Backward compatible and opt-in; the layered path is inert unless the fork handler is present and the flag is on.

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -615,20 +615,40 @@ func (m *Manager) PauseVM(ctx context.Context, vmID, snapshotDir string) (snapsh
 	case layered:
 		memPath = overlayPath
 		baseMemPath = instBaseMem
+		// A first layered pass (VM loaded straight from the template base, not from
+		// this overlay) must start from a fresh overlay: its dirty set is the whole
+		// overlay, so any leftover mem.diff from a failed/un-finalized earlier pause
+		// would survive as stale page extents — CreateDiffSnapshot writes only
+		// dirtied pages in place and never truncates — and override the base on the
+		// next layered restore, corrupting guest memory. Remove it; if it can't be
+		// removed, fall back to Full. An accumulating pass (resumed from this same
+		// overlay) must instead keep it, so this only runs for the first pass.
+		usable := true
+		if instMemFile == instBaseMem {
+			if rerr := os.Remove(overlayPath); rerr != nil && !os.IsNotExist(rerr) {
+				log.Warn().Err(rerr).Str("path", overlayPath).Msg("pause: stale overlay removal failed; falling back to Full")
+				usable = false
+			}
+		}
 		// Write the base sidecar BEFORE the diff so an overlay never exists on disk
 		// without its base record (a stateless cross-host restore relies on it). If
 		// the sidecar can't be written, fall back to a Full rather than produce an
 		// unrecoverable overlay.
-		if serr := os.WriteFile(layeredBaseSidecarPath(memPath), []byte(baseMemPath), 0o644); serr != nil {
-			log.Warn().Err(serr).Msg("pause: layered base sidecar write failed; falling back to Full")
-			memPath, baseMemPath = fullPath, ""
-			if err := CreateSnapshot(socketPath, snapshotPath, memPath, "", SnapshotNormal); err != nil {
-				return "", "", m.handleVMError(vmID, fmt.Errorf("create snapshot: %w", err))
+		if usable {
+			if serr := os.WriteFile(layeredBaseSidecarPath(memPath), []byte(baseMemPath), 0o644); serr != nil {
+				log.Warn().Err(serr).Msg("pause: layered base sidecar write failed; falling back to Full")
+				usable = false
 			}
-		} else {
+		}
+		if usable {
 			log.Info().Str("snapshot_path", snapshotPath).Msg("pausing VM — creating layered diff snapshot")
 			if err := CreateDiffSnapshot(socketPath, snapshotPath, memPath); err != nil {
 				return "", "", m.handleVMError(vmID, fmt.Errorf("create layered diff snapshot: %w", err))
+			}
+		} else {
+			memPath, baseMemPath = fullPath, ""
+			if err := CreateSnapshot(socketPath, snapshotPath, memPath, "", SnapshotNormal); err != nil {
+				return "", "", m.handleVMError(vmID, fmt.Errorf("create snapshot: %w", err))
 			}
 		}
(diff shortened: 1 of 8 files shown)


internal/vm/manager.go
```diff
@@ -580,13 +684,66 @@ func (m *Manager) PauseVM(ctx context.Context, vmID, snapshotDir string) (snapsh
...
 	m.persistState(inst)
 	log.Info().Msg("VM paused")
 	return snapshotPath, memPath, nil
 }
 
+// shouldWriteDiff reports whether a pause may write a Diff instead of a Full.
+// Diff is correct only when tracking was armed this run AND memPath is the exact
+// base this VM resumed from AND that base exists; any miss falls back to Full.
+// Diff onto a non-base or untracked run would silently corrupt the snapshot.
+func shouldWriteDiff(incremental, dirtyTracked bool, memPath, resumeBasePath string, baseExists bool) bool {
```
pavitrabhalla: `baseExists` checks whether the VM's current `memPath` exists on disk, not a template/layered base. `snapExists` or `memFileExists` avoids the naming confusion with the layered base concept used elsewhere.

internal/vm/manager.go
```diff
@@ -1088,9 +1331,41 @@ func (m *Manager) restoreVMSnapshot(ctx context.Context, vmID, snapshotPath, mem
...
+		// its base reconstructed from the sidecar (else refuse — loading it
+		// standalone reads the base as zero holes); a template create instead arms
+		// tracking and records the template as the base. Layered needs resume-UFFD.
+		isTemplate := m.isTemplateMemPath(memPath)
+		canLayered := m.cfg.UffdEnabled && m.cfg.ResumeUffdEnabled
+		basePath := ""
+		armLayered := false
+		switch {
+		case hasSidecar:
+			if !canLayered {
```
pavitrabhalla: `canLayered` is `m.cfg.UffdEnabled && m.cfg.ResumeUffdEnabled`. We're inside `case useUffd:` which requires both flags plus tap presence, so `canLayered` is always `true` here — this `!canLayered` branch is unreachable. The outer `case overlayNeedsLayered && !(useUffd && m.cfg.ResumeUffdEnabled):` already handles the "has overlay but can't UFFD" case before we get here. Safe to remove?

internal/vm/manager.go
```diff
@@ -564,11 +588,91 @@ func (m *Manager) PauseVM(ctx context.Context, vmID, snapshotDir string) (snapsh
...
+		// unrecoverable overlay.
+		if usable {
+			if serr := os.WriteFile(layeredBaseSidecarPath(memPath), []byte(baseMemPath), 0o644); serr != nil {
+				log.Warn().Err(serr).Msg("pause: layered base sidecar write failed; falling back to Full")
+				usable = false
+			}
+		}
+		if usable {
+			log.Info().Str("snapshot_path", snapshotPath).Msg("pausing VM — creating layered diff snapshot")
+			if err := CreateDiffSnapshot(socketPath, snapshotPath, memPath); err != nil {
```
pavitrabhalla, line 651: If `CreateDiffSnapshot` fails here, the `.base` sidecar (written three lines above) stays on disk alongside a corrupt/partial overlay. On the next restore attempt, `readLayeredBase` succeeds and a layered restore is attempted against that partial data. The PR comment acknowledges crash-atomicity is out of scope — does `handleVMError` or the reconciler clean up the sidecar, or is that left to the operator?

pavitrabhalla (commented):
Good overall structure — the fallback-to-Full paths are thorough and the sidecar approach for cross-host/vmd-restart recovery is clean. A few items below.

**Test coverage gaps**
- No test for the `readLayeredBase` round-trip (write sidecar → read back → verify path).
- No test for the Full fallback when the stale overlay pre-clearing fails in the `layered` arm of `PauseVM`.

**Questions**
- `DirtyTracked = true` is set on `inst` (lines 1351, 1360) before `RestoreSnapshotUffdInternalWithOverrides` is called. If that call fails, the field is left `true` on the instance. Since a restore failure should kill the VM this presumably doesn't matter — but is there a path where the instance lingers with a stale `DirtyTracked=true`?
- `isOverlayMemFile` keys on the literal basename `mem.diff`. The sidecar check in `restoreVMSnapshot` provides an additional guard, but is the naming convention documented anywhere beyond `manager.go`?

pavitrabhalla (commented):
All three inline issues addressed, and both test gaps filled.

**Resolved**
- `baseExists` → `memFileExists` with a clarifying doc comment. Good.
- `!canLayered` dead code removed; replaced with an inline comment explaining why the outer guard makes it redundant.
- `CreateDiffSnapshot` failure now drops the `.base` sidecar before returning, with a clear comment on why the overlay itself is left in place. The stale-`DirtyTracked` question is also answered: it's cleared in the restore error path.

**New: `freshenFirstPassOverlay`**
Extracting this into a named function is a nice improvement — `TestFreshenFirstPassOverlay` now covers missing, present, and un-removable cases.

**One minor nit**
The `TestShouldWriteDiff` struct field is still named `baseExists` (was the old parameter name); the function param is now `memFileExists`. No functional impact — just a test-internal field name.

**Open**
The `isOverlayMemFile` / `mem.diff` naming convention still lives only in `manager.go`. Not blocking, but worth a line in an internal doc or CLAUDE.md so the convention is findable before someone names a standalone snapshot `mem.diff` in a future path.

pavitrabhalla (approved):

PR #154, opened 2026-06-29
vmd: incremental memory snapshots — append-only chain, async pause, flatten/GC

meAmitPatil:
Server-side support for incremental, off-critical-path memory snapshots. **All behind flags, default off** — a flags-off vmd behaves identically to today, and new wire fields are omitted so an older Firecracker binary sees unchanged requests. Pairs with the companion Firecracker changes (layered chain loader + async write + memfd bridge).

## Append-only layered chain (`VMD_LAYER_APPEND`)
Immutable per-pause diff layers + an atomic manifest (`{base, overlays[], vmstate}`, published via temp-write + fsync + rename — the single commit point for the chain *and* its vmstate). Each pause appends a fresh `mem.N.diff`; restore composes `base → diff₁ → … → diffₙ` (newest-wins). An interrupted write can't corrupt a prior layer — the committed manifest still names the intact chain.

## Async pause (`VMD_ASYNC_PAUSE`)
A layered pause returns at the snapshot point and writes the diff on a Firecracker background thread; a goroutine awaits the write + fsync, commits the manifest (the durability point), then stops Firecracker. Resume / re-pause / delete wait for an in-flight flush so they only act on the committed chain. The first pause of a chain stays synchronous.

## Memfd-bridge fast resume (`VMD_SHARED_MEM`)
With memfd-backed guest RAM, a pause captures the memfd, snapshots vmstate + dirty offsets (no memory dump), stops the old Firecracker (the memfd keeps the RAM resident), and copies the dirty pages from the held memfd in the background. A resume *during* that flush serves faults from the memfd as the newest overlay instead of waiting for the disk write — so pause and resume stay ~constant regardless of dirty-set size. `VMD_MAX_BRIDGE_MEMFDS` bounds the extra RAM held fleet-wide.

## Flatten + GC (`VMD_LAYER_FLATTEN`)
Merges the overlay chain into one sparse overlay (newest-wins, shared base preserved) off the hot path once it reaches a depth or byte threshold (`VMD_FLATTEN_CHAIN_DEPTH` / `_BYTES`). Crash-safe (fresh output name + atomic manifest swap = commit point); orphan-GC reclaims interrupted-flatten leftovers and stale temps.

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -144,8 +144,9 @@ type VMInstance struct {
 
 	// holdingBridgeMemfd is true between a memfd fast-resume and the resumed VM's next
 	// pause/stop — the window where this sandbox pins ~2× its RAM (the new FC's mmap of
-	// the prior memfd). Counted against MaxBridgeMemfds. Not persisted; a vmd restart
-	// drops the memfd, so the count resets to zero on its own.
+	// the prior memfd). Counted against MaxBridgeMemfds. Persisted (VMRecord) so a
+	// reattached holder still counts after a vmd restart — its FC keeps the memfd mapped
+	// independently of vmd, so the RAM stays pinned; cleared at the next pause.
 	holdingBridgeMemfd bool
 
 	// restoring is true while a resume/verify is starting a Firecracker and loading a
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -702,12 +702,10 @@ func (m *Manager) PauseVM(ctx context.Context, vmID, snapshotDir string) (snapsh
 	// and doesn't race the background writer, FC stop, or layer deletion.
 	_ = m.waitForFlush(inst)
 	m.waitForFlatten(inst)
-
-	// If this VM was fast-resumed from a bridge memfd, pausing now stops that FC and
-	// frees the pinned RAM — drop the hold so it stops counting against MaxBridgeMemfds.
-	inst.mu.Lock()
-	inst.holdingBridgeMemfd = false
-	inst.mu.Unlock()
+	// A prior memfd fast-resume's hold (holdingBridgeMemfd) is dropped only once this pause has
+	// actually stopped that Firecracker — see clearBridgeHold at each stop point below. Clearing
+	// it here at entry would undercount the cap whenever the pause then fails with the FC (and
+	// its memfd mapping) still alive.
 
 	if snapshotDir == "" {
 		snapshotDir = filepath.Join(m.cfg.SnapshotDir, vmID)
@@ -840,6 +838,7 @@ func (m *Manager) PauseVM(ctx context.Context, vmID, snapshotDir string) (snapsh
 				if err := stopUnit(ctx, systemdUnitName(vmID)); err != nil {
 					log.Warn().Err(err).Msg("stop FC after bridge snapshot")
 				}
+				m.clearBridgeHold(inst) // prior fast-resume's FC is stopped; release its memfd hold
 				// Mark flushing BEFORE status=paused (same ordering as the in-FC async path)
 				// so the reconciler never mistakes the flush window for a stranded FC.
 				m.beginFlush(inst)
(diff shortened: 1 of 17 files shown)


internal/vm/manager.go
```diff
@@ -696,10 +1004,128 @@ func (m *Manager) PauseVM(ctx context.Context, vmID, snapshotDir string) (snapsh
...
+	manifest.Overlays = append(manifest.Overlays, layerName)
+	manifest.Vmstate = vmstateName
+	if err := writeManifestAtomic(snapshotDir, manifest); err != nil {
+		failClean(err, "bridge pause: manifest commit failed; keeping previous durable snapshot")
+		return
+	}
+	log.Info().Int("pages", len(offsets)).Msg("VM paused (memfd bridge flush durable)")
+	// Arm/launch flatten BEFORE endFlush (see completeAsyncPause) so a woken resume/delete
+	// observes the pending flatten instead of racing the manifest swap.
+	m.maybeFlatten(inst, snapshotDir)
```
awille-ai, line 1136: **P1: prevent bridge fast-resume from racing flatten.** A memfd fast-resume deliberately skips `waitForFlush`, then reads the currently committed manifest and builds `lowerOverlays` from the pre-commit chain while using the held memfd as the newest layer. If this bridge completion reaches `maybeFlatten` before that restore has finished, flatten can swap the manifest and delete the old overlay files that the in-progress restore is still using as lower layers. That makes the fast path race exactly at the scale boundary where flatten is triggered, causing intermittent restore failures or missing-layer reads under concurrent pause/resume. Either make bridge fast-resumes participate in the flatten guard before they read the manifest, or defer flatten while a bridge memfd has been claimed / held for restore.

awille-ai (commented):
Architecture/security review pass on current head `2a9d36c89666529655451da080a428b4c2c58b2e`. I focused on scaling, security, parallelism, and performance. I did not approve because there are still high-impact concurrency/correctness findings on the PR, including the existing unresolved comments around stateless restore / verify / reconciler behavior plus the additional memfd/flatten race below. I am not requesting changes because I did not classify the new finding as security-critical by itself, but I would want it addressed before enabling these flags broadly.

pavitrabhalla:
**<sub><sub>P2 Badge</sub></sub>  `flattenInBackground` does not update `inst.MemFilePath` — `VMD_LAYER_APPEND` + `VMD_LAYER_FLATTEN` break each other after first flatten**

After `flattenChain` commits a merged overlay (e.g. `mem.flat.N.diff`), `inst.MemFilePath` still holds the pre-flatten newest layer path (e.g. `mem.M.diff`). On the next `PauseVM`, `planLayerAppend` resolves the manifest to `mem.flat.N.diff` and compares against the cached `inst.MemFilePath` — mismatch, `ok=false`. The pause falls back to the standalone/full path, which then calls the manifest-removal cleanup block and discards the entire append chain. The two flags are mutually self-defeating after the first successful flatten. Fix: at the end of the success branch in `flattenInBackground`, update `inst.MemFilePath` to `filepath.Join(snapshotDir, outName)` under `inst.mu`.

Useful? React with 👍 / 👎.

pavitrabhalla:
**<sub><sub>P1 Badge</sub></sub>  `dirtyOffsetsPath` not removed on `CreateDiffSnapshotBridge` failure**

The error cleanup in the bridge-pause path removes `layerPath` and `vmstatePath` but not `dirtyOffsetsPath`. If a subsequent bridge pause succeeds, Firecracker writes a new offsets file but — depending on open flags — may not truncate first; `readDirtyOffsets` in `completeBridgePause` can then read stale offsets from the prior failed attempt, causing `dumpMemfdPages` to copy wrong dirty pages into the new layer. The manifest commits and the sandbox resumes with silently-corrupted guest memory at the stale offsets. Add `_ = os.Remove(dirtyOffsetsPath)` to the `CreateDiffSnapshotBridge` error cleanup alongside the existing removes.

Useful? React with 👍 / 👎.

pavitrabhalla:
**<sub><sub>P1 Badge</sub></sub>  Panic after last `failClean` in `completeAsyncPause`/`completeBridgePause` permanently blocks flush waiters**

Both goroutines call `endFlush` only on the explicit success tail and in `failClean` for explicit early-return error branches. Neither defers `endFlush`. A panic from `stopAsyncPauseFC`, `maybeFlatten` (which reads the manifest from disk), or `clearBridgeHold` after the last `failClean` site fires other defers (`<-m.pauseSem`, memfd close) but leaves `flushDone` unclosed. Every caller of `waitForFlush` for that instance — `PauseVM`, `ResumeVM`, `DeleteSnapshotFiles` — blocks permanently; the sandbox is unserviceable until vmd restarts. Restructure so `endFlush` is unconditionally deferred (with a sentinel error variable updated at each failure branch) rather than called on the success tail.

Useful? React with 👍 / 👎.

pavitrabhalla:
**<sub><sub>P2 Badge</sub></sub>  `mergeOverlayPages` silently skips pages when an overlay is shorter than the base**

`mergeOverlayPages` bounds the SEEK_DATA/SEEK_HOLE scan of each overlay by the overlay's own `info.Size()`, while the output is sized to `baseInfo.Size()`. An overlay shorter than the base — from a partial write or truncation bug — leaves `written[idx]` false for the out-of-range pages; those slots are filled by the next-older overlay's data instead. If the short overlay is the newest and owns pages in that range, the merged result has stale or absent guest-memory data at those offsets with no error surfaced. Add a size check before the merge loop and return an error if any overlay's size does not match the base.

Useful? React with 👍 / 👎.

pavitrabhalla:
**<sub><sub>P2 Badge</sub></sub>  `captureGuestMemFd` matches by name only — a second `guest_mem` memfd yields the wrong fd**

The scan returns the first fd whose readlink target equals `/memfd:guest_mem (deleted)`. If Firecracker or any library it links creates a second memfd with the same name, the lower-numbered fd is selected regardless of content. `dumpMemfdPages` then reads dirty-page offsets from the wrong memory object and writes them into the new layer; the manifest commits and the sandbox resumes with guest-memory pages from unrelated memory at every dirty-page offset. After matching by name, validate the candidate's apparent size against the expected guest RAM size (derivable from the max dirty offset or a config field) and continue scanning if it doesn't match.

Useful? React with 👍 / 👎.

PR #157, opened 2026-07-01
Aw/billingsummary

awille-ai:
Adds proper billing summary for users

### DIFF
--- a/internal/api/handlers_billing.go
+++ b/internal/api/handlers_billing.go
@@ -1,14 +1,19 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"time"
 
 	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/rs/zerolog/log"
+	"golang.org/x/sync/errgroup"
 
+	"github.com/superserve-ai/sandbox/internal/billing"
 	"github.com/superserve-ai/sandbox/internal/db"
 )
 
@@ -39,35 +44,41 @@ type billingPricingResponse struct {
 	Rates    []billingPricingRateResponse `json:"rates"`
 }
 
+type billingSummaryResponse struct {
+	CurrentChargesUSD        float64                     `json:"current_charges_usd"`
+	CreditsAppliedUSD        float64                     `json:"credits_applied_usd"`
+	CreditsRemainingUSD      float64                     `json:"credits_remaining_usd"`
+	ExpectedInvoiceAmountUSD float64                     `json:"expected_invoice_amount_usd"`
+	CostBreakdownUSD         billingSummaryCostBreakdown `json:"cost_breakdown_usd"`
+	BillingPeriod            billingSummaryPeriod        `json:"billing_period"`
+	PricingTier              billingSummaryPricingTier   `json:"pricing_tier"`
+	CalculatedAt             time.Time                   `json:"calculated_at"`
+}
+
+type billingSummaryCostBreakdown struct {
+	Compute float64 `json:"compute"`
+	Memory  float64 `json:"memory"`
+	Storage float64 `json:"storage"`
+}
+
+type billingSummaryPeriod struct {
+	Start time.Time `json:"start"`
+	End   time.Time `json:"end"`
+}
+
+type billingSummaryPricingTier struct {
+	PlanKey  string `json:"plan_key"`
+	PlanName string `json:"plan_name"`
+	Currency string `json:"currency"`
+}
+
 func (h *Handlers) GetPublicBillingPricing(c *gin.Context) {
 	h.writeBillingPricing(c, publicPricingPlanKey, "public", publicPricingCacheControl)
 }
 
 func (h *Handlers) GetBillingPricing(c *gin.Context) {
-	teamID, err := teamIDFromContext(c)
-	if err != nil {
-		respondError(c, ErrUnauthorized)
-		return
-	}
-	actorID := actorIDFromContext(c)
-	if actorID == nil {
-		respondError(c, ErrUnauthorized)
-		return
-	}
-
-	authzSvc := h.authzService()
-	if authzSvc == nil {
-		respondError(c, ErrInternal)
-		return
-	}
-	allowed, err := authzSvc.CanTeam(c.Request.Context(), *actorID, teamID, "billing:read")
-	if err != nil {
-		log.Error().Err(err).Str("team_id", teamID.String()).Str("user_id", actorID.String()).Msg("RBAC billing read check failed")
-		respondError(c, ErrInternal)
-		return
-	}
-	if !allowed {
-		respondError(c, ErrForbidden)
+	teamID, ok := h.requireBillingRead(c)
+	if !ok {
 		return
 	}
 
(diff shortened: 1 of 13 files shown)


internal/api/handlers_billing.go
```diff
@@ -218,3 +405,10 @@ func numericFloat64(n pgtype.Numeric) (float64, error) {
 	}
 	return v.Float64, nil
 }
+
+func minFloat(a, b float64) float64 {
```
pavitrabhalla: `math.Min` from the stdlib does the same thing — the test already uses it. Drop `minFloat`.

internal/api/handlers_billing.go
```diff
@@ -39,46 +40,232 @@ type billingPricingResponse struct {
...
-		respondError(c, ErrForbidden)
+	pricingRates := pricingRatesFromTeamCurrentRows(rates)
+	if len(pricingRates) == 0 {
+		log.Error().Str("team_id", teamID.String()).Msg("billing pricing plan has no active rates for summary")
+		respondPricingUnavailable(c)
 		return
 	}
+	rateByResource := make(map[string]billingPricingRate, len(pricingRates))
+	for _, rate := range pricingRates {
+		if _, ok := requiredPricingResources[rate.Resource]; !ok || rate.Unit != "second" {
```
pavitrabhalla: This bails if the plan has any rate row outside `requiredPricingResources`. If a new resource type (e.g. bandwidth) is added to the pricing table later, this endpoint returns 503 even though compute/memory/storage rates are all present. Should `continue` on unknown resources rather than bail — the completeness check below already catches missing required ones.

internal/api/handlers_billing.go
```diff
@@ -39,46 +40,232 @@ type billingPricingResponse struct {
...
+	if !ok {
 		return
 	}
 
-	authzSvc := h.authzService()
-	if authzSvc == nil {
+	now := time.Now().UTC()
+	periodStart, periodEnd := currentBillingPeriod(now)
+
+	usage, err := h.DB.GetTeamBillingUsage(c.Request.Context(), db.GetTeamBillingUsageParams{
```
pavitrabhalla: Does `GetTeamBillingUsage` include open intervals (sandboxes currently running without an `ended_at`)? The integration test only exercises closed intervals (`end_reason = 'paused'`). Want to confirm live sandboxes contribute to the summary.

internal/api/handlers_billing.go
```diff
@@ -39,46 +40,232 @@ type billingPricingResponse struct {
...
+	}
+
+	breakdown := billingSummaryCostBreakdown{
+		Compute: vcpuSeconds * vcpuRate,
+		Memory:  memoryGibSeconds * memoryRate,
+		Storage: storageGibSeconds * storageRate,
+	}
+	currentCharges := breakdown.Compute + breakdown.Memory + breakdown.Storage
+
+	creditBalance, err := h.DB.GetTeamCreditBalance(c.Request.Context(), teamID)
```
pavitrabhalla: Is the balance returned here net of prior charges already invoiced, or the gross grant total? The `remaining_usd` column name suggests it's already reduced, but worth confirming — if it's the gross grant, applying it against current-period charges would double-count credits consumed by past invoices.

internal/api/handlers_billing.go
```diff
@@ -39,46 +40,232 @@ type billingPricingResponse struct {
...
+	}
+	creditsAvailable, err := numericFloat64(creditBalance)
+	if err != nil {
+		log.Error().Err(err).Str("team_id", teamID.String()).Msg("convert credit balance failed")
+		respondError(c, ErrInternal)
+		return
+	}
+	creditsApplied := minFloat(currentCharges, creditsAvailable)
+
+	c.Header("Cache-Control", "private, max-age=60")
```
pavitrabhalla: Other handlers use named constants (e.g. `privatePricingCacheControl`). `"private, max-age=60"` inline is inconsistent — worth a constant if the TTL is reused.

internal/integration/integration_test.go
```diff
@@ -537,6 +537,129 @@ func TestIntegration_GetBillingPricing(t *testing.T) {
...
+	}
+	if tier["plan_key"] != "payg" || tier["plan_name"] != "Pay-as-you-go" || tier["currency"] != "USD" {
+		t.Fatalf("pricing tier = %v, want payg/Pay-as-you-go/USD", tier)
+	}
+	if _, ok := body["calculated_at"].(string); !ok {
+		t.Fatalf("calculated_at missing or not a string: %v", body["calculated_at"])
+	}
+}
+
+func TestIntegration_GetBillingSummaryDeniesUserAdmin(t *testing.T) {
```
pavitrabhalla, line 822: No test for the 503 path (team with no active pricing rates). The empty-rates branch is the most likely misconfiguration scenario — worth a case here.

meAmitPatil:
## Review

The auth extraction into `requireBillingRead` is faithful to the old inline block, units line up end to end (vcpu-seconds / GiB-seconds vs per-second rates), and the no-rows/NULL paths are safe via `COALESCE`. Findings below, most severe first.

### 1. Credits are never decremented, so they re-apply every billing period — `internal/api/handlers_billing.go:222`
`ApplyTeamCreditGrant` and `RecordTeamCreditLedgerEntry` have zero callers anywhere in the codebase, so `team_credit_grant.remaining_usd` is only ever set at grant time. The comment above `creditsApplied` ("net of any credit already consumed by prior billing application") doesn't hold — nothing consumes credit. A team with a one-time $100 grant spending $40/month sees `credits_applied_usd=40, expected_invoice_amount_usd=0` in July, and the exact same thing in August, indefinitely. Either the summary needs to subtract charges already credited in prior periods, or credit application needs to be wired up before this number is trustworthy.

### 2. Unknown priced resources are silently dropped; `/billing/pricing` 503s on the same state — `internal/api/handlers_billing.go:130`
The rate loop `continue`s past any resource outside `{vcpu, memory_gib, storage_gib}`, while `writeBillingPricingRates` responds `pricing_unavailable` for the identical condition. If a plan ever gains a fourth billable rate, `/billing/pricing` returns 503 but `/billing/summary` returns 200 with that resource's cost silently excluded — an understated bill with no error surfaced. The validation is duplicated rather than shared; extracting one validated-rate-map helper used by both endpoints would fix the divergence and prevent future drift.

### 3. Billing period is hardcoded to UTC calendar months, ignoring `team_billing_period` — `internal/api/handlers_billing.go:271`
`team_billing_period` stores arbitrary `(period_start, period_end)` windows (only `CHECK (period_end > period_start)`), and the finalize/export/anomaly queries all key on that tuple. `currentBillingPeriod` independently assumes calendar months, so once periods are actually finalized the summary window can disagree with the invoiced window. Latent today (nothing writes `team_billing_period` in production yet), but worth deriving the window from the same source before invoices exist.

### 4. The `tenant_usage_dashboard` flag doesn't gate this endpoint — `internal/api/handlers_billing.go:92`
The migration in this PR enables the flag to grant billing visibility, but `GetBillingSummary` never checks `feature_enabled('tenant_usage_dashboard', team_id)` — only RBAC. A per-team `team_feature_flag` disable (the override mechanism exists and takes precedence) hides the dashboard but leaves `/billing/summary` fully open. If the flag is meant to be frontend-only, fine — but nothing documents that, and the per-team off switch silently not applying to the API is surprising.

### 5. Charge math lives in the HTTP handler — `internal/api/handlers_billing.go:201`
`internal/billing` owns the usage pipeline but contains no pricing math; usage×rate and the credit `min()` appear for the first time inside the handler. When finalize/export grow real charge computation, it will be implemented separately and the provisional summary can diverge from the actual invoice. Consider moving the computation next to the usage pipeline so both consumers share it.

### 6. Tautological test assertion — `internal/integration/integration_test.go:693`
`currentCharges` is read from `body["current_charges_usd"]` at line 690, then line 693 asserts that same field equals `currentCharges` — it can never fail. The line-692 sum check is the real assertion; delete this one.

### 7. Three independent queries run sequentially — `internal/api/handlers_billing.go:101`
`GetTeamBillingUsage`, `ListActivePricingRatesForTeamCurrent`, and `GetTeamCreditBalance` share no data dependency but run back to back, so latency is the sum of three round-trips instead of the max. `errgroup` is already a dependency and the package already fans out concurrent work elsewhere.

### 8. Test re-derives the billing period inline — `internal/integration/integration_test.go:592`
Lines 592–593 duplicate `currentBillingPeriod` byte for byte (it's unexported, so the test can't call it). If the period definition ever changes, the test keeps asserting the stale calendar-month contract and passes anyway. Export the helper or move it somewhere shared.

### 9. Two unreachable 503 branches in the rate loop — `internal/api/handlers_billing.go:132`
The `rate.Unit != "second"` branch can't fire (`pricing_rate` has `CHECK (unit = 'second')`), and the duplicate-key branch can't either (the query partitions `row_number()` by `(resource, unit)` and keeps `rate_rank = 1`, so at most one row per key). Relatedly, the `resource + ":" + unit` map key's unit half is a forced constant — keying by resource alone removes the hardcoded `":second"` suffixes at every lookup. Suggest deleting both branches (or reducing to a comment citing the constraint and the `rate_rank=1` invariant).

### 10. Private cache headers duplicated — `internal/api/handlers_billing.go:224`
The `Cache-Control` + `Vary: X-API-Key` pair is hand-inlined here and emitted independently by `writeBillingPricingRates`. If the private caching policy changes at one site only, the other endpoint keeps the old policy — and a missed `Vary` on per-team billing data is the bad kind of drift. Factor into a shared helper.

---

Also verified and deliberately not flagged: float USD values and whole-period current-rate pricing are both explicitly documented as current scope in the spec/comments, and the live interval query (rather than the hourly rollup) is the right choice given the freshness the test asserts.

*Minor, below the cutoff: the six copy-paste `numericFloat64` error blocks could collapse into one helper; the test's month-boundary clamping can spuriously `t.Fatalf` if the clock lands on exactly 00:00:00 UTC on the 1st.*

meAmitPatil (approved):

PR #159, opened 2026-07-02
controlplane: make ActivateSandbox + FinalizePause fire-and-forget

pavitrabhalla:
Takes the post-VMD bookkeeping DB writes (`ActivateSandbox`, `FinalizePause`) off the hot path: create/pause/resume respond as soon as the VM transition physically happens, and the status flip lands in the background. First step of the two-region split — hot-path latency becomes gate write + VMD call regardless of where the bookkeeping DB lives.

**Technical decisions/tradeoffs**
- Safe by construction: every transition is a status-gated compare-and-swap (an UPDATE guarded on the expected current status, matching zero rows otherwise), and only the background write unlocks the next one — a racing request just 409s until it lands. Exec/get follow-ups additionally wait out a bounded settle window (2s, normally single-digit ms) on `starting`/`resuming` rows, so the canonical create→exec flow keeps read-your-writes; a mutating follow-up (pause right after create) may still need to retry on 409.
- Pause error paths change shape: deleted-mid-pause (was 410) and a failed finalize write (was 500) are now background log lines; the physical outcome is identical.
- Quota frees the slot at `BeginPause` instead of `FinalizePause` (pause→create at the cap would otherwise transiently 429). Pause reverts (`pausing→active`) are exempt from the cap: the VM never stopped, so the write only records reality — capping it would let the reaper mark a healthy VM `failed` when a create raced into the freed slot. Costs: cap+1 physical VMs for the seconds a snapshot takes (or until a reverted pause is retried), and a crash mid-pause leaves a running VM uncounted where it previously leaked the slot in the opposite direction — same stranded-transitional-row class, covered by the reconciler follow-up below.
- Create's egress-rule push stays blocking (201 still implies rules applied); reaper and error-path reverts keep their synchronous saga semantics.
- Known gap: a crash between the response and the background commit strands the row in a transitional state the reconciler doesn't repair. Pre-existing window, but wider now — reconciler repair of transitional rows is a follow-up before real multi-region traffic.

**Testing**
- New integration tests exercise the exact windows: pause gate write with no finalize then create at the cap (must 201), and a `pausing→active` revert at the cap (must not SS001). The `starting→active` settle is unit-tested; async failure paths are unit-tested.
- Bookkeeping goroutines are tracked (mutex+cond counter — WaitGroup forbids Add during Wait, which concurrent requests do) so tests can wait for quiescence; integration `do()` waits after every request.
- `go test -race ./internal/api/` and the full integration suite under `-race` pass locally; CI green.

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -72,6 +72,54 @@ type Handlers struct {
 	Analytics *analytics.Client // when set, emits product-usage events; nil is a no-op
 	Encryptor secrets.Encryptor // KMS envelope used by /secrets endpoints; nil disables them
 	Signer    *SecretsSigner    // signs sandbox JWTs and serves the JWKS; nil disables both
+
+	// asyncMu/asyncCond/asyncCount track fire-and-forget bookkeeping goroutines
+	// (ActivateSandbox, FinalizePause) so tests can wait for quiescence;
+	// production never waits. Not a sync.WaitGroup: concurrent requests Add
+	// while a test's Wait is in flight, which WaitGroup forbids once the
+	// counter can touch zero mid-wait (the race detector flags the reuse).
+	asyncMu    sync.Mutex
+	asyncCond  *sync.Cond // lazily created by WaitAsyncBookkeeping, guarded by asyncMu
+	asyncCount int
+}
+
+// asyncBookkeeping runs a fire-and-forget post-VMD bookkeeping DB write in a
+// background goroutine. These writes are off the hot path by design: the gate
+// write (BeginPause/BeginResume/create INSERT) already owns the row, and every
+// subsequent transition is status-gated, so nothing can proceed until the
+// bookkeeping lands — a racing request just sees the transitional status and
+// 409s until the write commits.
+func (h *Handlers) asyncBookkeeping(name string, fn func()) {
+	h.asyncMu.Lock()
+	h.asyncCount++
+	h.asyncMu.Unlock()
+	go func() {
+		defer func() {
+			h.asyncMu.Lock()
+			h.asyncCount--
+			if h.asyncCount == 0 && h.asyncCond != nil {
+				h.asyncCond.Broadcast()
+			}
+			h.asyncMu.Unlock()
+		}()
+		defer sentrylog.Recover(name)
+		fn()
+	}()
+}
+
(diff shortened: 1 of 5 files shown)


meAmitPatil:
### Code review

Found 2 issues:

1. Excluding `pausing` from the quota-counted set makes the `pausing→active` revert quota-capped, so an auto-paused-but-healthy sandbox can be marked terminally `failed`. Before this PR the previous migration counted `pausing`, so `pausing→active` was `was_counted && is_counted` — a no-op that could never fail. Now it hits the `NOT was_counted && is_counted` branch and can raise `SS001`. The reaper's revert paths (`revertToActiveOrFail`, `rollbackPausedVM`) treat that `UpdateSandboxStatus` error by calling `markSandboxFailed`, i.e. a terminal `failed` state needing "manual recovery." Sequence: reaper claims `active→pausing` (frees the slot) → a concurrent create at cap fills it → the VMD pause hits a transient error → the revert now `SS001`s → healthy VM's row is marked `failed`. This is worse than the outcome the migration comment documents ("the revert fails with SS001 and the row waits for stale-transition cleanup") — `failed` is terminal, not a transitional row a reconciler would repair, and no such reconciler exists yet. Auto-pause is the default lifecycle, so this path is routine.

https://github.com/superserve-ai/sandbox/blob/b0398886e6459843b2d048f68d8579c55df6fcdd/supabase/migrations/20260701000001_quota_exclude_pausing.sql#L76-L89

2. `CreateSandbox` returns `201` with `status: "active"` before the fire-and-forget `ActivateSandbox` write commits the row (still `starting`). A client that trusts the `201` and immediately calls exec/get hits `loadActiveOrResumeSandbox`, whose `default` case 409s on `starting`/`resuming` — a lost read-your-writes for the sandbox's own owner. The code comment frames the 409 as acceptable for "a racing pause/exec," but that framing doesn't cover the caller's own follow-up on a sandbox the response already reported as `active`. The window is small (the async write usually wins the race) but widens under DB contention.

https://github.com/superserve-ai/sandbox/blob/b0398886e6459843b2d048f68d8579c55df6fcdd/internal/api/handlers.go#L1814-L1842

https://github.com/superserve-ai/sandbox/blob/b0398886e6459843b2d048f68d8579c55df6fcdd/internal/api/handlers.go#L402-L413

meAmitPatil (approved):

PR #160, opened 2026-07-02
api: cache successful API-key lookups in-process

pavitrabhalla:
Caches successful X-API-Key lookups in-process so steady-state auth skips the per-request DB round trip — the largest hot-path cost with the control plane in us-west and the DB in us-east. Part of the two-region prep.

## Technical decisions

- Only positive lookups are cached: invalid/revoked keys always hit the DB, so revocation propagates within one TTL (default 30s, `API_KEY_CACHE_TTL`, `0` disables) and bad-key floods can't grow the map. Revocation happens out-of-band (Console writes the DB directly), so the TTL is the bound — no in-process invalidation is possible.
- `expires_at` is stored and checked on every hit, so key expiry stays exact inside the TTL window.
- The fire-and-forget `last_used_at` UPDATE — itself a hidden per-request DB write — is throttled to once per key per minute.

## Testing

9 new unit tests: hit/miss, TTL expiry, exact `expires_at` enforcement, touch throttling, disabled mode, capacity sweep/reset. Existing middleware tests pass unchanged.

### DIFF
--- a/internal/api/apikey_cache.go
+++ b/internal/api/apikey_cache.go
@@ -1,0 +1,116 @@
+package api
+
+import (
+	"os"
+	"sync"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgtype"
+)
+
+// apiKeyCache is an in-process, TTL-bounded cache of successful API-key
+// lookups, keyed by the key's SHA-256 hash. It removes the per-request DB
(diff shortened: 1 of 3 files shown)


meAmitPatil:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

meAmitPatil (approved):

PR #166, opened 2026-07-02
vmd: gate resume StatusRunning on boxd readiness

pavitrabhalla:
## Summary

`ResumeVM` marked the instance running the moment Firecracker restored the snapshot — before boxd inside the guest was accepting connections. The proxy then saw `running`, dialed a dead port, and exec/data-plane clients hit slow timeouts instead of a fast, retriable `503 sandbox is paused`. Under host pressure (this week's ARP-table overflow incident) that window stretched to 1.7–4.8s, tripping the SDK's ~2s exec timeout. Part of hardening the resume path after that incident.

## Technical decisions

- Mirrors the existing gate in the template-restore path (`restoreVMSnapshot`): poll boxd `/health` up to 5s before flipping status, same timeout.
- On timeout: stop the Firecracker unit and return an error without touching instance state — snapshot files are intact, the sandbox stays `paused`, and the resume is retriable. No rundir/network teardown, matching `ResumeVM`'s other post-start error paths.
- Logs `wait_boxd_ms` on success, matching the restore path's instrumentation, so resume-readiness latency is graphable.

## Testing

- `GOOS=linux go build ./...` and `go vet` pass. The vm package's tests are Linux-only (nftables dep) and don't exercise `ResumeVM`; relying on CI + staging resume soak for behavioral verification.

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -888,6 +888,20 @@ func (m *Manager) ResumeVM(ctx context.Context, vmID, snapshotPath, memPath stri
 		return nil, fmt.Errorf("restore snapshot: %w", err)
 	}
 
+	// Gate StatusRunning on boxd answering inside the guest. Firecracker
+	// resuming the vCPUs doesn't make the data plane reachable — under host
+	// pressure boxd's first response can lag by seconds, and reporting
+	// running early sends the proxy dialing a dead port, so clients see
+	// slow 5xx instead of a fast, retriable "sandbox is paused". The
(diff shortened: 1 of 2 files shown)


pavitrabhalla:
Parking this: the failure path (auto re-pause on gate timeout) can amplify host-stress incidents with snapshot-write storms, and bricks sandboxes whose in-guest boxd died. Shipping the readiness telemetry separately first; will revisit gating with real distribution data.

PR #172, opened 2026-07-02
vmd: serve during startup with on-demand reattach instead of blocking on full-fleet reattach

meAmitPatil:
On restart, vmd reattached every VM in BoltDB before binding the gRPC port, so a large host spent minutes not listening — every control-plane call during that window failed with `connection refused`. Startup work scaled with fleet size and blocked serving.

This binds and serves gRPC immediately, gated with a retriable `Unavailable` until fast pre-serve init (slot reservation + orphan-namespace sweep + pool prime) completes. The full per-VM reattach moves to a background pass, and any VM a request touches before that pass reaches it is reattached on demand (singleflight, shared with the eager path). Slot indices held by existing VMs are reserved before the pool starts, so the pool never hands out a colliding slot.

Correctness details:
- Liveness uses `unitDefinitelyDead`, which treats a cancelled/timed-out/errored `systemctl` as "unknown, don't destroy" — so a shutdown or an overloaded host can't mistake a live VM for a dead one and SIGKILL + delete it. The eager loop also bails on context cancellation.
- Resume loads its VM before the in-place check, so it reuses the existing slot/IP instead of allocating a fresh one.
- The eager pass re-reads each record before acting, so a VM destroyed mid-startup isn't resurrected.
- gRPC graceful-shutdown stays registered last (closes first on SIGTERM); resume/create latency is unchanged in steady state.

Rewrites startup and request resolution — validate on staging (restart + chaos, especially SIGTERM mid-reattach) before enabling in prod.

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -1730,13 +1730,22 @@ func (m *Manager) ReserveStartupSlots() {
 		m.log.Error().Err(err).Msg("failed to read state for startup slot reservation")
 		return
 	}
-	nss := make([]string, 0, len(recs))
+	// Paused VMs keep their slot across a host reboot (which wipes every netns)
+	// and rebuild the same slot on resume, so reserve them unconditionally.
+	// Running records only hold a slot while their process — and thus their
+	// netns — is live; the network layer drops any whose netns is already gone.
+	var resumable, liveOnly []string
 	for _, rec := range recs {
-		if rec.Namespace != "" {
-			nss = append(nss, rec.Namespace)
+		if rec.Namespace == "" {
+			continue
+		}
+		if rec.Status == StatusPaused {
+			resumable = append(resumable, rec.Namespace)
+		} else {
+			liveOnly = append(liveOnly, rec.Namespace)
 		}
 	}
-	m.netMgr.ReserveSlotsAbove(nss)
+	m.netMgr.ReserveSlotsAbove(resumable, liveOnly)
 }
 
 // ---------------------------------------------------------------------------
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -1649,10 +1649,9 @@ func (m *Manager) reattachRecord(ctx context.Context, rec VMRecord, cleanupStale
 		}
 	}
 
-	// Commit only if the record wasn't deleted (DestroyVM) during reattach —
-	// the gate is open, so a delete can race the background pass. Running VMs
-	// persist atomically (write-if-present); paused VMs don't re-persist, so just
-	// re-check existence. Either way, undo the in-memory reattach if it's gone.
+	// Commit only if a concurrent DestroyVM didn't delete the record during
+	// reattach (the gate is open, so deletes race the background pass); otherwise
+	// undo the in-memory reattach.
 	if rec.Status == StatusPaused {
 		if m.recordDeleted(rec.ID) {
 			m.undoReattach(rec.ID, inst.Namespace)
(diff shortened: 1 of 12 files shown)


awille-ai (commented):
Architecture/security review of the current head (`f56b357a09`): I don’t see a critical issue that should block this PR, and the main startup-serving shape looks directionally right. The gRPC gate, on-demand reattach, singleflight dedupe, `PutIfPresent`, and removal of the post-pool broad sweep address most of the scary slot-reuse races from earlier revisions.

One medium risk remains and is worth fixing or explicitly accepting before production rollout: `ReserveSlotsAbove` still treats a running record with a missing `/run/netns/ns-N` bind mount as relinquishable. Linux can keep the process network namespace alive after that bind mount is gone, so an active `firecracker@` unit can be marked relinquished, refused by `reattachRecord`, and still considered healthy by the reconciler because systemd is active. That leaves a live sandbox untracked by VMD/local resolver, hard to destroy through normal control-plane flows, and potentially still consuming CPU/memory/billing resources. It is not an obvious tenant breakout by itself, but it is a security/ops concern because a live VM can become unreachable to lifecycle controls while remaining active.

Suggested direction: before marking a running record relinquished solely on `nsExists == false`, also consult systemd/unit liveness and either reserve the slot when the unit is active, or stop/mark-stale the unit through the same bounded/retriable cleanup path used elsewhere. At minimum, add a regression test for “active unit + missing netns bind mount” so this edge does not quietly regress.

Scaling-wise, this PR improves restart availability by removing O(fleet size) reattach from the serving critical path, but the background pass still does O(number of BoltDB records) systemd/socket checks. That is acceptable if staged with host-size telemetry: restart time to ready, background reattach duration, stale cleanup count, lazy reattach latency/error count, and pool allocation failures/slot skips. I would also watch the new `Unavailable` gate response rate during deploys to make sure clients retry instead of surfacing restart blips.

Recommendation: non-blocking comment. Do not treat this as production-ready without staging restart/chaos coverage around active-unit/missing-netns, SIGTERM during reattach, and concurrent destroy/create/resume during startup.

internal/vm/manager.go
```diff
@@ -1570,62 +1617,115 @@ func (m *Manager) ReattachAll(ctx context.Context) (reattached, stale int) {
...
 
-		m.mu.Lock()
-		m.vms[rec.ID] = inst
+	inst := toInstance(rec)
+	m.mu.Lock()
+	if existing, ok := m.vms[rec.ID]; ok {
 		m.mu.Unlock()
+		return existing, true
+	}
+	m.vms[rec.ID] = inst
```
pavitrabhalla: reattachRecord publishes to m.vms (line 1660) before ReattachVM restores network state (1666), and the eager Phase A loop calls this function directly — bypassing reattachSF, which only dedupes lazyReattach's own callers. So a request's lazy reattach and the background pass's eager call for the same vmID aren't mutually exclusive. If the eager call's unit-dead/socket check (1608-1639) frees the slot back to the pool while a concurrent lazy call (cleanupStale=false, skips those checks entirely) is mid-flight building/publishing inst, the lazy call's ReattachVM can bind the old vmID's firewall/devices state onto whatever the pool has since handed that slot to. persistStateIfPresent/undoReattach only guard the BoltDB+m.vms layer, not netMgr's kernel-level writes. Is there a reason the eager loop doesn't route through reattachSF too, so both paths serialize per-vmID?

pavitrabhalla:
internal/proxy/resolver.go:81 configures a 2s client timeout (and 500ms cache TTL) for calling vmd's local /instances/{id} endpoint, tuned when LookupInstance was a pure map read. This PR makes LookupInstance fall through to lazyReattach on a miss, which does uncancellable netlink/nftables work (ReattachVM/nsExecGo/AttachFirewall take no ctx) — not actually bounded by the 5s timeout in manager.go:1709 either, since that's only wired through unitDefinitelyDead/stopUnit and lazyReattach always calls with cleanupStale=false. Under host load right after a restart, could this exceed 2s and turn a reattach that would've succeeded into a client-side proxy timeout instead? Worth bumping the resolver's timeout or backing off its retry during the startup window?

pavitrabhalla (approved):
non-blocking gap: there’s no test directly exercising reattachByID’s concurrent-caller guarantee (e.g., two goroutines racing on the same vmID with different cleanupStale values)

PR #174, opened 2026-07-03
Adding TF resources

awille-ai:
adding our infrastructure to TF

arch


infra/modules/proxy-lb/main.tf
```diff
@@ -0,0 +1,175 @@
...
+      google_certificate_manager_dns_authorization.dns_authorizations[key].id
+    ]
+  }
+}
+
+resource "google_certificate_manager_certificate_map" "this" {
+  count = var.certificate_map_name == null ? 0 : 1
+
+  project = var.project_id
+  name    = "projects/${var.project_id}/locations/global/certificateMaps/${var.certificate_map_name}"
```
pavitrabhalla, line 110: name is set to a full resource path here (and reused as-is for certificate_map_entry.map below). google_certificate_manager_certificate_map.name has to be a short identifier ([a-z0-9-]{1,63}), not projects/.../certificateMaps/.... Won't this fail apply with an invalid name error? Should this just be var.certificate_map_name?

infra/envs/production/us-central1/main.tf
```diff
@@ -0,0 +1,178 @@
...
+  tags          = []
+  labels = merge(local.common_labels, {
+    component               = "vmd"
+    sandbox_role            = "vmd"
+    "goog-ops-agent-policy" = "v2-template-1-7-0"
+  })
+
+  service_account_email = local.api_service_account_email
+  boot_disk_image       = "projects/ubuntu-os-cloud/global/images/family/ubuntu-2404-lts"
+  can_ip_forward        = false
```
pavitrabhalla, line 167: can_ip_forward is false here but true on staging's sandbox_host. VMD's Firecracker guests send traffic over TAP interfaces with an IP that differs from the host's GCE IP — GCP's SDN drops that at the hypervisor when can_ip_forward is false, regardless of the host's own ip_forward sysctl. Is this intentionally different from staging, or should it be true?

.github/workflows/terraform-rollout-production.yml
```diff
@@ -0,0 +1,123 @@
...
+        run: |
+          set -euo pipefail
+          scripts/smoke-test-region.sh production us-west2
+          echo "production/us-west2 apply and smoke validation passed." >> "$GITHUB_STEP_SUMMARY"
+
+  central1:
+    name: Apply production us-central1
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main' && (inputs.target == 'central1' || inputs.target == 'all')
+    needs: [west2]
```
pavitrabhalla, line 76: When target=central1, west2's if evaluates false and it's skipped — and needs: [west2] here means central1 gets skipped too (GitHub Actions skips a job if a needed job is skipped, unless the dependent if checks always() or needs.west2.result). So a target=central1 dispatch does nothing. Should this be needs.west2.result == 'success' || needs.west2.result == 'skipped', or should needs: [west2] only apply when target == 'all'?

.github/workflows/terraform-cd.yml
```diff
@@ -0,0 +1,430 @@
...
+      - name: Verify latest revision serves traffic
+        run: |
+          SVC=${{ vars.CLOUD_RUN_SERVICE_USW }}
+          ARGS="--region ${{ vars.GCP_REGION_USW }} --project ${{ vars.GCP_PROJECT }}"
+          DESC=$(gcloud run services describe $SVC $ARGS --format=json)
+          LATEST=$(echo "$DESC" | jq -r '.status.latestReadyRevisionName')
+          PCT=$(echo "$DESC" | jq --arg r "$LATEST" '[.status.traffic[]? | select(.revisionName == $r or .latestRevision == true) | .percent] | add // 0')
+          if [ "$PCT" != "100" ]; then
+            echo "::error::$SVC: latest revision $LATEST serving ${PCT}% (want 100); deploy is not live."
+            exit 1
```
pavitrabhalla, line 306: staging-api and central1-api both end this verify step with curl -sf .../health after the traffic-percent check, but west2-api doesn't. So west2 only confirms 100% traffic is routed to the latest revision, not that the revision actually serves 200s — a crash-looping revision would pass. Intentional, or missing the health curl?

infra/modules/proxy-lb/main.tf
pavitrabhalla: Checked further — proxy_lb isn't wired into any env in this PR. production/us-central1, production/us-west2, and staging/us-central1 main.tf only call network/iam/api/sandbox_host/observability (staging also calls artifact_storage); none call proxy_lb. There's also no infra/envs/new-region despite the PLAN_REVIEW.md files referencing one, so that doc looks like stale/copy-pasted content. Since the module isn't invoked anywhere, this won't fire on merge — terraform apply stays a no-op for this PR. Still worth fixing before proxy_lb gets wired up to import the real LB, but not blocking here.

infra/envs/production/us-central1/main.tf
pavitrabhalla: Retracting this — if the intent here is to mirror the real current instance config 1:1 (so apply is a no-op), then can_ip_forward=false presumably matches what production's host is actually running today, not a mistake introduced by this PR. VMD's TAP traffic is likely NAT'd through the host's own IP before it leaves the NIC, so GCP's can_ip_forward check wouldn't be in the path either way. Only flag if it turns out the real instance is actually true and this diff has it backwards.

pavitrabhalla:
Re-verified the earlier review pass against the current head (\`f2a28f47cd\`) and resolved those threads. Two findings are still live and worth tracking as follow-ups rather than blockers — neither affects the default \`target=all\` rollout path:

1. **\`terraform-rollout-production.yml:76\`** — \`central1\` job has \`needs: [west2]\`. On a \`target=central1\`-only dispatch, \`west2\`'s \`if\` evaluates false and it's skipped, which causes \`central1\` to be skipped too (GitHub Actions skips a job whose dependency was itself skipped, absent an explicit check). A central1-only rollout silently no-ops today.
2. **\`terraform-cd.yml\` west2-api job (~line 297-309)** — \`staging-api\` and \`central1-api\` both end their revision-verify step with \`curl -sf .../health\` before the longer smoke test; \`west2-api\` is missing that fast fail-fast check. Not silently broken (the traffic-percent gate and \`smoke-test-region.sh\` still catch a bad deploy), just a slower failure signal for that region.

The other two flagged items (proxy-lb cert-map name, \`can_ip_forward\` mismatch) were checked and don't apply — proxy-lb isn't wired into any env in this PR, and \`can_ip_forward=false\` matches production's actual running config.

Approving — recommend picking up the two items above in a fast follow-up before west2 gets its own >1-host or selective-rollout usage.

pavitrabhalla (approved):
Approving — see consolidated follow-up comment for two non-blocking items to pick up separately.

pavitrabhalla:
Before the first production apply, worth checking the plan output against the resources hand-built during cell bring-up (2026-07-02→07) — anything the plan wants to create-anew (duplication) or that overlapping management would revert:

**On the usw2 Cloud Run service (if TF manages its env/IAM):** SANDBOX_ID_REGION=usw, the allUsers invoker binding, min-instances=2, DATABASE_URL/INTERNAL_API_TOKEN/SANDBOX_ACCESS_TOKEN_SEED as Secret Manager refs.

**Unmanaged by this PR (decide: import, add, or deliberately leave out of TF):**
- Global LB set: address dp-ip (136.68.212.233), health check dp-hc, backend services dp-use/dp-usw/dp-api-usw, serverless NEG dp-api-usw-neg, URL map sandbox-dataplane, cert-manager certs/authorizations + map dp-certs, target proxy dp-https, forwarding rule dp-fr
- Firewall: superserve-deny-central-to-usw2 (mirror of the usw2→central deny)
- Monitoring: log metric controlplane_db_deadline_errors, alert policy 'Control plane: sustained DB deadline errors', email notification channel
- z3 instance label component=vmd (deploy-workflow discovery depends on it)
- Legacy LBs sandbox-proxy-* are slated for retirement post-cutover (SS-159) — probably don't import those.

PR #175, opened 2026-07-03
api: server-side pagination, sorting & filtering for sandbox and template lists

mohamedsobhi777:
## What

Adds opt-in pagination, sorting, and filtering to `GET /sandboxes` and `GET /templates` so the console (and any client) can fetch one page at a time instead of the whole list.

New query params: `limit`, `offset`, `sort`, `order`, plus `status` (sandboxes) / `owner` (templates) and `q` (case-insensitive name substring). Total row count is returned in the **`X-Total-Count`** response header.

## How

- **SQL** (`db/queries/*.sql`, regenerated with sqlc 1.30.0): new paged + count queries. Dynamic sort is expressed with **guarded `CASE` ORDER BY terms** (one active per query, `created_at DESC` tiebreaker) so it stays type-safe in sqlc. `LIMIT NULL` returns all rows.
- **Handlers**: shared `parsePageParams` (whitelisted sort columns, `limit` clamped to 200, `400` on bad input; `owner` validated). The COUNT query is skipped on the unpaginated path (the returned slice length is the total).
- **OpenAPI**: documented the new params + `X-Total-Count` header; `TestOpenAPISpecMatchesRoutes` passes.

## Backward compatibility

The response body stays a **bare array**. Omitting `limit` returns the full list exactly as before, so the TS SDK, Python SDK, and MCP server are unaffected. The templates `name_prefix` (prefix) filter is kept alongside the new `q` (substring) for existing callers.

## Security

- **Tenancy**: every new query is team-scoped (`team_id = @team_id`, `destroyed_at`/`deleted_at IS NULL`); the count query uses the identical `WHERE`, so `X-Total-Count` can't leak cross-tenant totals. The template `owner` filter can only match the caller's team or the shared system team.
- **Injection**: all filters are bound params; `sort`/`order`/`owner` are compared to string literals inside `CASE`, never concatenated. `limit`/`offset` parsed to ints.
- **Resource bounds**: page size clamped to 200.

## Test plan

- [x] `go build ./internal/api/... ./internal/db/...` + `go vet` — clean
- [x] `sqlc generate` is idempotent (CI `sqlc-check` will pass)
- [x] `go test ./internal/api/` — unit tests pass (incl. new `pagination_test.go`)
- [x] `internal/integration/pagination_test.go` compiles; runs in CI against Postgres (verifies real sort/filter/pagination + `X-Total-Count` end-to-end)

## Paired change

Console consumer: superserve-ai/superserve#262 (backward-compatible, so merge order isn't strict).

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -389,9 +389,11 @@ paths:
         - name: status
           in: query
           required: false
-          description: Filter by exact sandbox status, e.g. `active` or `paused`.
+          description: Filter by exact sandbox status. Unknown values are a `400`.
           schema:
             type: string
+            enum:
+              [starting, active, pausing, paused, resuming, failed, deleted]
         - name: sort
           in: query
           required: false
@@ -1357,6 +1359,8 @@ paths:
                 type: array
                 items:
                   $ref: "#/components/schemas/TemplateResponse"
+        "400":
+          $ref: "#/components/responses/BadRequest"
         "401":
           $ref: "#/components/responses/Unauthorized"
         "500":
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -355,8 +355,16 @@ paths:
       tags: [Sandboxes]
       summary: List all sandboxes
       description: |
-        Returns sandboxes belonging to the authenticated team, optionally
-        filtered by metadata tags.
+        Returns sandboxes belonging to the authenticated team, ordered by
+        creation time (newest first) by default.
+
+        ## Pagination, sorting, and search
+
+        Pass `limit` (and `offset`) to fetch one page at a time; the
+        `X-Total-Count` response header reports the total across all pages.
+        Omitting `limit` returns the full list, so existing unpaginated
+        callers are unaffected. Sort with `sort` + `order`, filter by exact
+        `status`, and search names with `q` (case-insensitive substring).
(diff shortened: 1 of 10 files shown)


pavitrabhalla:
Pushed 1d28b9a with fixes from review:

- `X-Total-Count` was wrong for offset-without-limit requests (counted only the returned rows). Total resolution is now a shared `resolveTotal` helper, which also skips the COUNT round trip whenever the page already proves the total (short page, or empty page at offset 0).
- `q` is LIKE-escaped (`%`, `_`, `\`) so it matches literally per the documented substring semantics. `name_prefix` is left unescaped since that behavior predates this PR.
- Templates `sort=built_at|size&order=desc` now pin `NULLS LAST` — Postgres defaults DESC to NULLS FIRST, which put never-built templates on top.
- Unknown `?status=` on /sandboxes is now a 400 naming the valid set (mirroring `owner`) instead of silently matching zero rows; the spec enums the values.
- GET /templates documents the 400 it can return.
- `optStr` dropped for the pre-existing identical `nullableStr`.

Left the limit>200 clamp as-is since the param description documents it. Extended the integration tests to cover each fix; full suite passes locally against Postgres 16.

pavitrabhalla (approved):

PR #176, opened 2026-07-06
Add OpenTelemetry metrics config

awille-ai:
Adds Phase 1 OpenTelemetry metrics for the sandbox control plane. The branch wires a bounded telemetry recorder through config/control-plane startup, emits sandbox lifecycle, VMD RPC, DB pool, and host capacity metrics, and includes local OTEL Collector debug config plus docs for validating metrics before GMP ingestion.

Telemetry is disabled by default, avoids high-cardinality/customer identifiers, and keeps labels limited to operational fields like operation, result, region, host_id, method, service.name, and environment.

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -615,7 +615,7 @@ func (h *Handlers) resumePausedSandbox(c *gin.Context, sandbox *db.Sandbox, team
 			// VM unreachable — nothing to preserve; mark failed rather than
 			// wedge in 'resuming' (the CTE also closes any open interval).
 			log.Error().Err(perr).Str("sandbox_id", sandboxID.String()).Msg("resume revert: PauseInstance failed, marking sandbox failed")
-			h.markSandboxFailedAsync(revertCtx, sandboxID, teamID)
+			h.markSandboxFailedAsync(revertCtx, sandboxID, teamID, sandbox.HostID)
 			return
 		}
 		// Fresh deadline so a slow snapshot above can't starve the DB write.
@@ -879,6 +879,7 @@ func (h *Handlers) ResumeSandbox(c *gin.Context) {
 		respondError(c, ErrInternal)
 		return
 	}
+	SetTelemetryHostID(c, sandbox.HostID)
 
 	if sandbox.Status != db.SandboxStatusPaused {
 		respondError(c, ErrInvalidState)
@@ -936,6 +937,7 @@ func (h *Handlers) DeleteSandbox(c *gin.Context) {
 		return
 	}
 
+	SetTelemetryHostID(c, sandbox.HostID)
 	// Claim the sandbox for deletion atomically. The guarded soft-delete fires
 	// from a quiescent state (active/paused/failed) or a transitional state whose
 	// worker is provably gone (stale), so it serializes against a live resume/pause
@@ -1613,6 +1615,7 @@ func (h *Handlers) CreateSandbox(c *gin.Context) {
 	} else {
 		hostID = "default"
 	}
+	SetTelemetryHostID(c, hostID)
 
 	// Resolve the VMD client up front so we don't waste a DB INSERT on
 	// a host we can't reach.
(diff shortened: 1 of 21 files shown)


internal/telemetry/dbpool.go
```diff
@@ -0,0 +1,76 @@
...
+				AcquireDurationSecondsDelta: nonNegativeDurationDelta(acquireDuration, prevAcquireDuration).Seconds(),
+			})
+
+			prevAcquire = acquire
+			prevEmptyAcquire = emptyAcquire
+			prevCanceledAcquire = canceledAcquire
+			prevAcquireDuration = acquireDuration
+		}
+
+		sample()
```
pavitrabhalla, line 54: First `sample()` call here runs before any `prev*` baseline is set, so it reports the pool's full cumulative `AcquireCount`/`EmptyAcquireCount`/`CanceledAcquireCount`/`AcquireDuration` as the first delta instead of 0. If the pool's been running a while before metrics get enabled, that first export spikes `db_pool_acquire_total` by however many acquires happened since pool creation. Should this seed `prevAcquire` etc. on the first call without recording, or is the one-time startup spike acceptable?

internal/telemetry/otel.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	attrs := r.attrs(
+		attribute.String("operation", safeOperation(t.Operation)),
+		attribute.String("result", safeResult(t.Result)),
+		attribute.String("region", safeRegion(t.Region)),
+		attribute.String("host_id", safeHostID(t.HostID)),
+	)
+	r.sandboxTransitions.Add(ctx, 1, metric.WithAttributes(attrs...))
+	if t.Duration > 0 {
+		r.sandboxDuration.Record(ctx, t.Duration.Seconds(), metric.WithAttributes(attrs...))
```
pavitrabhalla: `RecordSandboxTransition` and `RecordVMDCall` (line 180) each call `metric.WithAttributes(attrs...)` twice — once for the counter, once for the histogram — while `RecordHostCapacity`/`RecordDBPoolStats` below precompute `opt := metric.WithAttributes(attrs...)` once and reuse it. Was the double call here intentional, or should these two match the `opt` pattern used elsewhere in the file?

internal/telemetry/host_capacity.go
```diff
@@ -0,0 +1,80 @@
...
+	if pool == nil || recorder == nil {
+		return
+	}
+	if interval <= 0 {
+		interval = 15 * time.Second
+	}
+
+	go func() {
+		sample := func() {
+			timeout := interval
```
pavitrabhalla: `timeout` is capped at `min(interval, 5s)`. At the default 15s interval that's a comfortable 5s budget, but if `OTEL_EXPORT_INTERVAL` is set below 5s the query timeout shrinks to match the interval with no slack at all. Is that intentional, or should the cap be independent of interval?

internal/api/handlers.go
```diff
@@ -1807,12 +1811,17 @@ func (h *Handlers) CreateSandbox(c *gin.Context) {
 		log.Error().Err(vmdErr).Str("sandbox_id", sandbox.ID.String()).Msg("VMD create/resume failed")
```
pavitrabhalla, line 1812: This path calls `RecordSandboxTransition(..., "fail", ...)` directly, then `respondError` returns a non-2xx status which `SandboxLifecycleTelemetry` (router.go:33) also picks up and records as `operation="create" result=error` for the same request. So a single failed create emits two transition records under different operation labels. Is the double emission intentional — e.g. `fail` and `create` meant to be queried separately — or should the explicit calls here be dropped in favor of the middleware capturing it via HTTP status?

pavitrabhalla (approved):
Approving. Left a few non-blocking questions inline, summarized here:

- `internal/telemetry/dbpool.go:50` — first `sample()` call runs before `prev*` counters are seeded, so the first delta reported is the pool's full cumulative count rather than 0. One-time startup spike, not a functional bug.
- `internal/telemetry/otel.go:164` — `RecordSandboxTransition`/`RecordVMDCall` call `metric.WithAttributes` twice (counter + histogram) instead of precomputing `opt` once like `RecordHostCapacity`/`RecordDBPoolStats` do.
- `internal/telemetry/host_capacity.go:41` — query timeout is `min(interval, 5s)`, so intervals under 5s leave no slack for the query to complete.
- `internal/api/handlers.go:1811` — a failed create emits both a `fail` transition (from the handler) and a `create`/error transition (from the `SandboxLifecycleTelemetry` middleware) for the same request — worth confirming that double emission is intended for how the dashboards will query this.

PR #177, opened 2026-07-07
cd: deploy VMD and proxy to the usw cell host after primary prod

pavitrabhalla:
Extends Deploy VMD and Deploy Proxy to deploy the us-west2 cell host (`superserve-vmd-usw2`) after the primary production deploy succeeds, closing the hand-deploy drift gap from — the host's binaries were hand-staged at bring-up and the proxy has already drifted once past a merged change. Same pattern as the Deploy API usw cell steps (#170).

**Technical decisions**
- **Repo-level var gate, step not job.** The usw deploy is a final step in each `deploy-production` job, gated on `vars.CLOUD_RUN_SERVICE_USW != ''` — the existing repo-level cell-exists flag from #170. Repo-level because environment-scoped variables don't resolve in job-level `if:` (the bug review caught on #170); a step also avoids a second production-environment job doubling any approval gate, matching Deploy API / CD Migrate / Seed Templates.
- **Build once, deploy per region.** The usw step reuses `bin/*` built earlier in the same job and just re-runs the deploy script with usw parameters — no rebuild.
- **Region-scoped label discovery.** The deploy scripts gain an optional `GCP_REGION` env: prod and the usw cell share a GCP project, so once the cell host carries `component=vmd` an unscoped `VMD_LABEL` filter would fold it into the primary parallel fan-out. Primary steps pass `vars.GCP_REGION`, the usw step passes the existing repo-level `vars.GCP_REGION_USW` (us-west2). Scoping matches the zone basename in Python instead of a gcloud filter expression to avoid URI-match ambiguity. Zero matches stays a hard failure — a silent skip is exactly the drift this closes. No new variables introduced.
- **Proxy env reuses prod values.** The usw proxy step passes the `_PROD` secrets/vars (`PROXY_DOMAIN_PROD`, `SANDBOX_ACCESS_TOKEN_SEED_PROD`, `PROXY_ALLOWED_ORIGINS_PROD`, `REQUIRE_DATA_PLANE_PROD`); the script rewrites `/etc/sandbox/proxy.env` on the host. **Reviewer, please confirm** the cell's hand-staged proxy.env matches these prod values (domain suffix + token seed shared across cells). If the cell diverges, we need `_USW` variants before merge, not after.
- `seed-templates.yml`'s "Wait for Deploy VMD" keys off the workflow file + commit and overall run conclusion — job names are unchanged, and the gate now correctly also waits for the cell deploy.

**Deploy prerequisite (must be confirmed before merge)**
`superserve-vmd-usw2` (us-west2-a, rayai-prod) carries the `vmd-usw2` network *tag*, but discovery filters on *labels*. It likely needs `component=vmd` added:
`gcloud compute instances add-labels superserve-vmd-usw2 --zone=us-west2-a --project=rayai-prod --labels=component=vmd`
Without it, the usw step fails loudly on next push to a matching path (intentional fail-visible behavior, but it will block deploys until labeled). Also confirms needed: the host's `VMD_SERVICE`/`VMD_INSTALL_DIR` layout and `/etc/sandbox/*.env` paths match the primary hosts, since the scripts assume them.

**Testing**
- `actionlint` clean on both workflows (v1.7.12 via `go run`).
- Both YAML files parse; verified step ordering (usw step last in `deploy-production`) and `py_compile` on both scripts.
- Unit-checked the region filter against basename and full-URL zone values (matches `us-west2-a`, rejects `us-central1-a` and `us-west21-a`).
- Not verifiable from here: a live run against GCP (label presence on the z3, IAP SSH to it, whether staging has any host outside `vars.GCP_REGION` that the new scoping would drop — believed none). Suggest a manual `workflow_dispatch` with environment=production right after merge + labeling, and watching both usw steps.

### DIFF
--- a/.github/workflows/deploy-proxy.yml
+++ b/.github/workflows/deploy-proxy.yml
@@ -1,7 +1,8 @@
 # Deploy Proxy — edge proxy on the VMD hosts.
 #
 # Production NEVER deploys without staging first. Same pattern as Deploy API /
-# Deploy VMD.
+# Deploy VMD. The us-west2 cell host deploys as a final production step,
+# after the primary region succeeds — same shape as Deploy API's usw steps.
 
 name: Deploy Proxy
 
@@ -54,9 +55,13 @@ jobs:
       - name: Set up Cloud SDK
         uses: google-github-actions/setup-gcloud@v2
 
+      # GCP_REGION scopes label discovery to the primary region so the
+      # usw cell host (same project, same label) isn't folded into this
+      # fan-out — the cell deploys sequentially after production below.
       - name: Discover and deploy proxy to VMD instances
         env:
           GCP_PROJECT: ${{ vars.GCP_PROJECT }}
+          GCP_REGION: ${{ vars.GCP_REGION }}
           VMD_LABEL: ${{ vars.VMD_LABEL }}
           VMD_INSTALL_DIR: ${{ vars.VMD_INSTALL_DIR }}
           SHA: ${{ github.sha }}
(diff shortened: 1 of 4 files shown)


awille-ai (commented):
Architecture review focused on scaling/performance, security, edge cases, and compatibility with the multi-region user-management work.

No critical/high issues found, so I am not blocking this PR. The shape is consistent with the existing Deploy API usw rollout: primary deploy/verify first, then the cell step inside the same production environment gate. The region filter also avoids the important blast-radius issue where the usw VMD host gets pulled into the primary-region fan-out once it shares the `component=vmd` label.

A few notes before/around merge:

- **Operational edge, non-blocking:** merging this PR alone will not automatically exercise these changes, because the `push.paths` filters do not include `.github/workflows/deploy-vmd.yml`, `.github/workflows/deploy-proxy.yml`, or `.github/workflows/scripts/deploy-*.py`. The PR body calls out a manual `workflow_dispatch` after merge; I would either keep that as an explicit merge checklist item or add the workflow/script paths to the triggers so deployment workflow changes self-validate on merge. Otherwise the hand-deploy drift this is meant to close can persist silently until the next code change that happens to touch a deploy-triggered path.

- **User-management/token compatibility:** the updated proxy cell step using `PROXY_DOMAIN_USW` plus the Secret Manager `sandbox-access-token-seed` is the right direction for the region-tagged user/API-key model. The critical invariant is that the usw control plane and usw proxy read the exact same HMAC seed; if either side is pinned to a different secret version/value, `/files`, `/exec`, and terminal access will all fail with `invalid access token` for otherwise-valid sandboxes. I would verify the Cloud Run cell service is also sourced from that same Secret Manager value/version before the first run.

- **Scaling/perf:** no major scaling regression. The scripts still list by the VMD label first and then filter by region locally, so the project-wide instance list stays bounded by labeled VMD hosts rather than all compute instances. Sequential primary-then-usw deploys increase production deploy wall time, but the reduced blast radius is a reasonable tradeoff for this class of VM rollout.

- **Security:** I do not see a new direct exposure path in the proxy or VMD code. The injected proxy env values are still validated before being written into the remote shell script, and the usw seed is not printed by the deploy script. One operational security consideration: fetching the seed in Actions means the production deploy identity needs Secret Manager access to that secret, so keep that grant scoped to the one secret rather than broad project secret access.

Recommendation: proceed, with the manual post-merge deploy/verification called out explicitly and the usw control-plane/proxy seed match confirmed before relying on this for production traffic.

meAmitPatil (approved):

PR #178, opened 2026-07-07
api: auto-delete paused sandboxes via auto_delete_seconds; patchable timeout_seconds

meAmitPatil:
Opt-in garbage collection for paused sandboxes, plus a patchable auto-pause timeout.

- **`auto_delete_seconds`** (create + PATCH): deletes a sandbox once it has been continuously paused that long. The deadline (`auto_delete_at`) arms on every pause, is cleared by resume, and — when set on an already-paused sandbox — counts from the request, never retroactively. `0` = delete on pause; unset = keep forever (default). Max 30 days.
- **`timeout_seconds` is now patchable**: PATCH can set or clear (`null`) the auto-pause timeout on an existing sandbox.
- Reaper sweep soft-deletes due rows in one guarded statement (claims from `paused` only, deadline re-checked under the row lock), then reuses the shared teardown path; a snapshot-row backstop sweep covers teardown abandoned by crash/shutdown.
- Migration is idempotent and a no-op for existing sandboxes.

Tests: `go test ./internal/api/...`. OpenAPI updated for create/PATCH/response.

### DIFF
--- a/db/queries/sandboxes.sql
+++ b/db/queries/sandboxes.sql
@@ -6,8 +6,8 @@
 -- template_id is optional (NULL when sandbox is not derived from a template).
 -- snapshot_path / mem_path / base_path / delta_path pin the sandbox to a
 -- specific build's artifacts so a later template rebuild can't corrupt it.
-INSERT INTO sandbox (id, team_id, name, status, vcpu_count, memory_mib, host_id, ip_address, pid, snapshot_id, timeout_seconds, metadata, template_id, snapshot_path, mem_path, base_path, delta_path)
-VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17)
+INSERT INTO sandbox (id, team_id, name, status, vcpu_count, memory_mib, host_id, ip_address, pid, snapshot_id, timeout_seconds, metadata, template_id, snapshot_path, mem_path, base_path, delta_path, auto_delete_seconds)
+VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18)
 RETURNING *;
 
 -- name: CreateSandboxFromTemplate :one
@@ -22,8 +22,8 @@ WITH tpl AS (
     AND (t.team_id = $14 OR t.team_id = $15)
   FOR KEY SHARE
 )
-INSERT INTO sandbox (id, team_id, name, status, vcpu_count, memory_mib, host_id, ip_address, pid, snapshot_id, timeout_seconds, metadata, template_id, snapshot_path, mem_path, base_path, delta_path, disk_mib)
-SELECT $1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, tpl_id, $16, $17, $18, $19, disk_mib FROM tpl
+INSERT INTO sandbox (id, team_id, name, status, vcpu_count, memory_mib, host_id, ip_address, pid, snapshot_id, timeout_seconds, metadata, template_id, snapshot_path, mem_path, base_path, delta_path, disk_mib, auto_delete_seconds)
+SELECT $1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, tpl_id, $16, $17, $18, $19, disk_mib, $20 FROM tpl
 RETURNING *;
 
 -- name: GetSandbox :one
@@ -134,11 +134,6 @@ FROM activated a
 WHERE feature_enabled('billing_metrics_write', a.team_id)
 ON CONFLICT (sandbox_id) WHERE ended_at IS NULL DO NOTHING;
 
--- name: SetSandboxSnapshot :exec
-UPDATE sandbox
-SET snapshot_id = $2, updated_at = now()
-WHERE id = $1 AND team_id = $3 AND destroyed_at IS NULL;
-
 -- name: DestroySandbox :one
 -- Atomic, guarded soft-delete. Claims the sandbox from a quiescent state
 -- (active/paused/failed) or from a transitional state (starting/resuming/pausing)
@@ -150,6 +145,9 @@ WHERE id = $1 AND team_id = $3 AND destroyed_at IS NULL;
 -- so a crash after the claim can't strand a deleted sandbox with a live JWT or an
 -- open interval. Returns the claimed id; 0 rows means already-deleted or a still-live
 -- transition — the caller distinguishes the two with a re-read.
+--
+-- The revocation + interval-close CTEs are mirrored in ClaimAutoDeleteSandboxes;
+-- keep the side effects of both in sync.
 WITH destroyed AS (
   UPDATE sandbox
   SET destroyed_at = now(), status = 'deleted', updated_at = now()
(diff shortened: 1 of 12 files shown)


awille-ai (commented):
Architecture/security review pass complete. I did not find any critical or high-severity issues, so I am not blocking the PR and I am not requesting changes.

What looks good:
- The new auto-delete path is tenant-scoped at the API update layer and the reaper only claims rows already in `paused`, which keeps the destructive path narrower than user-initiated delete.
- The pause/resume/delete races are handled thoughtfully: `BeginResume` clears `auto_delete_at`, `FinalizePause` re-arms only from pause-eligible states, and the auto-delete claim uses a guarded query with `FOR UPDATE SKIP LOCKED`.
- The auto-delete path mirrors the important delete side effects: revocation row plus active/compute/storage interval closure, which is important for token safety and billing correctness.
- `timeout_seconds` and `auto_delete_seconds` validation is centralized in the handler, and the PATCH tri-state handling covers absent/null/value correctly.
- CI is green on the PR head.

Non-blocking concerns to consider:
- Scaling/backlog: `reapAutoDeleteOnce` claims only one `BatchSize` batch per tick and then waits for host teardown before the next claim. Because teardown can wait on VMD and snapshot cleanup, a large wave of `auto_delete_seconds=0` pauses could leave many due sandboxes undeleted for multiple ticks/minutes even though the DB claim itself is cheap. That is probably acceptable for launch, but if this becomes user-facing cleanup/SLO-sensitive, consider decoupling the DB soft-delete claim from host teardown or draining multiple DB batches per tick while bounding teardown concurrency.
- Defense in depth: the migration only constrains `auto_delete_seconds >= 0`; the public API enforces the 30-day max. If any internal job, SQL console repair, or future bulk path bypasses the handler, it can create unbounded retention windows and defeat the cost-control intent. I would consider mirroring the upper bound in the DB constraint (`<= 2592000`) unless there is a deliberate reason to keep the database looser than the API.
- Generated comments/docs drift: `internal/db/models.go` still describes `timeout_seconds` as a hard lifetime cap from `created_at`, while the PR changes it to active-session auto-pause. This is not runtime-impacting, but it is a trap for future maintainers because the generated model comments disagree with the query/OpenAPI semantics.

Recommendation: proceed with merge after considering the two small hardening tweaks above. I would not block on them unless the team expects very high-volume immediate auto-delete usage on day one.

db/queries/snapshots.sql
```diff
@@ -30,3 +30,16 @@ ORDER BY created_at DESC;
...
 DELETE FROM snapshot
 WHERE id = $1;
+
+-- name: DeleteSnapshotsOfDestroyedSandboxes :execrows
+-- Backstop for destroy teardown that never ran (crash or shutdown between a
+-- sandbox's guarded soft-delete and cleanupSandboxSnapshots): drop snapshot
+-- rows for long-destroyed sandboxes. The grace period keeps this clear of
+-- in-flight teardown; the orphaned on-disk files then fall to the vm
+-- reconciler's disk scan.
+DELETE FROM snapshot
```
pavitrabhalla: This filters purely on `destroyed_at` with no `host_id` predicate. The only index touching that column, `idx_sandbox_host_destroyed`, is compound on `(host_id, destroyed_at)` and won't serve this — so every 30s reaper tick does a sequential scan over all soft-deleted rows, growing over time since deletes are soft. Worth adding a plain `destroyed_at`-only partial index in the new migration alongside `idx_sandbox_auto_delete_due`?

internal/api/reaper.go
```diff
@@ -74,6 +74,10 @@ func (h *Handlers) reaperLoop(ctx context.Context, cfg ReaperConfig) {
 		sentrylog.RunSafe("reaper", func() { h.reapOnce(ctx, cfg.BatchSize, parallelism) })
 		sentrylog.RunSafe("interval-sweep", func() { h.sweepOrphanedIntervals(ctx) })
 		sentrylog.RunSafe("revocation-cleanup", func() { h.cleanupExpiredRevocations(ctx) })
+		sentrylog.RunSafe("snapshot-row-sweep", func() { h.sweepOrphanedSnapshotRows(ctx) })
+		// Last: this is the only sweep that waits on host RPCs, so it must not
```
pavitrabhalla: "only sweep that waits on host RPCs" — `reapOnce` above (dispatched first via `pauseExpired`) calls `vmd.PauseInstance`, so that's not accurate. Worth fixing so a future sweep doesn't get placed after `reapOnce` on the assumption it's DB-only?

internal/api/reaper.go
```diff
@@ -174,6 +178,93 @@ dispatch:
...
+		return
+	}
+
+	if len(due) == 0 {
+		return
+	}
+
+	log.Info().Int("count", len(due)).Msg("reaper: deleting expired paused sandboxes")
+
+	sem := make(chan struct{}, parallelism)
```
pavitrabhalla: This sem/waitgroup/dispatch loop is identical to `reapOnce`'s above it, differing only in the claim query and per-item handler. Worth pulling into a shared `dispatchBounded(ctx, items, parallelism, fn)` so a future fix to the cancellation/semaphore logic doesn't need to land twice?

internal/api/handlers.go
```diff
@@ -2382,9 +2445,81 @@ func (h *Handlers) PatchSandbox(c *gin.Context) {
 		h.logSandboxActivity(c.Request.Context(), sandbox.ID, teamID, actorIDFromContext(c), "sandbox", "metadata_updated", "success", &sandbox.Name, nil, nil)
 	}
 
+	if body.AutoDeleteSeconds.Set {
```
pavitrabhalla, line 2447: This block and the `TimeoutSeconds.Set` block below it are structurally identical (DB call → error/log → 0 rows → 404 → activity log), differing only in method, params, and the log string. The comment above `patchSandboxRequest` already anticipates more patchable fields landing the same way — worth factoring into a small helper now before a third copy shows up?

db/queries/sandboxes.sql
```diff
@@ -425,3 +445,89 @@ closed_billing_compute AS (
 SELECT p.id, p.team_id, p.name, p.snapshot_id, p.host_id
 FROM paused p
 LEFT JOIN closed_intervals ci ON ci.sandbox_id = p.id;
+
+-- name: UpdateSandboxAutoDelete :execrows
```
pavitrabhalla, line 444: `auto_delete_at` is now armed/cleared by hand in six separate statements (here, `FinalizePause`, `RevertResumeToPaused`, `BeginResume`, and both `MarkSandboxFailed*`). We already use `BEFORE UPDATE` triggers for cross-cutting row invariants (quota counters). Any reason not to centralize the arm/clear logic there instead, so a future write path to `sandbox` can't forget to touch it?

pavitrabhalla (approved):
Nothing blocking — status guards, locking, and the teardown refactor all hold up. Left a few non-blocking notes below.

awille-ai (approved):

The lifecycle/security shape looks sound: tenant-scoped updates, guarded paused-only auto-delete claims, `FOR UPDATE SKIP LOCKED` for reaper concurrency, revocation + active/compute/storage interval closure in the destructive path, and DB-level bounds on `auto_delete_seconds`. CI is green on the current head.

Non-blocking cleanup notes:
- `internal/api/reaper.go` still has top-level comments describing `timeout_seconds` as a hard cap measured from `created_at`; the implementation now evaluates the current active session and PATCH can change the setting.
- `internal/api/handlers.go` has the same stale field comment on `createSandboxRequest.TimeoutSeconds`, while OpenAPI/generated model comments were updated. Worth syncing these comments so future changes do not accidentally reintroduce creation-time lifetime semantics.

awille-ai (approved):

The lifecycle/security shape looks sound: tenant-scoped updates, guarded paused-only auto-delete claims, `FOR UPDATE SKIP LOCKED` for reaper concurrency, revocation + active/compute/storage interval closure in the destructive path, and DB-level bounds on `auto_delete_seconds`. CI is green on the current head.

Non-blocking cleanup notes:
- `internal/api/reaper.go` still has top-level comments describing `timeout_seconds` as a hard cap measured from `created_at`; the implementation now evaluates the current active session and PATCH can change the setting.
- `internal/api/handlers.go` has the same stale field comment on `createSandboxRequest.TimeoutSeconds`, while OpenAPI/generated model comments were updated. Worth syncing these comments so future changes do not accidentally reintroduce creation-time lifetime semantics.

PR #179, opened 2026-07-07
cd: seed the usw cell's templates after primary prod

pavitrabhalla:
Extends `seed-templates.yml` to seed the usw cell after the primary prod seed, following the same step-in-job pattern as the CD Migrate and Deploy API cell steps (#169, #170). Skipped until `CLOUD_RUN_SERVICE_USW` is set; one production environment approval gates both seeds.

Also raises the prod job timeout from 45 → 90 minutes to cover two blocking seeder invocations (primary + usw cell), each with the seeder's default 30-minute build wait.

Part of the usw cell rollout (#177).


pavitrabhalla:
Duplicate — this work already merged as #173.

PR #185, opened 2026-07-07
Aw/telemetry phase2

awille-ai:
Staging collectors export to GMP.
Grafana or Cloud Monitoring can query the Phase 1 metrics.
Dashboards show real data.
Collector failures are visible.
Metrics remain low-cardinality.
Production rollout plan exists but is not necessarily enabled yet.

### DIFF
--- a/.github/workflows/terraform-cd.yml
+++ b/.github/workflows/terraform-cd.yml
@@ -5,6 +5,7 @@ on:
     branches: [main]
     paths:
       - 'infra/**'
+      - '.github/workflows/terraform-cd.yml'
 
 jobs:
   wait-for-migrations:
@@ -81,7 +82,7 @@ jobs:
           path: /tmp/controlplane-image.tar.gz
           if-no-files-found: error
 
-  staging-infra:
+  staging-us-central1-infra:
     name: Apply staging/us-central1
     runs-on: ubuntu-latest
     needs: [wait-for-migrations]
@@ -103,8 +104,11 @@ jobs:
           workload_identity_provider: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}
           service_account: ${{ secrets.GCP_SERVICE_ACCOUNT }}
 
-      - name: Install validation dependencies
-        run: sudo apt-get update && sudo apt-get install -y jq
+      - name: Verify validation dependencies
+        run: |
+          set -euo pipefail
+          command -v jq
+          command -v curl
 
       - name: Terraform apply staging/us-central1
         run: |
@@ -116,11 +120,11 @@ jobs:
           terraform plan -input=false -out=tfplan
           terraform apply -input=false -auto-approve tfplan
 
-  staging-api:
+  staging-us-central1-api:
     name: Deploy API staging/us-central1
     runs-on: ubuntu-latest
     environment: staging
-    needs: [build-api-image, staging-infra]
+    needs: [build-api-image, staging-us-central1-infra]
     permissions:
       contents: read
       id-token: write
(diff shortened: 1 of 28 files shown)


.github/workflows/terraform-cd.yml
```diff
@@ -3,8 +3,6 @@ name: Terraform CD
 on:
   push:
     branches: [main]
-    paths:
-      - 'infra/**'
```
pavitrabhalla, line 7: Dropped the `paths: infra/**` filter — this now runs prod Terraform apply on every push to main, including app-only changes, and races with `deploy-api.yml` against the same Cloud Run service. Intentional, or should the path filter stay?

infra/modules/api/main.tf
```diff
@@ -92,7 +92,6 @@ resource "google_cloud_run_v2_service" "this" {
       client_version,
       scaling,
       template[0].containers[0].image,
-      template[0].containers[0].env,
```
pavitrabhalla, line 95: Removing `template[0].containers[0].env` from `ignore_changes` here, but production's `main.tf` doesn't declare `SENTRY_DSN` via Terraform the way staging's does. Won't the next `terraform apply` on prod strip `SENTRY_DSN` from the running service? Is prod's env config getting updated before this merges?

.github/workflows/terraform-cd.yml
```diff
@@ -178,253 +179,14 @@ jobs:
...
             exit 1
           fi
-          curl -sf "$(echo "$DESC" | jq -r '.status.url')/health"
-
-      - name: Smoke test production/us-central1
-        run: |
-          set -euo pipefail
-          scripts/smoke-test-region.sh production us-central1
-          echo "production/us-central1 API deploy and smoke validation passed." >> "$GITHUB_STEP_SUMMARY"
+          curl -sf "$STATUS_URL/health" >/dev/null
```
pavitrabhalla: This step used to fail if `PCT != 100` (traffic not fully cut over); now it only checks `status.url` is non-empty. A deploy live at 50% traffic would pass this check. Dropped on purpose, or should the percent check stay?

internal/api/vmd_errors.go
```diff
@@ -128,16 +134,24 @@ func (h *Handlers) markSandboxFailedAsync(reqCtx context.Context, sandboxID, tea
 // the VM is already running but unusable. vmd must be the client for the host the
 // VM booted on — destroying through the default client would leave the VM running
 // on a non-default host.
-func (h *Handlers) failSandboxAfterBoot(ctx context.Context, vmd VMDClient, sandboxID, teamID uuid.UUID, instanceID string) {
+func (h *Handlers) failSandboxAfterBoot(ctx context.Context, vmd VMDClient, sandboxID, teamID uuid.UUID, instanceID, hostID string, recordTransition bool) {
```
pavitrabhalla: `recordTransition` is `false` at all 3 call sites (2 in handlers.go, 1 in the test), so the `RecordSandboxTransition` calls below are currently dead code. Is there a caller planned that passes `true`, or should this param be dropped?

internal/api/reaper.go
```diff
@@ -238,6 +240,7 @@ func (h *Handlers) pauseExpired(ctx context.Context, sbx db.ClaimExpiredSandboxe
 	}
 
 	l.Info().Msg("reaper: sandbox paused due to timeout")
+	RecordSandboxTransition(ctx, "timeout_pause", telemetry.ResultSuccess, sbx.HostID, time.Since(started))
```
pavitrabhalla, line 324: `timeout_pause` is only recorded on the success path here — the `revertToActiveOrFail` branches above (VMD lookup/pause failure) don't emit a transition, so reaper pause failures won't show up on the new dashboard. Intentional, or worth recording there too?

pavitrabhalla (commented):
A few questions from review — nothing else stood out.

meAmitPatil (approved):
Approved. Before merging can you check if `deploy/otel/README.md`  is an internal file that got dragged.

PR #187, opened 2026-07-08
fix: cast auto_delete_seconds param so PATCH auto-delete stops 42P08-ing

meAmitPatil:
`PATCH auto_delete_seconds` (SDK `update`/`updateById`) 500s in prod: `42P08 inconsistent types deduced for parameter $2`. The param is used as both the `integer` column and `make_interval`'s `double precision` arg, so Postgres deduces conflicting types — fails for every value.

Fix: cast `::int::double precision`. Added a real-Postgres regression test (`internal/integration/auto_delete_test.go`); handler unit tests mock the DB, so this wasn't caught. Validated against Postgres 16 — old query fails, fix passes.

Only the PATCH path was broken; create/pause/resume/reaper work.

### DIFF
--- a/db/queries/sandboxes.sql
+++ b/db/queries/sandboxes.sql
@@ -447,10 +447,12 @@ LEFT JOIN closed_intervals ci ON ci.sandbox_id = p.id;
 -- is armed from now() — never retroactively from the pause — so applying a
 -- window can never destroy the sandbox in the same instant. On a non-paused
 -- sandbox the deadline stays NULL and FinalizePause arms it at the next pause.
+-- Keep the ::int::double precision cast: the param is used as both the integer
+-- column and make_interval's double arg; without it Postgres errors 42P08.
 UPDATE sandbox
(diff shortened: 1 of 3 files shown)


pavitrabhalla (approved):

awille-ai (commented):
Reviewed post-merge for security, scaling, concurrency, and edge cases.

No blocking issues found. This is a narrow query/type fix: the `auto_delete_seconds` parameter is still assigned to the integer column, and the `make_interval` usage now explicitly casts through `int` to `double precision`, which addresses the Postgres 42P08 ambiguity without changing the update predicate or widening caller-controlled behavior. I did not see new security exposure, concurrency behavior, or scaling risk in the changed path.

One non-blocking test-hardening note: the new integration test confirms the query no longer errors for 3600/0/nil and that `auto_delete_seconds` clears on nil. It could also assert `auto_delete_at` is cleared after nil, but the SQL path should already do that because the `CASE` expression yields NULL when the parameter is NULL. Not worth blocking this fix.

PR #189, opened 2026-07-08
feat: add paginated GET /activity endpoint for team audit log

mohamedsobhi777:
## Summary

Adds `GET /activity` — a paginated, filterable, team-scoped audit-log endpoint backing the console Audit Logs page. The console previously read the `activity` table directly (via Supabase) and rendered the whole list client-side; this moves it onto the REST API with real server-side pagination.

Mirrors the existing `GET /sandboxes` list: bare JSON array + `X-Total-Count`, shared `parsePageParams` + `resolveTotal`, new `ListActivityByTeamPaged` / `CountActivityByTeamPaged` sqlc queries. Gated on `settings:read`; team scoping is enforced in SQL from the authenticated context (never a request param).

**Filters** (all optional, AND'd): `category`, `status` (the console sends `error` for its Errors tab), `q` (case-insensitive substring across sandbox name, secret name, action, category), and a `start`/`end` `created_at` window. Every filter is covered by an existing `activity` index.

**Bounding:** unlike the sandbox/template lists (bounded by quota / few-per-team), the `activity` table is unbounded and ever-growing, and has no pre-existing unpaginated callers. So an omitted `limit` returns only the most recent page (capped at `maxPageSize` = 200), not the full history. `X-Total-Count` still reports the true total.

**Response** exposes every resource/actor FK — `sandbox_id`, `template_id`, `secret_id`, `actor_id` — so events of every category are identifiable, plus `metadata` (raw jsonb, `{}` when empty). A stable `id` tiebreaker keeps pagination deterministic under same-timestamp bursts.

## Test plan

- [x] `go build ./internal/api/... ./internal/db/...`, `go vet` (+ `-tags integration`)
- [x] `go test ./internal/api/...` — includes the OpenAPI route-sync test (endpoint added to `api/openapi.yaml`)
- [x] Integration tests (`internal/integration/activity_pagination_test.go`, run in CI with Postgres):
  - pagination window / created_at ordering / offset; `category`, `status=error`, `q` substring, and `start`/`end` filters; `X-Total-Count` including the `resolveTotal` short-circuit
  - omitted-`limit` cap: 205 rows → returns 200 with `X-Total-Count: 205`
  - resource-field exposure: `sandbox_id` / `template_id` / `secret_id` / `actor_id` present per category, `metadata` renders as `{}` not `null`

## Notes

- Paired with a console PR (superserve-ai/superserve) that consumes this endpoint. **Deploy this first** — the console and its docs nav entry depend on the route being live.
- Reviewed for tenancy (team-scoped, id from auth context), injection (all sqlc-parameterized, `q` LIKE-escaped), and resource bounds.
- Deferred follow-ups: a friendly template *name* in the UI needs `template_name` denormalized at write time; deep-`offset` paging over very large histories would benefit from keyset/cursor pagination (date filters mitigate today).

### DIFF
--- a/db/queries/activity.sql
+++ b/db/queries/activity.sql
@@ -40,3 +40,48 @@ SELECT * FROM activity
 WHERE team_id = $1 AND category = $2
 ORDER BY created_at DESC
 LIMIT $3;
+
+-- name: ListActivityByTeamPaged :many
+-- Paginated, filterable team activity list backing the console audit log.
+--
+-- Filters (all optional, AND'd): category equality, status equality (the
+-- console sends status='error' for the Errors tab), a case-insensitive
+-- substring across sandbox_name/secret_name/action/category, and a created_at
+-- window. Sort is created_at only (asc via the guarded CASE, otherwise the
+-- created_at DESC default + stable tiebreaker). A NULL @row_limit returns all
+-- rows, preserving the pre-pagination "return everything" default so
+-- unpaginated callers are unaffected.
+SELECT * FROM activity
+WHERE team_id = @team_id
+  AND (sqlc.narg('category')::text IS NULL OR category = sqlc.narg('category')::text)
+  AND (sqlc.narg('status')::text IS NULL OR status = sqlc.narg('status')::text)
+  AND (sqlc.narg('search')::text IS NULL
+       OR sandbox_name ILIKE '%' || sqlc.narg('search')::text || '%'
+       OR secret_name ILIKE '%' || sqlc.narg('search')::text || '%'
+       OR action ILIKE '%' || sqlc.narg('search')::text || '%'
+       OR category ILIKE '%' || sqlc.narg('search')::text || '%')
+  AND (sqlc.narg('created_after')::timestamptz IS NULL OR created_at >= sqlc.narg('created_after')::timestamptz)
+  AND (sqlc.narg('created_before')::timestamptz IS NULL OR created_at <= sqlc.narg('created_before')::timestamptz)
+ORDER BY
+  CASE WHEN @sort_by::text = 'created_at' AND @sort_dir::text = 'asc' THEN created_at END ASC,
+  created_at DESC,
+  -- id is a stable final tiebreaker so rows sharing a created_at (bulk events
+  -- written in one operation) keep a deterministic order across page boundaries.
+  id DESC
+LIMIT sqlc.narg('row_limit')::bigint
+OFFSET COALESCE(sqlc.narg('row_offset')::bigint, 0);
+
+-- name: CountActivityByTeamPaged :one
+-- Total rows matching the same filters as ListActivityByTeamPaged (ignoring
+-- pagination). Backs the X-Total-Count response header.
+SELECT COUNT(*) FROM activity
+WHERE team_id = @team_id
+  AND (sqlc.narg('category')::text IS NULL OR category = sqlc.narg('category')::text)
+  AND (sqlc.narg('status')::text IS NULL OR status = sqlc.narg('status')::text)
+  AND (sqlc.narg('search')::text IS NULL
+       OR sandbox_name ILIKE '%' || sqlc.narg('search')::text || '%'
+       OR secret_name ILIKE '%' || sqlc.narg('search')::text || '%'
+       OR action ILIKE '%' || sqlc.narg('search')::text || '%'
+       OR category ILIKE '%' || sqlc.narg('search')::text || '%')
+  AND (sqlc.narg('created_after')::timestamptz IS NULL OR created_at >= sqlc.narg('created_after')::timestamptz)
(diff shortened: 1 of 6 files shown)


db/queries/activity.sql
```diff
@@ -40,3 +40,48 @@ SELECT * FROM activity
...
+-- unpaginated callers are unaffected.
+SELECT * FROM activity
+WHERE team_id = @team_id
+  AND (sqlc.narg('category')::text IS NULL OR category = sqlc.narg('category')::text)
+  AND (sqlc.narg('status')::text IS NULL OR status = sqlc.narg('status')::text)
+  AND (sqlc.narg('search')::text IS NULL
+       OR sandbox_name ILIKE '%' || sqlc.narg('search')::text || '%'
+       OR secret_name ILIKE '%' || sqlc.narg('search')::text || '%'
+       OR action ILIKE '%' || sqlc.narg('search')::text || '%'
+       OR category ILIKE '%' || sqlc.narg('search')::text || '%')
```
pavitrabhalla, line 62: The `q` search (sandbox_name/secret_name/action/category ILIKE '%...%') isn't covered by any existing index — I checked the activity migrations and every index is a plain btree (idx_activity_team_time, idx_activity_category, etc.), none of which support a leading-wildcard ILIKE. That contradicts the PR description's claim that every filter is index-covered. Given this table is explicitly unbounded/ever-growing, a non-empty `q` forces a full scan of the team's activity rows. Worth a pg_trgm/GIN index, or at least flagging as a known gap.

internal/api/handlers_activity.go
```diff
@@ -0,0 +1,180 @@
...
+		return
+	}
+	createdBefore, err := parseTimeFilter(c.Query("end"))
+	if err != nil {
+		respondErrorMsg(c, "bad_request", "end must be an RFC3339 timestamp", http.StatusBadRequest)
+		return
+	}
+
+	category := nullableStr(c.Query("category"))
+	statusFilter := nullableStr(c.Query("status"))
```
pavitrabhalla, line 136: `category`/`status` go straight into the SQL equality filter with no allow-list validation. ListSandboxes validates its `status` param (handlers.go, sandboxStatusFilterValues) and 400s on an unrecognized value specifically because an unknown value would otherwise silently match zero rows. Same risk here — a typo'd `?status=Error` returns 200 with an empty page instead of a 400. The new integration tests only exercise valid category/status values. Intentional, or worth adding the same guard?

internal/api/handlers_activity.go
```diff
@@ -0,0 +1,180 @@
...
+// console Audit Logs page. Mirrors ListSandboxes: parsePageParams + optional
+// filters (category, status, q substring, created_at window) + resolveTotal +
+// X-Total-Count, returning a bare JSON array so omitting `limit` yields the
+// full list.
+func (h *Handlers) ListActivity(c *gin.Context) {
+	teamID, err := teamIDFromContext(c)
+	if err != nil {
+		return
+	}
+	if !h.requireTeamSandboxRead(c, teamID) {
```
pavitrabhalla, line 103: This gates on `settings:read` (via requireTeamSandboxRead), the same bucket already shared by sandboxes/templates/network/secrets. Now it also covers full audit-log read — a more sensitive scope (per-actor action history, secret-rotation events) — with no way to grant one without the other. rbac_phase4.go shows we already do fine-grained permissions elsewhere (roles:read, users:write). Was a dedicated activity:read considered, or is bundling into settings:read intentional for now?

internal/api/handlers_activity.go
```diff
@@ -0,0 +1,180 @@
...
+		Error:       a.Error,
+		Metadata:    meta,
+		CreatedAt:   a.CreatedAt,
+	}
+}
+
+// nullableUUIDString renders a nullable pgtype.UUID as a *string for JSON
+// (null when the column is NULL), matching how the console consumes
+// sandbox_id / secret_id.
+func nullableUUIDString(u pgtype.UUID) *string {
```
pavitrabhalla, line 71: nullableUUIDString duplicates two existing helpers with identical bodies: uuidPGTypeString (rbac_phase2b.go) and uuidPGPtrString (rbac_platform_sandboxes.go). Worth reusing one of those instead of a third copy?

pavitrabhalla (commented):
Reviewed the new endpoint end-to-end (pagination, filters, permission gating). A few things worth a look before merge — nothing blocking, but the index gap is worth confirming given the table's growth profile.

awille-ai (approved):

The tenant boundary looks correct (team ID comes from auth context, not request parameters), SQL remains parameterized, pagination is bounded, and the endpoint avoids the accidental unbounded scan that would have come from preserving the old 'no limit means everything' behavior.

I agree with the existing non-blocking review threads around future improvements (trigram indexing for substring search on a growing table, possible finer-grained audit-log permissions, and optional validation of known category/status values), but I don't consider any of those merge blockers.

PR #191, opened 2026-07-08
vmd: allocate template-build network slots from the authoritative allocator

meAmitPatil:
## Problem

Template builds and sandboxes share one kernel namespace space (`ns-<idx>` /
`veth-<idx>`), but they were allocated by **two independent allocators with no
authority between them**:

- Sandboxes go through vmd's `network.Manager` slot allocator (`slotOwner` map,
  reclaims freed indices), which runs 1 → `MaxSlots`.
- Builds used a separate `nextBuildSlot` counter (base 200) in the vm manager,
  and the build ran in a subprocess with its own `network.Manager`.

Two problems fell out:

1. **Overlapping ranges.** The "sandboxes stay 1–100, builds start at 200"
   assumption no longer holds — a host routinely runs far more than 200
   concurrent sandboxes, so sandbox slots extend well past 200 and overlap the
   build range. The only thing standing between a build and a sandbox on the
   same `ns-<idx>` was a cross-process `nsExists` TOCTOU check that both sides
   can pass before either creates the namespace.
2. **No reclamation.** `nextBuildSlot` only ever increments, so build slot
   indices climb for the life of the process and are never reused.

## Fix

Make vmd's **one authoritative allocator** own the build slot too. Before
spawning the subprocess, vmd reserves an index through the same allocator that
hands out sandbox slots (`ReserveSlot` → the `slotOwner` map), passes it to the
subprocess, and releases it — tearing down any kernel residue — when the build
exits (success, failure, or panic):

- New `network.Manager.ReserveSlot(owner)` claims a fresh index under the build's
  ID without building kernel state (the subprocess builds `ns/veth` itself). The
  reservation is what prevents a sandbox from being handed the same index — no
  disjoint range needed.
- `buildTemplateSync` reserves via `ReserveSlot` and `defer`s
  `CleanupVMOrNamespace` to release the index and reap any leftover namespace.
- Deleted the `nextBuildSlot` counter and the fixed 200 base.

Result: builds and sandboxes draw from a single allocator, so they can't collide;
the index is reclaimed on build exit instead of leaking upward; and the fragile
cross-process `nsExists` race is gone (the reservation is taken before the
subprocess creates the namespace, and the allocator already skips
kernel-present indices).

This is independent of the concurrency-limit change — it's a pre-existing
allocation bug, surfaced while reviewing that work.

### DIFF
--- a/internal/network/manager.go
+++ b/internal/network/manager.go
@@ -82,8 +82,9 @@ type Manager struct {
 	mu        sync.Mutex
 	devices   map[string]*VMNetInfo
 	freeSlots []int // recycled slot indices, guaranteed absent from slotOwner
-	nextSlot  int   // next new slot (used when freeSlots is empty)
-	maxSlot   int   // upper bound for new-slot allocation (0 = MaxSlots); WithExactSlot pins to one
+	nextSlot   int  // next new slot (used when freeSlots is empty)
+	maxSlot    int  // new-slot ceiling; meaningful only when slotPinned (WithExactSlot)
+	slotPinned bool // WithExactSlot: allocate exactly maxSlot or fail, never advance
 
 	// slotOwner is the single source of truth for slot allocation AND identity:
 	// slotOwner[idx] == the current owner of slot index idx, one of
@@ -158,7 +159,7 @@ type ManagerOption func(*Manager)
 // ReserveSlot; if that slot is somehow unavailable the build fails cleanly
 // instead of silently running on an index vmd never reserved.
 func WithExactSlot(idx int) ManagerOption {
-	return func(m *Manager) { m.nextSlot = idx; m.maxSlot = idx }
+	return func(m *Manager) { m.nextSlot = idx; m.maxSlot = idx; m.slotPinned = true }
 }
 
 // WithHTTPProxyPort sets the HTTP proxy port for egress REDIRECT rules.
@@ -793,7 +794,7 @@ func (m *Manager) claimSlotIndex(owner string) (int, error) {
 			m.freeSlots = m.freeSlots[:len(m.freeSlots)-1]
 		} else {
 			ceiling := MaxSlots
-			if m.maxSlot > 0 {
+			if m.slotPinned {
 				ceiling = m.maxSlot // WithExactSlot: allow only the pinned index
 			}
 			if m.nextSlot > ceiling {
--- a/internal/network/manager.go
+++ b/internal/network/manager.go
@@ -83,6 +83,7 @@ type Manager struct {
 	devices   map[string]*VMNetInfo
 	freeSlots []int // recycled slot indices, guaranteed absent from slotOwner
 	nextSlot  int   // next new slot (used when freeSlots is empty)
+	maxSlot   int   // upper bound for new-slot allocation (0 = MaxSlots); WithExactSlot pins to one
 
 	// slotOwner is the single source of truth for slot allocation AND identity:
 	// slotOwner[idx] == the current owner of slot index idx, one of
@@ -151,12 +152,13 @@ func (m *Manager) SetEgressProxy(p *EgressProxy) {
 // ManagerOption configures optional Manager behavior.
 type ManagerOption func(*Manager)
 
-// WithStartSlot sets the starting slot index for network allocation. Used by a
-// template-builder subprocess to build exactly the one slot vmd reserved for it
-// via ReserveSlot — the reservation, not a disjoint range, is what keeps a build
-// from colliding with vmd's sandbox slots.
-func WithStartSlot(idx int) ManagerOption {
-	return func(m *Manager) { m.nextSlot = idx }
+// WithExactSlot pins the Manager to exactly slot idx: it claims that one index
+// or fails with ErrNoSlots, never advancing to idx+1. A template-builder
+// subprocess uses this to build precisely the slot vmd reserved for it via
+// ReserveSlot; if that slot is somehow unavailable the build fails cleanly
+// instead of silently running on an index vmd never reserved.
+func WithExactSlot(idx int) ManagerOption {
+	return func(m *Manager) { m.nextSlot = idx; m.maxSlot = idx }
 }
 
 // WithHTTPProxyPort sets the HTTP proxy port for egress REDIRECT rules.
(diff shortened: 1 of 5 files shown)


internal/vm/build.go
```diff
@@ -114,7 +114,14 @@ func (m *Manager) buildTemplateSync(ctx context.Context, buildVMID string, req B
...
 
-	slotIndex := int(m.nextBuildSlot.Add(1)) + 199
+	// Reserve a slot from vmd's authoritative allocator so the subprocess's
+	// ns-<idx>/veth-<idx> can't collide with a sandbox slot, and release it (with
+	// any kernel residue) when the build exits — success, failure, or panic.
+	slotIndex, err := m.netMgr.ReserveSlot(buildVMID)
+	if err != nil {
+		return nil, fmt.Errorf("reserve build network slot: %w", err)
+	}
+	defer m.netMgr.ReleaseSlot(buildVMID, slotIndex)
```
pavitrabhalla, line 124: This always goes through `CleanupVMOrNamespace` → `cleanupFull`, which shells out `ip link del veth-N` / `ip netns del ns-N` — even on the success path, where the subprocess's own `setupNetwork` cleanup already tore down its ns/veth before exiting. Harmless no-op today, but is the unconditional kernel teardown here meant only for the failure/panic case (subprocess killed before it could clean up), or is the redundancy on the success path intentional belt-and-suspenders?

internal/network/manager.go
```diff
@@ -759,6 +762,23 @@ func (m *Manager) releaseIfOwned(idx int, owner string) bool {
 	return true
 }
 
+// ReserveSlot claims a fresh, unused slot index under owner without building any
+// kernel network state. The caller — a template-builder subprocess — creates
+// ns-<idx>/veth-<idx> itself; the reservation is what stops vmd's sandbox
+// allocator from handing the same index to a VM, so builds and sandboxes draw
+// from one authoritative allocator instead of racing disjoint ranges. Pair every
+// ReserveSlot with a ReleaseSlot.
+func (m *Manager) ReserveSlot(owner string) (int, error) {
```
pavitrabhalla: `ReserveSlot` claims a fresh index through the normal owned/`nsExists` allocator loop, while `ReserveSlotsAbove` below pins already-known indices via `reserveSlotLocked`, bypassing those checks entirely. Same verb, different contract on the same type — worth a more distinct name (e.g. `ClaimFreshSlot`) so the two don't read as variations of the same operation?

internal/network/manager.go
```diff
@@ -772,7 +792,11 @@ func (m *Manager) claimSlotIndex(owner string) (int, error) {
 			idx = m.freeSlots[len(m.freeSlots)-1]
```
pavitrabhalla, line 813: This branch pops from `freeSlots` without checking `maxSlot` — only the `else` (fresh `nextSlot`) path enforces the pin. Not reachable today since `WithExactSlot` is only ever used on a freshly-constructed `Manager` with an empty `freeSlots` before the single `SetupVM` call, but the "claims exactly idx" contract on `WithExactSlot` isn't actually enforced here — it just happens to hold given current callers. Worth guarding this branch too (`if m.maxSlot > 0 && idx != m.maxSlot { continue }`), or is relying on caller discipline intentional?

awille-ai (approved):

The important race from the earlier review has been addressed: vmd reserves the build slot through the authoritative allocator, the builder is pinned with `WithExactSlot` instead of treating the reserved value as only a starting point, and the slot is released with namespace cleanup on build exit. The added tests cover no build/sandbox collision, exact-slot non-advance, and reclamation after release.

I do not see a remaining blocker. The main operational tradeoff is expected: template builds now consume the same bounded network slot pool as sandboxes, so under extreme host saturation a build may fail with `ErrNoSlots` rather than racing into an unreserved namespace. That is the safer behavior.

meAmitPatil:
Thanks both — addressed in `e9a120b`:

**`ReleaseSlot` could tear down a colliding live VM (Codex, `manager.go:780`)** — fixed. `ReleaseSlot` now releases strictly by `(owner, idx)` (claim `teardownOwner` → `cleanupFull` → `releaseIfOwned`), never routing through `CleanupVMOrNamespace`'s tracked branch. So a caller-supplied build id that happens to equal a live sandbox's VM id can no longer run `CleanupVM` on that sandbox or leak the reserved index. Added `TestReleaseSlot_ReleasesByIndexNotTrackedVM` pinning it.

**`freeSlots` branch doesn't enforce the `WithExactSlot` pin (@pavitrabhalla `manager.go:793`)** — fixed. A pinned Manager now skips the `freeSlots` path entirely (`!m.slotPinned && len(m.freeSlots) > 0`), so "claims exactly `maxSlot` or fails" is actually enforced rather than just holding by caller discipline. (This is also the real residual behind the Codex `build.go:124` note — the builder now uses `WithExactSlot`, which pins and fails instead of advancing to `idx+1`.)

**`ReserveSlot` vs `ReserveSlotsAbove` naming (@pavitrabhalla `manager.go:772`)** — renamed to `ClaimFreshSlot`, with a doc note calling out the contrast with `ReserveSlotsAbove` (which pins known indices via `reserveSlotLocked`).

**`cleanupFull` on the success path (@pavitrabhalla `build.go:124`)** — yes, intentional belt-and-suspenders. On success the subprocess's own `setupNetwork` cleanup already tore down its ns/veth, so it's a tolerant no-op; the unconditional teardown is there for the failure/panic path where the subprocess is SIGKILLed (past the 30s `WaitDelay`) before its defers run, leaving `ns-<idx>`/`veth-<idx>` behind for vmd to reap.

Build/vet clean under `GOOS=linux`; full network suite green in a linux container. Still `BEHIND` main — I'll merge main in and run `-race` on the merged tree before this goes in.

awille-ai (approved):
Re-reviewed the current head after the additional allocator/cleanup changes. I did not find any critical or high-severity issues, so I’m approving.

The previously identified edge cases are now addressed:
- the template-builder is pinned to exactly the reserved slot and cannot fall through to another recycled or fresh slot;
- build-slot cleanup is keyed directly by `(owner, reserved index)` rather than the tracked-VM cleanup path, so a caller-supplied build ID collision cannot tear down a live sandbox;
- cleanup holds the slot under `teardownOwner` until kernel cleanup finishes, preventing premature reuse;
- tests cover collision, exact-slot behavior, and reclamation.

The remaining open comment about redundant cleanup on the successful subprocess path is non-blocking belt-and-suspenders behavior. CI is green on the current head.

PR #195, opened 2026-07-09
adding impersonation support

awille-ai:
The impersonation needs support on the API side, and that was missed in the earlier checkin.

### DIFF
--- a/internal/api/apikey_cache.go
+++ b/internal/api/apikey_cache.go
@@ -47,6 +47,8 @@ func apiKeyCacheTTLFromEnv() time.Duration {
 type apiKeyCacheEntry struct {
 	id        string
 	teamID    string
+	name      string
+	scopes    []string
 	createdBy pgtype.UUID
 	expiresAt pgtype.Timestamptz
 	fetchedAt time.Time
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -1450,7 +1450,9 @@ func (h *Handlers) GetSandboxByID(c *gin.Context) {
 	} else if canWrite {
 		resp = h.sandboxToResponseWithToken(sandbox)
 	}
-	resp.Secrets = h.fetchSandboxSecretBindings(c.Request.Context(), sandboxID)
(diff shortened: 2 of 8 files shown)


pavitrabhalla:
Re-reviewed after the addressing-review commit. The P1 write-RBAC bypass is fixed (actor_id is now suppressed for `__console_impersonation__` keys, so `requireTeamSandboxWrite`/`requireTeamSettingsWrite` fail closed), and the audit-log concern was already a non-issue since `ListActivity` gates on `requireTeamSettingsRead`, not the sandbox-read helper.

Still open: the secret-binding leak. `GetSandboxByID` (internal/api/handlers.go) unconditionally sets `resp.Secrets = h.fetchSandboxSecretBindings(...)` after the `requireTeamSandboxRead` check, and that check now passes for a `platform:sandbox:read` impersonation key. So a console-impersonation key scoped to sandbox-read only can still read `env_key`/`secret_name` for every secret bound to the sandbox via `GET /sandboxes/:id` — the same class of leak flagged in the original codex review. `TestConsoleImpersonation_SandboxReadResponsesRemainScoped` checks that `access_token` is omitted but never checks `secrets`, so this wasn't caught by the new tests either. Was this intentionally left for a follow-up, or should it get the same treatment (omit for impersonation reads, or gate behind a broader scope)?

awille-ai (commented):
The current branch already addresses both concerns:

- `actor_id` is suppressed for `__console_impersonation__` keys in `internal/api/middleware.go`, so write helpers do not inherit customer-owner/admin RBAC from `created_by`.
- `GET /activity` remains on `requireTeamSettingsRead` in `internal/api/handlers_activity.go`, so sandbox-read impersonation keys do not gain audit-log access.
- `GET /sandboxes/:id` omits `secrets` for impersonation reads in `internal/api/handlers.go`, and the integration test now asserts that `secrets` is absent for `platform:sandbox:read` keys.

I re-ran the targeted API and integration tests after those changes.

meAmitPatil (approved):

PR #198, opened 2026-07-09
vmd: verify a network slot's namespace is clear before recycling it

pavitrabhalla:
## Summary

- Fixes the root cause behind a prod incident where `POST /sandboxes` was failing (100% of creates for a stretch) with `RestoreSnapshot` errors: `Open tap device failed: Device or resource busy (os error 16)` on `tap0`.
- `Pool.Return` recycled a destroyed VM's namespace/TAP straight to the next claimant based on process-liveness signals — a 500ms SIGKILL-wait poll, or systemd reporting a unit "not active" — that don't actually guarantee the kernel has released the TAP fd. SIGKILL is asynchronous against a process blocked in uninterruptible I/O (a UFFD page fault resolving against local SSD is exactly that), and "not active" also covers "still deactivating". Under host I/O contention this let a new VM's Firecracker race the old occupant's still-open `tap0`.
- `Return` now hands off to an async verifier (`verifyAndRecycle`) that re-kills defensively and polls until the namespace has no attached processes before the slot becomes claimable, falling back to a full teardown instead of recycling if it's still occupied after 20s (comfortably past `firecracker@.service`'s `TimeoutStopSec=10`).

- Fixed this at the single choke point (`Pool.Return`/`Claim`) rather than patching every caller that can return a slot (`DestroyVM`'s SIGKILL fallback, the reconciler's `markStale`, the startup reattach sweep) — defends all of them uniformly, present and future.
- Verification runs in its own goroutine, off the caller's hot path — `DestroyVM` and the reconciler still return immediately. This only delays how soon a *destroyed* VM's slot becomes reusable, never blocks a create.
- Reused the existing `killProcessesInNs` primitive (already used by `SweepOrphanNamespaces`), refactored to expose a `pidsInNs` lookup — avoids shelling out to `ip netns pids`, and incidentally closes the gap where an untracked VM (e.g. right after a vmd restart) could skip its kill entirely.
- Deliberately did not walk back the create-concurrency ceilings from #190 (`MaxConcurrentRestores`, net pool sizes) — this fix makes recycling safe under that concurrency instead of trading it away.

## Testing

- `go build ./...` and `go vet ./...` clean under `GOOS=linux`.
- Full test suite green, including `-race`.
- Added three tests pinning the actual behavior: recycles once the namespace is confirmed clear, withholds a slot from `Claim` while still occupied, and tears down (never recycles) if a namespace never clears within the max wait.
- Validated live in prod as an operational mitigation before this PR: manually shrank `VMD_NET_POOL_RECYCLE_SIZE` on the affected host and restarted vmd — error rate went from 100% of creates failing to 0/7 failing over the following ~17 minutes, with zero recurrences of the tap0 error. This PR is the permanent fix so the recycle pool can be restored to full size.

### DIFF
--- a/internal/network/manager.go
+++ b/internal/network/manager.go
@@ -850,9 +850,10 @@ func (m *Manager) SweepOrphanNamespaces(keep map[string]bool) (swept int) {
 }
 
 // killProcessesInNs SIGKILLs every process whose net namespace matches
-// /run/netns/<name>. Returns the number of pids signalled.
+// /run/netns/<name>. Returns the number of pids signalled. Best-effort: a
+// failed /proc scan (see pidsInNs) just means nothing gets killed this pass.
 func killProcessesInNs(name string) int {
-	pids := pidsInNs(name)
+	pids, _ := pidsInNs(name)
 	killed := 0
 	for _, pid := range pids {
 		if err := syscall.Kill(pid, syscall.SIGKILL); err == nil {
@@ -864,20 +865,23 @@ func killProcessesInNs(name string) int {
 
 // pidsInNs returns the PIDs whose net namespace matches /run/netns/<name>,
 // found by comparing /proc/<pid>/ns/net's inode against the namespace file's.
-// A dead or nonexistent namespace yields nil, not an error — callers treat
-// "can't stat it" the same as "nothing is in it."
-func pidsInNs(name string) []int {
+// ok is false only when the /proc scan itself failed (e.g. transient
+// resource pressure) — a genuine "don't know" that callers must not treat as
+// "clear" the way an actually-empty scan is; conflating the two would let a
+// still-occupied namespace be recycled, reintroducing the exact race this
+// guards against. A namespace file that no longer exists can't have
+// anything attached to it, so that case is a confident (nil, true).
+func pidsInNs(name string) (pids []int, ok bool) {
 	nsPath := netnsDir + "/" + name
 	nsStat, err := os.Stat(nsPath)
 	if err != nil {
-		return nil
+		return nil, true
 	}
 	nsIno := nsStat.Sys().(*syscall.Stat_t).Ino
 	procs, err := os.ReadDir("/proc")
 	if err != nil {
-		return nil
+		return nil, false
 	}
-	var pids []int
 	for _, e := range procs {
 		if !e.IsDir() {
 			continue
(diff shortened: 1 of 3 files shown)


meAmitPatil (commented):
Correct fix — the quarantine is the right shape. Marking the slot `poolOwner` but keeping it out of the `recycled` channel until `verifyAndRecycle` confirms the namespace is empty means `Claim` can never pop a slot whose `tap0` is still held, and doing it in one goroutine off the caller's path keeps destroys/creates non-blocking. Namespace-empty is a good jailer-agnostic proxy for "tap released", the `wg`/`stopCh` ordering avoids a send-on-closed-channel, and the 20s teardown fallback (> `TimeoutStopSec=10`) stops a stuck ns from poisoning the pool. Re-killing at the choke point also closes the untracked-VM gap.

Two notes, one worth acting on before merge:

1. (worth doing) Poll cost scales worst exactly under the condition that triggers this. `pidsInNs` scans all of `/proc` each tick; the trigger is I/O contention → slow teardown → verifiers linger up to 20s and pile up, each scanning every 25ms. That's a lot of added VFS load on an already-contended host, off the hot path but able to prolong recovery. Suggest bumping `defaultVerifyPollInterval` to 100–250ms (negligible slot-reuse latency vs. a 10s stop window, ~4–10× less scan load).

2. (minor) A transient `os.ReadDir("/proc")` failure returns nil and reads as "namespace clear", which would recycle a still-occupied slot. Very low probability, but the downside is re-triggering the exact race — consider treating only a successful, empty scan as clear rather than any nil.

Nice that the recycle-size mitigation already drove failures to 0 in prod; this is the right permanent version.

meAmitPatil (approved):

pavitrabhalla:
Both addressed in 6e74368:

1. Bumped `defaultVerifyPollInterval` 25ms → 150ms — cuts /proc scan load ~6x for negligible added slot-reuse latency against the 10s+ stop window.
2. `pidsInNs` now returns `(pids []int, ok bool)` — `ok=false` on a failed `/proc` readdir, distinct from a genuinely empty scan. `verifyAndRecycle` keeps polling on `ok=false` instead of recycling, so a transient scan failure can't be misread as "clear." Added a test (`TestPoolReturn_FailedScanIsNotTreatedAsClear`) pinning that.

Full suite still green under `-race`.

PR #201, opened 2026-07-10
feat: preview URL authentication — per-sandbox public/private policy with preview tokens

mohamedsobhi777:
## Motivation

Customer feedback during onboarding: every process bound to a port in a sandbox is publicly reachable at `https://{port}-{id}.<sandbox-domain>` with no authentication. Teams that front previews with their own authenticated proxy still can't accept a public URL existing at all, and benchmarked us against Daytona's preview-token flow (`x-daytona-preview-token` header + per-sandbox `public` flag). This PR adds that capability: an opt-in, per-sandbox preview access policy enforced at the edge proxy.

## What this adds

**Policy** — `preview_access: "public" | "private"` per sandbox. Default `public` (today's behavior, fully backward compatible). Settable at create — so a private sandbox never has a public window — and toggleable live via `PATCH /sandboxes/:id`.

**Tokens** (`internal/auth/preview_token.go`) — `spv1.<base64url claims>.<hmac-sha256>`, signed with the existing `SANDBOX_ACCESS_TOKEN_SEED`, domain-separated from the boxd access token (cross-acceptance rejected in both directions, tested). Claims: bare sandbox UUID, token generation (`preview_token_version`), optional port scope, optional expiry (≤ 7 days). Verification is stateless at the proxy; MAC is checked before any payload parsing; 512-byte input cap; constant-time compare.

**Edge proxy enforcement** (`internal/proxy/preview.go`) — on numeric-port requests to a private sandbox, accepts the token from:
- `X-Superserve-Preview-Token` header (machine clients / customer proxies injecting auth upstream),
- a `__Host-` cookie (browsers), or
- a one-time `?superserve_preview_token=` query param: verified, exchanged for the cookie (`Secure; HttpOnly; SameSite=None; Partitioned` for iframe embedding), and stripped from the URL via 302 — so signed links work in a browser and the token never lingers in the address bar, history, or Referer.

Details that matter: CORS preflights pass through (they carry no credentials by spec); non-GET and WebSocket-upgrade requests with a query token are accepted without the redirect (WS clients can't follow redirects — keeps HMR working); all three carriers are scrubbed before the request enters the VM, so the sandbox's own code never sees the credential; auth runs before the status check so unauthenticated callers can't distinguish paused from running; every deny is the same 401.

Fail-closed everywhere: missing seed, `private` with an unplumbed version, and unknown future policy values all deny rather than fall open.

**vmd plumbing** — `preview_access` + `preview_token_version` ride the `RestoreSnapshot` spec (same slot as `team_id`/`owner_id`), persist in the BoltDB record (survive vmd restarts), and are served to the proxy through the existing local `/instances` lookup — zero added per-request latency, proxy stays DB-free. New `UpdateSandboxPreviewPolicy` RPC applies live toggles/rotation to the record in any VM state; NotFound is tolerated (the next restore seeds policy from the DB row).

**Control plane**
- Migration `20260710000001`: two `sandbox` columns with defaults + CHECK constraints.
- `POST /sandboxes` and `PATCH /sandboxes/:id` accept `preview_access`; responses include it. The PATCH updates the host record **before** the DB row and hard-fails (500) if enforcement can't be applied — a success response never means "recorded but not enforced".
- `POST /sandboxes/:id/preview-token` mints (optional `port`, `expires_in_seconds`); response carries the carrier names so clients don't hardcode them. Tokens are deliberately **never** included in any sandbox GET response.
- `POST /sandboxes/:id/preview-token/rotate` bumps the generation, revoking every outstanding token (revocation without a sandbox restart).
- Resume paths re-carry the row's policy — the stateless restore fallback builds a fresh instance record, and omitting it would silently reopen a private sandbox.
- Activity log + analytics events for policy changes and rotations; OpenAPI spec updated (sync test passes).

## Backward compatibility

- Existing sandboxes: records without the new fields decode to zero values = public; behavior byte-for-byte unchanged.
- Old clients: never send `preview_access` → public; API changes are additive.
- Old control plane vs. new schema: safe — sqlc expands `SELECT *` to explicit column lists, and both columns have defaults.

## Deploy order (matters)

vmd + proxy hosts **before** the control plane (per environment): an old vmd ignores the new proto fields, so a control plane exposing `private` first could accept a policy no host enforces. Order: `CD — Migrate` → `Deploy VMD` → `Deploy Proxy` → `Deploy API`. No template reseed needed (boxd untouched).

## Testing

- `internal/auth`, `internal/proxy`, `internal/api` suites pass, including ~30 new adversarial cases: wrong sandbox/version/port, expired, payload-spliced, boxd-token cross-use (both directions), status-leak, credential-scrubbing, stale-cookie + fresh-link, unknown-policy fail-closed, vmd-push failure semantics.
- `CGO_ENABLED=0 GOOS=linux go build ./...` and `go vet` (incl. `-tags integration`) clean.
- [ ] Staging verification (branch dispatch of the four workflows + end-to-end matrix: 401/mint/header/signed-link-cookie/toggle/rotate/pause-resume/backward-compat) — will run before marking ready for review.

## Follow-ups (separate PRs, superserve monorepo)

- SDK methods (TS + Python) for minting tokens/signed URLs; MCP tool update; docs — including rewording the three places that currently promise "no authentication".
- Console UI: create-form toggle, settings toggle, signed-URL open/iframe in the preview section.
- Team-level default policy for orgs that want private-by-default.


awille-ai (changes requested):
Blocking security review — requesting changes.

## High: unpublished ports are still routable

The proxy still treats every numeric `{port}-{sandbox}` hostname as a valid preview target. `preview_access=private` only changes whether a token is required; it does not require the port to have been explicitly published. A sandbox-wide token (`port=0`) therefore authenticates every listening port, and a caller can mint a token for any numeric port whether or not a preview URL was created for it.

This does not fix the prior exposure model. The proxy needs an explicit per-sandbox published-port allowlist and must return **404 before proxying** for any port that is not in that allowlist. Example: publishing/authenticating port 3001 must not make `3000-{sandbox-host}` reachable; that request should return 404 even if a valid sandbox token is supplied.

The allowlist must be enforced at the edge/data plane and persisted through restart/pause/resume. Relying on whether a process happens to be listening is not sufficient.

## High: tokens are not independent per preview URL

The current design has one sandbox-wide `preview_token_version`, allows sandbox-wide tokens (`Port=0`), and deterministically produces the same token for identical claims. Rotation revokes every token for the sandbox. That does not provide independently managed authentication credentials for multiple preview URLs in the same sandbox.

Each explicitly published port/preview URL should have its own credential identity and revocation lifecycle (for example, a per-port publication record with its own token version/secret or token ID). Creating or rotating the credential for port 3001 must not replace or revoke the credential for port 3000, and tokens should not be valid for other ports.

I recommend removing sandbox-wide tokens entirely from this flow and requiring every token to be scoped to an existing published port.

## High: `settings:read` can mint a live data-plane credential

`MintSandboxPreviewToken` calls `requireTeamSandboxRead`. A read-only user can therefore mint a non-expiring token that grants access to an application running inside the sandbox. This is a privilege escalation from control-plane read access to data-plane access, especially because the token can currently cover every port.

Token creation should require an explicit write/manage-preview permission (at minimum the existing sandbox/settings write permission), with activity logging for each mint. Reading sandbox configuration should never be enough to generate a new access credential.

CI is green, and the cryptographic verification, credential scrubbing, fail-closed behavior, and host-policy plumbing otherwise look thoughtful. I am requesting changes because the three issues above affect the core authorization boundary and the explicit deny-by-default requirement.

awille-ai (changes requested):
Re-review against the same criteria as my earlier review.

The three original blocking findings are addressed:
- private sandboxes now enforce an explicit per-port allowlist at the proxy and return 404 for unpublished ports before authentication;
- tokens and token versions are scoped independently per published port, with no sandbox-wide token;
- minting/publishing/rotation require sandbox write access, and token minting is activity-logged.

However, I found one remaining high-severity authorization-state bug, so I’m still requesting changes.

## High: failed unpublish can leave the port routable with no recoverable retry

`UnpublishSandboxPreviewPort` deletes the DB row first, then calls `syncPreviewPortsToHost`. If the VMD/host push fails, the handler returns 500, but the host’s enforcement copy still contains the port and its old token version, so the supposedly unpublished URL remains reachable with existing tokens.

More importantly, retrying the same DELETE does not repair the state: `UnpublishPort` returns 0 rows because the DB row is already gone, and the handler returns 404 before attempting another host sync. That can leave the stale published port active indefinitely until some unrelated policy/port update or a pause/resume refreshes the host record.

This is blocking because unpublish is a security revocation operation. A transient host-sync failure must not create an unrecoverable split where the DB says “not published” while the proxy continues routing it.

Please make the operation retry-safe/atomic from the caller’s perspective. Reasonable options include:
- treat DELETE as idempotent and always push the full current DB allowlist even when the row is already absent;
- or push the intended new allowlist before committing/removing the DB row;
- or persist an outbox/pending-sync state with automatic retry and fail closed at the proxy.

At minimum, a second DELETE after a failed first attempt must re-sync the empty/current allowlist rather than returning 404 without touching the host. Add a test covering: host push fails after DB deletion → retry DELETE → host receives the corrected allowlist.

CI is green, and aside from this consistency issue the revised design now matches the requested deny-by-default and per-preview credential model.

awille-ai (changes requested):
Re-review of the current head: the previous three findings are substantially addressed, but I still see two blocking authorization issues.

## High: the published-port allowlist is only enforced for `private` sandboxes

`enforcePreviewAccess` returns immediately for `PreviewAccess == "public"`, before consulting `PreviewPorts`. That preserves the old behavior where every listening numeric port is routable. It does not meet the required boundary that **only explicitly created/published preview URLs exist at the proxy**, regardless of whether that published URL is public or token-authenticated.

A public preview should mean “this explicitly published port needs no token,” not “all ports in this sandbox are reachable.” If port 3001 is the only published preview, `3000-{sandbox-host}` must return 404 in both public and private modes.

Please enforce publication first for every numeric preview request:
1. port absent from the published set → 404;
2. port published and access mode public → route without a token;
3. port published and access mode private → require that port’s token.

Backward compatibility can be handled as an explicit migration/legacy mode if required, but the new preview-URL model must not leave arbitrary ports addressable.

## High: full-snapshot VMD pushes can race and reopen/re-authorize stale ports or token versions

The API mutates PostgreSQL and then independently loads/pushes the entire `port -> version` map to VMD. These operations are not serialized or revision-checked. Concurrent publish/unpublish/rotate requests can apply snapshots out of order.

Concrete exposure case:
1. unpublish 3000 computes and pushes a set without 3000;
2. a concurrent publish/rotate operation previously loaded a snapshot that still contains 3000;
3. that stale snapshot reaches VMD afterward;
4. the DB deletes 3000, but the proxy again routes it because VMD now contains the stale entry.

The same ordering problem can push an older token version after rotation, temporarily making revoked tokens valid again.

This needs monotonic/serialized enforcement updates, not best-effort full snapshots. Viable approaches include a sandbox-scoped policy revision included in every DB mutation and VMD update (VMD rejects revisions <= current), a transactional outbox/ordered reconciler, or a sandbox-scoped lock that covers mutation + authoritative reload + push. Unpublish and rotation must remain fail-closed under concurrent calls.

CI is green, and the per-port token format, write permission, 404 path for unpublished private ports, and independent per-port rotation are good improvements. I’m requesting changes because the two issues above still affect the core proxy authorization boundary.

awille-ai:
I think this would be safer and easier to roll out if it were split into phases.

The end state makes sense as a per-port model: users can expose any number of ports, and each port should independently be one of:

* unpublished → 404
* published + public → routed without Superserve auth
* published + private → routed only with a token scoped to that sandbox and port

That suggests the published port should be the primary resource/state, rather than `preview_access` being primarily sandbox-wide. A sandbox-level setting could still exist as a default for newly published ports, but enforcement should be per port.

I’d suggest phasing this roughly as:

1. **Explicit port publication**

   * Existing sandboxes remain in a legacy mode.
   * New/updated sandboxes only route explicitly published ports.
   * Users can publish as many ports as needed.
   * Published ports remain public for now.
   * This alone adds meaningful security value by preventing accidental exposure of every listening port.

2. **Per-port public/private policy**

   * Each published port gets its own access mode.
   * Unpublished ports remain 404 regardless of token.

3. **Per-port token authentication**

   * Tokens are scoped to sandbox + port.
   * Each port has its own token version and rotation.
   * Header auth can likely ship before the browser signed-link/cookie flow.

4. **Browser signed links and UI/SDK support**

   * Query-token bootstrap, cookies, redirects, WebSockets, iframe behavior, console integration, etc.

I’m also hesitant about the “deploy order matters” requirement. Some rollout ordering is unavoidable, but security correctness should not depend only on humans deploying services in the right sequence.

Could we add both:

* product rollout flags, e.g. `preview_port_publication_enabled`, `preview_port_private_access_enabled`, and `preview_signed_links_enabled`
* host capability checks, e.g. `preview_ports_v1` and `preview_port_tokens_v1`

The API should reject enabling a policy if the sandbox’s assigned host cannot enforce it. That way a partially deployed fleet results in a failed operation rather than a successful API response for a security policy that is not actually active.

Importantly, feature flags should control whether users can enable the feature, not weaken enforcement. Once a port is recorded as private, the proxy should always fail closed.

awille-ai (changes requested):
Re-review of the current head, including the earlier security findings and the later phasing/rollout comment.

The earlier authorization findings appear addressed:
- unpublished ports are denied for strict sandboxes before auth;
- policy and token lifecycle are per published port;
- mint/publish/rotate require write access;
- unpublish is idempotent and retryable;
- policy revisions prevent stale concurrent snapshots from reopening ports or restoring old

mohamedsobhi777:
Superseded by the independently deployable stack requested in [the staged review](https://github.com/superserve-ai/sandbox/pull/201#issuecomment-5007470378) and confirmed in the follow-up meeting:

1. [#239](https://github.com/superserve-ai/sandbox/pull/239) — explicit preview-port publication
2. [#240](https://github.com/superserve-ai/sandbox/pull/240) — per-port public/private policy
3. [#242](https://github.com/superserve-ai/sandbox/pull/242) — per-port header-token authentication
4. [#243](https://github.com/superserve-ai/sandbox/pull/243) — browser signed-link authentication

Client/SDK/docs companion: [superserve#287](https://github.com/superserve-ai/superserve/pull/287).

Each backend PR has its own capability and fail-closed mixed-version boundary. There are no preview-auth product feature flags; the stacked PR and deployment order is the rollout mechanism. Closing this bundled PR so review can continue on the isolated phases.

PR #202, opened 2026-07-10
proxy: accept a set of sandbox domains via PROXY_DOMAINS

pavitrabhalla:
## Summary

During the edge DNS/LB transition, sandbox traffic arrives on both the legacy hostname and the region-prefixed hostname, and the same proxy deployment has to serve both. This adds `PROXY_DOMAINS` (comma-separated, whitespace-tolerant); when unset, the single-value `PROXY_DOMAIN` is used as before, so current deploys are unaffected. The effective list is logged at startup.

## Technical decisions

- `ParseHost`/`ParseRequest`/`NewHandler` now take a domain slice; per-domain semantics (port stripping via `SplitHostPort`, case-sensitive suffix match, shared-host exact match) are unchanged — only the number of accepted suffixes grows.
- The proxy never constructs sandbox URLs from the configured domain (it only validates Hosts and echoes them back), so nothing needed a canonical pick; the `Handler.domains` field documents that the first entry is canonical should that ever change.
- The LB health-check gate in `cmd/proxy/main.go` checks all configured suffixes, so the boxd-label lockdown holds on every accepted domain.
- `EDGE_PROXY_DOMAIN` (control plane, `internal/config`) is deliberately untouched: the control plane keeps minting URLs on one canonical domain per cell; the proxy accepting more hostnames is what unblocks the DNS cutover.

## Testing

- New table-driven tests: multi-domain `ParseHost` (each domain accepted, unknown/near-miss rejected, port-suffixed Hosts, case variance), shared-host routing on a non-first domain, handler-level accept/reject, and `proxyDomains()` fallback/precedence/whitespace handling.
- `gofmt`, `go vet`, `go test ./cmd/proxy/... ./internal/proxy/...` all pass; `GOOS=linux go build ./...` clean.

### DIFF
--- a/.github/workflows/scripts/deploy-proxy.py
+++ b/.github/workflows/scripts/deploy-proxy.py
@@ -14,6 +14,8 @@ Env vars:
   VMD_INSTALL_DIR            required — bin install dir on the host
   SHA                        required — commit SHA (only first 8 chars used)
   PROXY_DOMAIN               required — host suffix the proxy serves (e.g. sandbox.superserve.ai)
+  PROXY_DOMAINS              optional — comma-separated host suffixes; overrides
+                             PROXY_DOMAIN on the proxy when set (DNS transitions)
   SANDBOX_ACCESS_TOKEN_SEED  optional — hex, >=32 bytes (>=64 hex chars)
   PROXY_ALLOWED_ORIGINS      optional — comma-separated origin patterns
   REQUIRE_DATA_PLANE         optional — "", "0", or "1"
@@ -44,10 +46,14 @@ def main() -> int:
         return 1
     # Optional multi-domain set for DNS/hostname transitions; the proxy
     # prefers PROXY_DOMAINS over PROXY_DOMAIN when both are present.
+    # Validated per-entry (not as one blob) so a space-separated list
+    # fails loudly here instead of deploying a domain string that can
+    # never match a Host — the proxy only splits on commas.
     proxy_domains = os.environ.get("PROXY_DOMAINS", "")
-    if proxy_domains and not re.fullmatch(r"[A-Za-z0-9.,\- ]+", proxy_domains):
-        print("ERROR: PROXY_DOMAINS contains disallowed characters", file=sys.stderr)
-        return 1
+    for entry in filter(None, (s.strip() for s in proxy_domains.split(","))):
+        if not re.fullmatch(r"[A-Za-z0-9.\-]+", entry):
+            print("ERROR: PROXY_DOMAINS contains an invalid domain", file=sys.stderr)
+            return 1
     access_seed = os.environ.get("SANDBOX_ACCESS_TOKEN_SEED", "")
     if access_seed and not re.fullmatch(r"[0-9a-fA-F]{64,}", access_seed):
         print("ERROR: SANDBOX_ACCESS_TOKEN_SEED must be hex-encoded, >= 32 bytes (64 hex chars)", file=sys.stderr)
--- a/.github/workflows/scripts/deploy-proxy.py
+++ b/.github/workflows/scripts/deploy-proxy.py
@@ -42,6 +42,12 @@ def main() -> int:
     if not re.fullmatch(r"[A-Za-z0-9.\-]+", proxy_domain):
         print("ERROR: PROXY_DOMAIN contains disallowed characters", file=sys.stderr)
         return 1
+    # Optional multi-domain set for DNS/hostname transitions; the proxy
+    # prefers PROXY_DOMAINS over PROXY_DOMAIN when both are present.
+    proxy_domains = os.environ.get("PROXY_DOMAINS", "")
+    if proxy_domains and not re.fullmatch(r"[A-Za-z0-9.,\- ]+", proxy_domains):
+        print("ERROR: PROXY_DOMAINS contains disallowed characters", file=sys.stderr)
+        return 1
     access_seed = os.environ.get("SANDBOX_ACCESS_TOKEN_SEED", "")
     if access_seed and not re.fullmatch(r"[0-9a-fA-F]{64,}", access_seed):
         print("ERROR: SANDBOX_ACCESS_TOKEN_SEED must be hex-encoded, >= 32 bytes (64 hex chars)", file=sys.stderr)
@@ -125,6 +131,7 @@ def main() -> int:
             sudo mkdir -p /etc/sandbox
             sudo tee /etc/sandbox/proxy.env > /dev/null <<PROXYENV
             PROXY_DOMAIN={proxy_domain}
+            PROXY_DOMAINS={proxy_domains}
             SANDBOX_ACCESS_TOKEN_SEED={access_seed}
             PROXY_ALLOWED_ORIGINS={terminal_origins}
             REQUIRE_DATA_PLANE={require_data_plane}
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -38,19 +38,19 @@ func main() {
 	addr := envOrDefault("PROXY_ADDR", ":5007")
 	redirectAddr := envOrDefault("PROXY_REDIRECT_ADDR", ":5008")
 	vmdAddr := envOrDefault("VMD_ADDR", "http://127.0.0.1:9090")
-	domain := envOrDefault("PROXY_DOMAIN", "sandbox.superserve.ai")
+	domains := proxyDomains()
 
 	log.Info().
 		Str("addr", addr).
 		Str("vmd_addr", vmdAddr).
-		Str("domain", domain).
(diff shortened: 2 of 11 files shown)


cmd/proxy/main.go
```diff
@@ -148,6 +153,18 @@ func main() {
...
 }
 
+// proxyDomains resolves the set of sandbox domains the proxy accepts.
+// PROXY_DOMAINS (comma-separated) takes precedence so one deploy can serve
+// both the legacy hostname and the region-prefixed hostname during a DNS
+// transition; otherwise the single-value PROXY_DOMAIN keeps existing
+// deploys unchanged.
+func proxyDomains() []string {
+	if domains := splitCSV(os.Getenv("PROXY_DOMAINS")); len(domains) > 0 {
+		return domains
```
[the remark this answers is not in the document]
pavitrabhalla: Good catch — and it's the exact drift class this env file has bitten us with before (a live edit surviving only until the next deploy rewrite). Fixed: the workflow now passes optional per-environment `PROXY_DOMAINS` repo vars and the deploy script validates and writes them into proxy.env alongside `PROXY_DOMAIN`. Unset vars write an empty value, which the proxy treats as absent (single-domain fallback), so environments without the var are untouched.

.github/workflows/scripts/deploy-proxy.py
```diff
@@ -42,6 +42,12 @@ def main() -> int:
     if not re.fullmatch(r"[A-Za-z0-9.\-]+", proxy_domain):
         print("ERROR: PROXY_DOMAIN contains disallowed characters", file=sys.stderr)
         return 1
+    # Optional multi-domain set for DNS/hostname transitions; the proxy
+    # prefers PROXY_DOMAINS over PROXY_DOMAIN when both are present.
+    proxy_domains = os.environ.get("PROXY_DOMAINS", "")
+    if proxy_domains and not re.fullmatch(r"[A-Za-z0-9.,\- ]+", proxy_domains):
```
meAmitPatil: This char class lets a space-separated list through. `PROXY_DOMAINS="sandbox.superserve.ai usw-sandbox.superserve.ai"` passes validation, but `splitCSV` on the Go side only splits on commas, so it comes out as a single domain string that can never match a Host — and since a non-empty `PROXY_DOMAINS` wins over `PROXY_DOMAIN`, the proxy comes up cleanly with routing fully broken. Nothing downstream catches it either: the post-deploy check is just `systemctl is-active`, and `/health` returns 200 on any host that *doesn't* match a configured suffix, so the LB stays green while every sandbox request 400s.

The whole point of validating here is to fail the deploy instead of the fleet, so I'd validate per entry with the same charset `PROXY_DOMAIN` uses:

```python
for d in filter(None, (s.strip() for s in proxy_domains.split(","))):
    if not re.fullmatch(r"[A-Za-z0-9.\-]+", d):
        print("ERROR: PROXY_DOMAINS contains an invalid domain", file=sys.stderr)
        return 1
```

That keeps the whitespace-around-commas tolerance while rejecting space-as-separator. Worth noting `PROXY_ALLOWED_ORIGINS` right below is also a comma list and its regex deliberately excludes spaces.

While you're in here: the env-var docstring at the top of this file doesn't mention `PROXY_DOMAINS` — that's the first place anyone looks when setting up a cell, and it currently says `PROXY_DOMAIN` is the knob.

cmd/proxy/main.go
```diff
@@ -148,6 +153,18 @@ func main() {
 	log.Info().Msg("proxy stopped")
 }
 
+// proxyDomains resolves the set of sandbox domains the proxy accepts.
+// PROXY_DOMAINS (comma-separated) takes precedence so one deploy can serve
+// both the legacy hostname and the region-prefixed hostname during a DNS
+// transition; otherwise the single-value PROXY_DOMAIN keeps existing
+// deploys unchanged.
+func proxyDomains() []string {
+	if domains := splitCSV(os.Getenv("PROXY_DOMAINS")); len(domains) > 0 {
```
meAmitPatil, line 157: Does the second variable need to exist? `splitCSV` of a single value is a one-element list, so `PROXY_DOMAIN` could just learn commas:

```go
func proxyDomains() []string {
	return splitCSV(envOrDefault("PROXY_DOMAIN", "sandbox.superserve.ai"))
}
```

That's fully backward compatible with every existing deploy, and it deletes the precedence rule, the second regex in the deploy script, the three new repo vars, and half the cases in `TestProxyDomains`. The deploy-side regex change is one character.

If you do keep `PROXY_DOMAINS` (I can see the argument for not touching the semantics of an existing var mid-transition), please add a startup warning when `PROXY_DOMAIN` is set but its value isn't in the effective list. The names are one letter apart and read as additive — someone setting the new prod var to just the region-prefixed hostname would silently drop the legacy domain on the next deploy, and the only trace is the startup info log. That's the exact failure mode this PR exists to prevent.

cmd/proxy/main.go
```diff
@@ -101,16 +101,21 @@ func main() {
 
 	// Health check for the GCP LB. Only responds on non-sandbox hosts
 	// so the boxd-label lockdown isn't bypassed.
-	domainSuffix := "." + domain
+	domainSuffixes := make([]string, len(domains))
```
meAmitPatil: This is now the second copy of "is this host one of ours", and the two already disagree: this one strips the port with `IndexByte` (mangles IPv6 literal hosts), while `ParseHost` uses `net.SplitHostPort`. The comment above says this gate exists so `/health` can't bypass the boxd-label lockdown — which only holds if the accept set here stays exactly in sync with the routing accept set, and that won't survive two hand-rolled copies. Next person to touch matching semantics in `internal/proxy` will not know this loop exists.

I'd export one predicate from the handler — `h.ServesHost(host string) bool` wrapping the port-strip + `matchesDomain` — and call it here. Deletes the precomputed slice too.

internal/proxy/proxy.go
```diff
@@ -54,7 +54,12 @@ const (
 
 // Handler is the core reverse proxy handler.
 type Handler struct {
-	domain       string // expected hostname suffix, e.g. "sandbox.superserve.ai"
+	// domains are the accepted hostname suffixes, e.g.
+	// ["sandbox.superserve.ai", "usw-sandbox.superserve.ai"]. All entries
+	// are valid for routing; the first is canonical should the proxy ever
```
meAmitPatil: Nit: the "first is canonical" sentence documents a contract that nothing enforces or tests, and readers will treat it as one — operators reordering the comma list have no idea order is supposed to matter. If the proxy ever needs a canonical domain, that's the moment to define and test it. Until then I'd keep this comment to what's true today (all entries are equivalent for validation).

meAmitPatil (commented):
The routing change itself looks right — `ParseHost`/`isSharedHost` generalize cleanly, per-domain semantics are genuinely unchanged, and the test coverage is good (the near-miss and `attacker.com` cases especially). My comments are all in the config plumbing rather than the routing code: one validation gap that can brick a cell while health checks stay green, and a design question about whether the second env var needs to exist at all.

One thing I noticed while tracing deploy paths, unrelated to this PR: `deploy/deploy-proxy.py` is stale — it writes `/etc/superserve/proxy.env` with `TERMINAL_ALLOWED_ORIGINS`, but the unit reads `/etc/sandbox/proxy.env` and the binary reads `PROXY_ALLOWED_ORIGINS`. Nothing references it anymore. Worth deleting in a follow-up so nobody reaches for it mid-transition.

.github/workflows/scripts/deploy-proxy.py
pavitrabhalla: Fixed — validating per-entry now with the same charset PROXY_DOMAIN uses, so a space-separated list fails the deploy instead of shipping broken routing. Also added PROXY_DOMAINS to the env-var docstring.

cmd/proxy/main.go
pavitrabhalla: Kept PROXY_DOMAINS as a separate var rather than teaching PROXY_DOMAIN commas — didn't want to change the semantics of an existing prod var mid-transition. Added the startup warning you suggested: logs when PROXY_DOMAIN is set but not in the effective domain list, so the one-letter typo/override case doesn't silently drop the legacy domain.

cmd/proxy/main.go
pavitrabhalla: Exported Handler.ServesHost wrapping the port-strip (via net.SplitHostPort, fixing the IPv6-literal mangling) + matchesDomain, and pointed the /health gate at it. One implementation now.

internal/proxy/proxy.go
pavitrabhalla: Done — reworded to state what's true today rather than implying an enforced contract.

meAmitPatil (approved):

PR #205, opened 2026-07-10
vmd: pair the presence side-car with the mem overlay across delete, freshen, and restore

meAmitPatil:
Companion to superserve-ai/firecracker#15, which makes diff-snapshot saves persist an explicit page-presence bitmap (`<mem>.presence`) next to the overlay and makes layered restores prefer it over inferring presence from the file's extent map. This wires the control plane to treat that side-car as part of the overlay pair:

- **`DeleteSnapshotFiles`** removes `mem.diff.presence` alongside `.base` — same rationale as the existing side-car removals: VMD reuses mem paths, and a stale bitmap next to a future same-size overlay passes Firecracker's geometry checks and silently resolves pages against the wrong layer.
- **`freshenFirstPassOverlay`** removes the presence side-car with the overlay it describes; failure falls back to a Full snapshot, matching the existing semantics.
- **Restore gate (`VMD_REQUIRE_PRESENCE_SIDECAR`, default off):** a layered restore of an overlay with no presence side-car is refused with an actionable error instead of letting Firecracker fall back to extent scanning. Intended for hosts that restore transferred artifacts, where a missing side-car more likely means "dropped in transit" than "legacy snapshot". When the gate is off, VMD logs a warning and restores as before.
- **`snapcheck -presence`:** compares two overlays as layered overlay/side-car *pairs* — presence bits page by page, bytes only where both sides provide the page. Raw byte comparison is structurally blind to the failure this catches (two byte-identical overlays whose side-cars disagree restore differently); this is the verification mode transfer validation should use. Includes a Go reference decoder for the side-car format, with the CRC-64 variant pinned to the snapshotter's by known-answer tests.

**Deploy ordering (REQUIRED):** superserve-ai/firecracker#15 merges and deploys to every host BEFORE this PR. A host running new vmd + pre-side-car Firecracker converges and goes strict, but its diff re-saves rewrite overlays without refreshing side-cars — stale bitmaps that the newer Firecracker would later trust. The in-code guard (`verifyPresenceRefreshed`) turns that misorder into loud warnings + removed side-cars rather than silent corruption, but the ordering is the actual contract. (Mirror of the downgrade note below.)

Ops notes (not in this PR): artifact transfer must copy the overlay/side-car pair atomically and only capture quiesced sandboxes; deploy tooling should delete `.presence` files when downgrading Firecracker across the side-car boundary, since older binaries re-save diffs without rewriting it.

Tests: side-car removal added to the delete/freshen tests; `overlayPresenceMissing` unit test; presence decode/sentinel/checksum tests plus a regression test where byte comparison passes and presence-aware comparison catches the flipped page; full `internal/vm` and `cmd/snapcheck` suites pass.

### DIFF
--- a/cmd/snapcheck/main.go
+++ b/cmd/snapcheck/main.go
@@ -16,6 +16,10 @@ import (
 func main() {
 	pageSize := flag.Int("page-size", 4096, "page size in bytes")
 	maxReport := flag.Int("max-report", 20, "max differing page offsets to print")
+	presence := flag.Bool("presence", false,
+		"compare as layered overlay/side-car pairs (<mem>.presence required next to each input): "+
+			"presence bits page by page, bytes only where both sides provide the page. "+
+			"Raw byte comparison is blind to presence — two byte-identical overlays can restore differently.")
 	flag.Usage = func() {
 		fmt.Fprintln(os.Stderr, "usage: snapcheck [flags] <mem-a> <mem-b>")
 		flag.PrintDefaults()
@@ -29,6 +33,32 @@ func main() {
 	if *pageSize <= 0 {
 		fmt.Fprintln(os.Stderr, "page-size must be positive")
 		os.Exit(2)
+	}
+
+	if *presence {
+		res, err := comparePresenceAware(flag.Arg(0), flag.Arg(1), *pageSize)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "snapcheck: %v\n", err)
+			os.Exit(2)
+		}
+		fmt.Printf("pages: %d  presence-differing: %d  content-differing: %d\n",
+			res.totalPages, len(res.presenceDiff), len(res.contentDiff))
+		report := func(label string, pages []int) {
+			for i, p := range pages {
+				if i >= *maxReport {
+					fmt.Printf("  ... %d more\n", len(pages)-*maxReport)
+					break
+				}
+				fmt.Printf("  %s page %d  offset 0x%x\n", label, p, int64(p)*int64(*pageSize))
+			}
+		}
+		report("presence", res.presenceDiff)
+		report("content", res.contentDiff)
+		if len(res.presenceDiff) > 0 || len(res.contentDiff) > 0 {
+			os.Exit(1)
+		}
+		fmt.Println("identical")
+		return
 	}
 
 	res, err := compareFiles(flag.Arg(0), flag.Arg(1), *pageSize)
--- a/cmd/snapcheck/main.go
+++ b/cmd/snapcheck/main.go
@@ -43,17 +43,8 @@ func main() {
 		}
 		fmt.Printf("pages: %d  presence-differing: %d  content-differing: %d\n",
 			res.totalPages, len(res.presenceDiff), len(res.contentDiff))
-		report := func(label string, pages []int) {
-			for i, p := range pages {
-				if i >= *maxReport {
-					fmt.Printf("  ... %d more\n", len(pages)-*maxReport)
-					break
-				}
-				fmt.Printf("  %s page %d  offset 0x%x\n", label, p, int64(p)*int64(*pageSize))
-			}
-		}
-		report("presence", res.presenceDiff)
-		report("content", res.contentDiff)
+		reportPages("presence ", res.presenceDiff, *pageSize, *maxReport)
+		reportPages("content ", res.contentDiff, *pageSize, *maxReport)
 		if len(res.presenceDiff) > 0 || len(res.contentDiff) > 0 {
 			os.Exit(1)
 		}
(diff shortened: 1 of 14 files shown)


pavitrabhalla (approved):

meAmitPatil:
Final design change, replacing the operator-timed flag with derived enforcement — a mistimed flag flip on a host with legacy layered overlays would have refused their resumes, which is an unacceptable footgun for prod:

**`internal/presence` (new):** the side-car format's single source of truth — encode/decode (CRC-pinned to the snapshotter by known-answer tests), the `SEEK_DATA`/`SEEK_HOLE` extent scan, and an atomic generator (temp + rename + dir fsync — vmd isn't under Firecracker's seccomp, so it gets the atomicity the fork's writer couldn't have). snapcheck now consumes this package instead of carrying its own copy of the format.

**Convergence sweep (reconciler pass):** pre-convergence, every quiesced layered overlay lacking a side-car gets one generated from its own extents — sound for exactly these files, since a pre-convergence side-car-less overlay is a legacy local artifact whose extents are the same evidence the warn-and-scan restore fallback already trusts. Dormant sandboxes whose owners never resume are healed in place, no user involvement. Once zero remain, a durable per-host marker is written and **the sweep never generates again**: post-marker, a side-car-less overlay is by definition suspect (a damaged or half-copied transfer), and healing it would launder the corruption the side-car exists to prevent. Torn-save/un-layerable sentinels are present files — never regenerated, their loud refusal preserved.

**Derived strictness:** `VMD_REQUIRE_PRESENCE_SIDECAR` is now tri-state — `auto` (default) enforces the gate exactly when the converged marker exists, so enforcement engages only when it provably affects nothing on the host; `always` for fresh migration-target hosts; `never` as break-glass. Legacy `true`/`false` map to `always`/`never`.

Net effect on rollout: deploy binary, deploy vmd, done — convergence and enforcement happen without an operator in the loop, and the config state that could break legacy resumes is no longer reachable in normal operation.

Tests: format round-trip/sentinels/corruption/atomicity in `internal/presence`; scan + generate against real sparse files (written-zero pages included); sweep lifecycle (defers busy VMs, converges, freezes post-marker, preserves sentinels, marker survives restart); strict-mode derivation table. All suites green on Linux.

pavitrabhalla:
Re-reviewed the full diff at d7bbab3 — the Codex rounds are all addressed properly and the sweep's provenance/race guards look right. One deploy-ordering hazard remains, the mirror image of the downgrade note in the description:

If this merges and deploys while the hosts still run pre-#15 Firecracker, the auto sweep converges the host (strict on), and any subsequent resume→re-pause cycle rewrites `mem.diff` without touching the sweep-generated side-car — old FC doesn't know the file exists, and PauseVM only removes it on the error path and first-pass freshen. The stale bitmap is symptom-free while old FC ignores side-cars on restore, then becomes silent wrong-layer page resolution the moment firecracker#15 deploys and restores start preferring it. On the use host that's under live customer traffic.

A code-side fix is awkward: removing the side-car before each diff save makes old-FC hosts accumulate missing side-cars post-convergence (the sweep early-returns once the marker exists), which strict mode then refuses — loud, but breaks resumes. So I'd pin the ordering instead: **firecracker#15 merges and deploys to both hosts before this PR merges.** Sweep generation over never-resaved legacy overlays is sound; the hazard only exists for overlays re-saved by old FC after the sweep ran. Optionally, a belt-and-suspenders guard — after a successful diff save, check the side-car's mtime advanced past the save start and remove-plus-log loudly if not — turns a misordered deploy into loud refusals instead of silent corruption.

Two questions: is VMD_INCREMENTAL_SNAPSHOT currently on for the use host (if off, no in-place diff re-saves happen today and the window is theoretical — but the ordering should still be pinned in the deploy plan)? And what's your ETA on the #15 fixes for the restore-dirty divergence and the huge-page boundary — that PR is now the head of the migration's critical path.

meAmitPatil:
Ordering agreed and now pinned in the description: firecracker#15 deploys everywhere before this merges. And your belt-and-suspenders suggestion is implemented in c761123 — after a successful layered diff save, vmd stats the side-car and removes it with a loud "fix deploy ordering" warning if its mtime didn't advance past the save start (i.e. the running Firecracker didn't write it). Misorder now degrades to warnings + strict refusals / pre-convergence scans instead of a newer Firecracker trusting a bitmap that doesn't describe the rewritten overlay. Unit-tested for the fresh/stale/missing cases.

On your questions:

**VMD_INCREMENTAL_SNAPSHOT on the use host:** on, per the pause-latency rollout record ("prod flags ALL ON" — RESUME_UFFD, INCREMENTAL_SNAPSHOT, HANDLER_DEATH_ABORT), and the west bring-up validated diff re-saves end-to-end. So the window is real, not theoretical — hence implementing the guard rather than relying on ordering alone. Worth re-confirming against the live vmd.env during the deploy itself.

**Fork PR ETA:** the huge-page boundary is already fixed and pushed there (6c70ebf8a — layered restore now requires exactly host-page granularity, closing the save/restore page-size split; awaiting your re-review). "Restore-dirty divergence" I can't map to anything currently open on that PR — if you mean the DirtyTracked state loss we observed on staging (vmd restart → reattached VM's next pause silently degrades to Full), that's vmd-side reattach state, not fork code, and I have it flagged as a follow-up ticket. Point me at the specific concern if it's something else and I'll fold it into the fork PR this week.

pavitrabhalla:
Field report from the first day of the sweep running in prod, one gap worth a follow-up: the provenance check misses locally-dumped overlays whose BoltDB records were lost (host reboot/reattach amnesia — we had 56 on the use host, 15 of them belonging to a customer, plus 1 on usw). They're genuinely local — extents trustworthy, exactly what the sweep exists to cover — but `sweepEligibilityOf` can't see them, so they converge as unknown-provenance orphans and strict mode then refuses their resumes. On use that was a live customer-facing refusal risk from the moment the marker landed at 02:38.

We healed them out-of-band (one-off tool doing Scan + WriteIfAbsent for ids the control-plane DB confirms as paused sandboxes on this host), but the fix belongs in the sweep: the reconciler already joins DB sandbox rows for this host (Source C in `runOnce`) — passing those ids to the sweep as a second provenance source alongside BoltDB would close the amnesia window structurally. A DB row that says paused-on-this-host is at least as authoritative as a BoltDB record.

Suggest also folding a sidecar census (`overlays without side-cars that have live DB rows`) into the drift checks so this class pages instead of hiding behind the converged marker.

PR #208, opened 2026-07-13
vmd: batch the recycle tap rebuild into one exec and bound its concurrency

meAmitPatil:
With `VMD_RECYCLE_TAP_RESET` enabled, a mass delete returns hundreds of slots at once and every recycle rebuilds its tap concurrently — five `ip netns exec` invocations each, all serializing on the kernel's netlink lock. Under that contention each rebuild blows its deadline, so every reset fails to the teardown fallback: safe, but it disables recycling under exactly the churn it exists for and emits an error per slot.

- Run the rebuild as a single `sh -c` exec (one netns entry instead of five fork pairs); interpolants are package constants.
- Bound in-flight rebuilds with a semaphore (8), acquired before the rebuild's deadline starts so queueing doesn't eat the budget.
- Add a test pinning the concurrency bound under a 32-slot mass return.

With this, a "tap reset failed" error is a rare, meaningful signal rather than mass-delete noise.

### DIFF
--- a/internal/network/manager.go
+++ b/internal/network/manager.go
@@ -420,19 +420,17 @@ func (m *Manager) setupSlot(ctx context.Context, idx int) (*VMNetInfo, string, e
 // decline to recycle. nftables rules match tap0 by name, so reusing the name
 // keeps them valid. Also the tap-construction path for setupSlot (the delete
 // is a no-op on a fresh namespace), keeping fresh and recycled taps identical.
+//
+// The rebuild runs as a single exec: per-command `ip netns exec` invocations
+// fork twice each and serialize on the kernel's netlink lock, which under a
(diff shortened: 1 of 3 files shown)


awille-ai (approved):
Security/scalability review complete. I did not find any critical or high-severity issues, so I’m approving.

The earlier concerns around orphaned child processes on timeout and unnecessary tap rebuilds after the recycle pool filled have both been addressed: the reset shell now runs in its own process group with group cancellation, and recycle capacity is rechecked after semaphore acquisition before spending time rebuilding a TAP that will be torn down anyway.

The remaining design tradeoffs (using `sh -c` for a fixed constant-only script, bounded concurrency of 8, and best-effort recycle under heavy contention) look reasonable. CI is green, and I don't see a remaining correctness, security, or scalability issue that would justify withholding approval.

PR #209, opened 2026-07-13
vmd: tighten boxd health probe interval cap to 10ms

meAmitPatil:
The boxd health-check probe interval ramps 1ms and doubles to a cap. With a 50ms cap, readiness gets quantized to a coarse grid: nearly every restore pays for a full extra tick waiting on a guest that is already up. Dropping the cap to 10ms bounds that dead time at 10ms while keeping probe cost negligible (each probe is a sub-millisecond HTTP GET over the local veth, and the handler does no work). Timeout and the dense 1ms ramp-up are unchanged; only steady-state probe frequency increases.

Also in this file:
- Disable keep-alives on the health-probe and init clients. Both defaulted to the shared transport with keep-alives on, so a connection parked by a probe could be reused after the VM's IP was recycled to a different VM — the exact hazard the exec client already disables keep-alives for. A stale reused connection on the init POST fails non-retryably.
- Surface the last probe error in the health-wait timeout, so connection-refused, no-response, and non-200 failures are distinguishable instead of collapsing into the same generic message.

### DIFF
--- a/internal/vm/vsock_exec.go
+++ b/internal/vm/vsock_exec.go
@@ -35,10 +35,14 @@ var boxdHTTPClient = &http.Client{
 func waitForHTTPHealth(ctx context.Context, vmIP string, timeout time.Duration) error {
 	url := fmt.Sprintf("http://%s:%d/health", vmIP, boxdPort)
 	deadline := time.Now().Add(timeout)
-	client := &http.Client{Timeout: 500 * time.Millisecond}
+	client := &http.Client{
+		Timeout:   500 * time.Millisecond,
+		Transport: &http.Transport{DisableKeepAlives: true},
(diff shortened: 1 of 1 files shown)


pavitrabhalla (approved):

awille-ai (approved):
Reviewed with the usual security/correctness/scalability bar.

This is a narrowly scoped tuning change that reduces the maximum health probe interval from 50ms to 10ms while leaving the exponential ramp-up and timeout behavior unchanged.

I did not find any critical or high-severity concerns:
- no change to correctness or failure semantics;
- no additional attack surface;
- the increased probe frequency is still very small for a local veth HTTP health check;
- CI is green.

Approving.

PR #210, opened 2026-07-13
controlplane: retry vmd Unavailable long enough to cover a daemon restart

meAmitPatil:
A vmd restart (e.g. during a deploy) refuses gRPC connections for several seconds, then answers Unavailable from its startup gate until it is serving. The existing service-config retry policy cannot cover that: gRPC hard-caps retry attempts at 5 with jittered backoff, so its worst-case envelope is a few seconds — creates and resumes that land in the restart window fail through to the client even though the daemon is seconds from healthy.

Replace the service-config policy with a client-side unary interceptor that retries Unavailable on a deterministic backoff (250ms doubling, capped at 2s) for up to 15s or until the caller's context expires. The dead-host interceptor is chained outside retry, preserving its view of the post-retry outcome only.

Details that matter:
- The interceptor calls ResetConnectBackoff between attempts: the channel's own reconnect backoff spaces dials out to multi-second gaps that can overshoot the retry window even when the daemon recovers inside it.
- StreamBuildLogs (not covered by a unary interceptor) gets an equivalent open-retry, gated on no event having been delivered yet — vmd replays buffered history on a fresh stream, so retrying after delivery would duplicate events.
- Exhausting the window logs a warning with the attempt count — the signal that the window no longer outlasts a restart.
- Unavailable almost always means the request never executed, but a connection severed mid-call also surfaces as Unavailable, so retried RPCs must stay idempotent (documented on the interceptor).

Trade-off: RPCs to a genuinely dead host now spend up to ~15s (bounded by the caller's context) before failing, instead of a few seconds.

### DIFF
--- a/cmd/controlplane/main.go
+++ b/cmd/controlplane/main.go
@@ -619,10 +619,9 @@ func (c *grpcVMDClient) CancelBuild(ctx context.Context, buildVMID string) error
 }
 
 func (c *grpcVMDClient) StreamBuildLogs(ctx context.Context, buildVMID string, onEvent func(vmdclient.BuildLogEvent) error) error {
-	// The unary retry interceptor doesn't cover streams, so ride out a
-	// daemon restart here — but only while no event has been delivered:
-	// vmd replays buffered history on a fresh stream, so retrying after
-	// delivery would duplicate events to the caller.
+	// The unary retry interceptor doesn't cover streams. Retry the open
+	// only while nothing was delivered — a fresh stream replays history,
+	// which would duplicate events.
 	deadline := time.Now().Add(retryUnavailableWindow)
 	backoff := 250 * time.Millisecond
 	for {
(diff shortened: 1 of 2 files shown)


meAmitPatil:
Deliberate: gating by method would reintroduce the deploy-window failures this PR removes, and the previous service-config policy already retried Unavailable service-wide with the same semantics — the exposure isn't new, and clean shutdowns drain in-flight RPCs before severance. The root fix is server-side idempotency for the pause/restore handlers (so retry-after-execution is safe from any client, not just this interceptor); that lands in a follow-up PR.

cmd/controlplane/main.go
```diff
@@ -645,10 +663,49 @@ func (c *grpcVMDClient) StreamBuildLogs(ctx context.Context, buildVMID string, o
...
+			// this window; reset it so re-dials follow this loop's
+			// cadence. Nil cc only in tests.
+			if cc != nil {
+				cc.ResetConnectBackoff()
+			}
+			select {
+			case <-ctx.Done():
+				// Stale Unavailable, not ctx.Err(): the dead-host
+				// interceptor above keys off it.
+				return err
```
pavitrabhalla: On ctx cancellation mid-retry this returns the last Unavailable error instead of ctx.Err(), so deadHostUnaryInterceptor treats it as a dead host. Any caller with a timeout shorter than the 15s window — exactly the population racing a real restart — will now evict the host's cached client (hostreg.Invalidate -> DB lookup + redial) purely because it gave up, not because vmd is unhealthy. Is that trade-off intentional, or should this return ctx.Err() when it's actually the context that expired rather than a genuine Unavailable?

cmd/controlplane/main.go
```diff
@@ -626,17 +619,42 @@ func (c *grpcVMDClient) CancelBuild(ctx context.Context, buildVMID string) error
 }
 
 func (c *grpcVMDClient) StreamBuildLogs(ctx context.Context, buildVMID string, onEvent func(vmdclient.BuildLogEvent) error) error {
```
pavitrabhalla, line 622: retryUnavailableUnaryInterceptor has solid tests (retry-then-succeed, non-retryable codes, window exhaustion, ctx cancel) but this stream retry loop — including the delivered gate that avoids replaying buffered history — has none. It's the trickier half of this change (partial-delivery dedup); worth a test or two before merge.

meAmitPatil:
Both addressed in ebca6a6:

1. **ctx cancel now returns ctx.Err()** (wrapping the last attempt's error for context) instead of the stale Unavailable — you're right that a caller giving up shouldn't read as a dead host. Callers that outlive the window still exhaust it with a genuine Unavailable, so dead-host eviction is unaffected where it matters.

2. **Stream retry tests added** (open-retry-until-success, no-retry-after-delivery, no-retry-on-other-codes) — and the no-retry-after-delivery test immediately caught a real bug: `delivered` was only set on the callback-error and Finished paths, so a partial stream that died mid-Recv would have been retried and replayed history. Fixed. Good call.

pavitrabhalla (approved):

PR #211, opened 2026-07-13
vmd: pause/resume lifecycle correctness under retries and interrupted stops

meAmitPatil:
Two related classes of lifecycle bugs, one theme: an interrupted pause or a retried RPC must never condemn a healthy sandbox, roll back guest state, or leave a stale firecracker behind.

Retried RPCs (response lost to a daemon restart or connection reset):
- A retried pause of an already-paused VM returns the recorded snapshot artifacts (after confirming they still exist on disk). Previously it dialed the dead firecracker socket, failed, and the error handler deleted the VM's record.
- A retried resume or restore that finds the VM already running — same artifacts requested, boxd healthy, instance still tracked after the probe — returns the live instance instead of relaunching. Both paths share one guard (retriedLaunchTarget); previously a duplicate resume killed the healthy VM (the launch path removes the socket file first, so even the old no-op start ended in a timeout and an error-path stop).

Interrupted stops:
- Launch uses systemctl restart instead of start: an already-active unit — a stale firecracker left by an interrupted stop — is replaced inline instead of silently no-op'ing and stranding the launch on a socket the old process never re-binds.
- If the socket misses its 5s window while systemd still has a job in flight for the unit (a restart tearing down the old process, up to TimeoutStopSec), the wait extends once, capped by the caller's remaining deadline. Gating on the Job property means settled failures (exit-code, start-limit-hit) still fail fast.
- A pause whose stop fails still completes as paused — the snapshot is already valid, and failing the pause would leave the guest frozen-but-active and make a retried pause re-diff an already-consumed dirty bitmap (destroying the overlay). The stop is retried on a detached context and escalated to an error log.
- New reconciler drift rule: a systemd unit still active for a DB-paused sandbox is stopped after the grace period — reclaiming the guest RAM an interrupted stop would otherwise pin for as long as the sandbox sleeps. The record stays; the snapshot is valid.
- The socket-timeout launch failure now logs daemon-side with the VM id and unit state.

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -256,6 +256,39 @@ type Manager struct {
 	// until process exit so late pollers can read terminal outcomes.
 	buildsMu sync.RWMutex
 	builds   map[string]*buildRecord
+
+	// vmOpLocks serializes lifecycle operations (restore, resume, pause) for
+	// a single vmID, so a retry or a concurrent actor can't stomp an
+	// in-flight launch (kill a still-booting VM, re-diff a consumed dirty
+	// bitmap). Keyed by vmID → *sync.Mutex. Entries are never deleted:
+	// vmIDs are unique UUIDs so they don't collide, and deleting on destroy
+	// would let a same-vmID retry acquire a fresh mutex and lose mutual
+	// exclusion. DestroyVM deliberately does NOT take these locks — it must
+	// interrupt a wedged op (a hung firecracker socket call holds the lock
+	// until destroy SIGKILLs the process), so blocking it would turn a
+	// recoverable wedge into a permanent hang.
+	vmOpLocks sync.Map
+}
+
+// lockVMOp blocks until it holds vmID's lifecycle lock; the returned func
+// releases it. Never called from a read/reattach path (the mutex is not
+// reentrant) or from DestroyVM (see vmOpLocks).
+func (m *Manager) lockVMOp(vmID string) func() {
+	mu, _ := m.vmOpLocks.LoadOrStore(vmID, &sync.Mutex{})
+	l := mu.(*sync.Mutex)
+	l.Lock()
+	return l.Unlock
+}
+
+// tryLockVMOp acquires vmID's lifecycle lock without blocking. ok=false
+// means an operation is in flight for this vmID and the caller must not act.
+func (m *Manager) tryLockVMOp(vmID string) (unlock func(), ok bool) {
+	mu, _ := m.vmOpLocks.LoadOrStore(vmID, &sync.Mutex{})
+	l := mu.(*sync.Mutex)
+	if !l.TryLock() {
+		return nil, false
+	}
+	return l.Unlock, true
 }
 
 // NewManager creates a new VM manager.
@@ -643,6 +676,12 @@ func (m *Manager) DestroyVM(ctx context.Context, vmID string, force bool) error 
 
 // PauseVM snapshots the VM state and then stops the process.
 func (m *Manager) PauseVM(ctx context.Context, vmID, snapshotDir string) (snapshotPath, memPath string, err error) {
+	// Serialize with any other launch/pause for this vmID (see lockVMOp): a
+	// duplicate pause must wait and then hit the already-paused guard, not
+	// re-diff a dirty bitmap the first pass already consumed.
+	unlockOp := m.lockVMOp(vmID)
+	defer unlockOp()
+
 	inst, err := m.getInstance(vmID)
 	if err != nil {
 		return "", "", err
@@ -775,20 +814,25 @@ func (m *Manager) PauseVM(ctx context.Context, vmID, snapshotDir string) (snapsh
 		}
 	}
 
-	// Stop the Firecracker process — snapshot is already on disk. Detached
-	// ctx: the request budget may be spent after a slow snapshot. A stop
+	// Stop the Firecracker process — snapshot is already on disk. A stop
 	// that still fails must NOT fail the pause: the artifacts are valid and
 	// the record must reach Paused — a retry against a Running record would
 	// re-diff an already-consumed dirty bitmap and destroy the overlay. The
 	// straggler unit is replaced at the next launch and reclaimed by the
 	// reconciler meanwhile.
 	unit := systemdUnitName(vmID)
-	stopCtx, stopCancel := context.WithTimeout(context.WithoutCancel(ctx), 15*time.Second)
-	defer stopCancel()
-	if err := stopUnit(stopCtx, unit); err != nil {
+	if err := stopUnitWithBudget(ctx, unit); err != nil {
 		log.Warn().Err(err).Msg("systemctl stop failed during pause; retrying")
-		if serr := stopUnit(stopCtx, unit); serr != nil && !unitDefinitelyDead(stopCtx, unit) {
-			log.Error().Err(serr).Msg("unit still running after pause; reconciler will reclaim it")
+		// Fresh budget for the retry — sharing the first attempt's would
+		// leave the retry (and the dead-check) no time under host I/O
+		// contention, exactly when the retry is needed.
+		if serr := stopUnitWithBudget(ctx, unit); serr != nil {
+			deadCtx, deadCancel := context.WithTimeout(context.WithoutCancel(ctx), 3*time.Second)
+			dead := unitDefinitelyDead(deadCtx, unit)
+			deadCancel()
+			if !dead {
+				log.Error().Err(serr).Msg("unit still running after pause; reconciler will reclaim it")
+			}
 		}
 	}
 
@@ -915,6 +959,12 @@ func fileExists(path string) bool {
 func (m *Manager) ResumeVM(ctx context.Context, vmID, snapshotPath, memPath string) (*VMInstance, error) {
 	log := m.log.With().Str("vm_id", vmID).Logger()
 
+	// Serialize with any other launch/pause for this vmID (see lockVMOp): a
+	// duplicate resume must wait and then be recognized as a retry, not
+	// relaunch over a VM the first attempt is still booting.
+	unlockOp := m.lockVMOp(vmID)
+	defer unlockOp()
+
 	inst, err := m.getInstance(vmID)
 	if err != nil {
 		return nil, err
@@ -934,7 +984,7 @@ func (m *Manager) ResumeVM(ctx context.Context, vmID, snapshotPath, memPath stri
 	// the prior attempt. Relaunching would kill it and roll the guest back
 	// to the snapshot, so return the live instance — same guard as the
 	// stateless restore path.
-	if existing := m.retriedLaunchTarget(ctx, vmID, snapshotPath, memPath); existing != nil {
+	if existing := m.retriedLaunchTarget(vmID, snapshotPath, memPath); existing != nil {
 		log.Info().Msg("resume: VM already running and healthy, returning it")
 		return existing, nil
 	}
@@ -1122,6 +1172,12 @@ func (m *Manager) restoreForResume(socketPath, snapshotPath, memPath, basePath s
 // snapcheck). Non-destructive: stops the throwaway FC, leaving the sandbox paused.
 func (m *Manager) VerifySnapshot(ctx context.Context, vmID string) (string, error) {
 	log := m.log.With().Str("vm_id", vmID).Logger()
+
+	// Serialize with launch/pause for this vmID (see lockVMOp): the
+	// throwaway firecracker this starts must not race a concurrent resume
+	// onto the same unit.
+	unlockOp := m.lockVMOp(vmID)
+	defer unlockOp()
 
 	inst, err := m.getInstance(vmID)
 	if err != nil {
@@ -1391,6 +1447,13 @@ func (m *Manager) restoreVMSnapshot(ctx context.Context, vmID, snapshotPath, mem
 		vmID = uuid.New().String()
 	}
 
+	// Serialize with any other launch/pause for this vmID, so a retry lands
+	// AFTER an in-flight attempt finishes (and is then recognized as a retry
+	// by retriedLaunchTarget below) instead of racing into the inPlace block
+	// and stopping a still-booting VM.
+	unlockOp := m.lockVMOp(vmID)
+	defer unlockOp()
+
 	// Deterministic artifact/config preconditions, checked before ANY state
 	// changes: no provisional instance published (a refusal must not leave a
 	// phantom StatusError record for retries to trip over as inPlace), no
@@ -1420,7 +1483,7 @@ func (m *Manager) restoreVMSnapshot(ctx context.Context, vmID, snapshotPath, mem
 	// Retried restore (response lost mid-RPC): re-restoring a VM the prior
 	// attempt brought up would roll the guest back to the snapshot. Healthy
 	// boxd proves the prior restore completed — return that VM instead.
-	if existing := m.retriedLaunchTarget(ctx, vmID, snapshotPath, memPath); existing != nil {
+	if existing := m.retriedLaunchTarget(vmID, snapshotPath, memPath); existing != nil {
 		log.Info().Msg("restore: VM already running and healthy, returning it")
 		return existing, nil
 	}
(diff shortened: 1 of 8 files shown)


meAmitPatil:
Both Codex findings addressed:

1. **Cancelled-probe fall-through** — fixed in 74f0621: the guard (now shared as `retriedLaunchTarget`) probes on `context.WithoutCancel(ctx)` with its own 3s bound, so a spent retry deadline can't false-negative a healthy VM into replacement. A probe that genuinely fails for 3s still falls through by design — replacing a VM whose boxd is truly unresponsive is the recovery path for wedged guests.

2. **Missing test fixtures** — fixed in 9ecb1e1: the pause-guard test creates real artifact files, plus a new negative test pinning FailedPrecondition for dangling artifacts.

meAmitPatil:
Real — fixed in 41b753b. With Type=simple the restart job clears at fork, before the socket exists, so no at-timeout probe (Job, SubState, or otherwise) can distinguish a just-replaced unit from a stalled fresh one — the stop phase can eat the whole 5s and the error path then kills a fresh firecracker milliseconds from binding. Moved the decision to where the information actually exists: one is-active check before the restart sizes the socket budget up front (5s normal, +12s capped by the caller's deadline when replacing a live unit). The Job-property helper is gone; the timeout-instant heuristic was unsalvageable.

meAmitPatil:
Both real — fixed in c8a5764:

1. **Deactivating units**: the pre-restart check now uses ActiveState (active | activating | deactivating) instead of is-active, so a resume that arrives while an interrupted pause's stop job is still winding the old process down gets the extended socket budget too. Same single systemctl exec.

2. **DB-down fallback**: Drift 5 now keys off the BoltDB record's paused status when the DB row set is unavailable — the fallback path already loads full records, so a failed pause stop is reclaimed even through a DB outage.

meAmitPatil:
Right — and it's a sharper race than the fallback was worth: a drift marker accumulated by a genuinely stale unit survives into a subsequent resume's launch window (the BoltDB record stays Paused until ResumeVM persists at the end), so condition-true + grace-already-elapsed can stop a just-launched resume. The DB row can't hit this (mid-resume reads 'resuming', which clears the drift), and BoltDB has no state distinguishing the two. Reverted the fallback arm in b87c823: Drift 5 is DB-rows-only, comment records why. During a DB outage the reclaim just waits — a wrongly-killed customer resume is worse than delayed RAM reclaim.

meAmitPatil:
Right in principle, but the flow it protects doesn't exist: there is no reset-to-snapshot feature, and the control plane's status CAS means a RestoreSnapshot for a running same-artifact VM has exactly one meaning today — a retry whose response was lost. Treating that as a rollback request was the confirmed silent-data-loss bug this guard fixes. A request-scoped idempotency token is the clean general solution but means a proto change on both sides; disproportionate while no legitimate rollback caller exists. Encoded the constraint in 961463d: the guard's doc comment now states that a deliberate reset-to-snapshot flow must not reuse these RPCs without an explicit bypass signal.

meAmitPatil:
Validated end-to-end on staging (branch build):
- 20-cycle create/pause/resume/destroy smoke: 20/20, latencies at baseline.
- Manufactured the stale-unit state (paused sandbox + manually-started firecracker unit) and confirmed both defenses: resume replaced the stale unit inline and the sandbox exec'd cleanly (the failure signature this fixes), and the reconciler's paused-unit rule reclaimed a second untouched one after its grace period, leaving the record resumable.
- 20-min soak: 0 errors, 0 'firecracker socket missing after launch', 22 restores + 22 pauses clean, no leaked units.

internal/vm/manager.go
```diff
@@ -1380,6 +1417,14 @@ func (m *Manager) restoreVMSnapshot(ctx context.Context, vmID, snapshotPath, mem
 	// new VM (not in BoltDB).
 	m.lazyReattach(vmID)
 
+	// Retried restore (response lost mid-RPC): re-restoring a VM the prior
+	// attempt brought up would roll the guest back to the snapshot. Healthy
+	// boxd proves the prior restore completed — return that VM instead.
+	if existing := m.retriedLaunchTarget(ctx, vmID, snapshotPath, memPath); existing != nil {
```
pavitrabhalla: retriedLaunchTarget only matches Status == StatusRunning. Between record creation and the Status=StatusRunning assignment after the 30s waitForBoxd below, this record sits as StatusCreating. A retry landing in that window skips this guard and falls into the inPlace block right after, which unconditionally deletes the record and stops the unit — killing the still-in-flight first restore. Should this guard also short-circuit on StatusCreating, or otherwise stop inPlace from treating an in-flight attempt as garbage?

internal/vm/manager.go
```diff
@@ -2746,3 +2814,40 @@ func waitForSocketPolling(path string, timeout time.Duration) error {
...
+		return nil
+	}
+	existing.mu.RLock()
+	running, ip := existing.Status == StatusRunning, existing.IP
+	sameArtifacts := existing.SnapshotPath == snapshotPath && existing.MemFilePath == memPath
+	existing.mu.RUnlock()
+	if !running || !sameArtifacts || ip == "" {
+		return nil
+	}
+	if probeBoxdHealth(context.WithoutCancel(ctx), ip, 3*time.Second) != nil {
```
pavitrabhalla: This probe only waits 3s, but ResumeVM sets Status=StatusRunning before boxd is confirmed reachable — the real boxd check right after is a detached 30s best-effort probe. A retry landing in that 3–30s warmup window gets a false negative here, falls through to startFirecrackerViaSystemd, and restartUnit kills the still-booting VM — the exact rollback this guard exists to prevent. Can this share the same 30s tolerance the rest of Resume/Restore already gives boxd readiness?

internal/vm/manager.go
```diff
@@ -759,9 +775,21 @@ func (m *Manager) PauseVM(ctx context.Context, vmID, snapshotDir string) (snapsh
...
-		log.Warn().Err(err).Msg("systemctl stop failed during pause")
+	// Stop the Firecracker process — snapshot is already on disk. Detached
+	// ctx: the request budget may be spent after a slow snapshot. A stop
+	// that still fails must NOT fail the pause: the artifacts are valid and
+	// the record must reach Paused — a retry against a Running record would
+	// re-diff an already-consumed dirty bitmap and destroy the overlay. The
+	// straggler unit is replaced at the next launch and reclaimed by the
+	// reconciler meanwhile.
+	unit := systemdUnitName(vmID)
+	stopCtx, stopCancel := context.WithTimeout(context.WithoutCancel(ctx), 15*time.Second)
```
pavitrabhalla: stopUnit blocks on systemctl stop up to TimeoutStopSec=10, and pool.go's defaultVerifyMaxWait comment says this same timeout needs ~20s margin under host I/O contention. Here the retry stop + unitDefinitelyDead all share one 15s budget. If the first stop eats most of that budget under contention, the retry and dead-check both silently no-op (unitDefinitelyDead returns false immediately on an expired ctx) — right when the retry is needed. Worth giving the retry its own budget instead of sharing the first attempt's remainder?

internal/vm/manager.go
```diff
@@ -2621,14 +2668,35 @@ func (m *Manager) startFirecrackerViaSystemd(ctx context.Context, vmID, socketPa
 		return 0, fmt.Errorf("write start script: %w", err)
 	}
 
+	// Replacing a live (or still winding down) stale unit spends the stop
+	// phase (TimeoutStopSec 10s) before the fresh process can even fork, so
+	// it needs a bigger socket budget. Decided up front: with Type=simple
+	// the restart job clears at fork, before the socket exists, so no
+	// at-timeout probe can tell a just-replaced unit from a stalled fresh
+	// one.
+	replacingLive := unitLingering(ctx, systemdUnitName(vmID))
```
pavitrabhalla, line 2787: unitLingering and restartUnit are two separate systemctl calls with no re-check between them and no per-vmID lock. The reconciler's new Drift 5 (and a duplicate Resume/Restore retry) can flip the unit's state in that gap, so replacingLive can be stale by the time restartUnit actually runs — sizing socketWait too short, causing a spurious waitForSocket timeout, which then calls stopUnit on what might be a legitimately-launching process from the other concurrent actor. Is there a lock we should hold around this pair, or should the state be re-checked after restartUnit before deciding the wait was too short?

internal/vm/manager.go
```diff
@@ -2746,3 +2814,40 @@ func waitForSocketPolling(path string, timeout time.Duration) error {
...
+	sameArtifacts := existing.SnapshotPath == snapshotPath && existing.MemFilePath == memPath
+	existing.mu.RUnlock()
+	if !running || !sameArtifacts || ip == "" {
+		return nil
+	}
+	if probeBoxdHealth(context.WithoutCancel(ctx), ip, 3*time.Second) != nil {
+		return nil
+	}
+	m.mu.RLock()
+	still := m.vms[vmID] == existing
```
pavitrabhalla, line 2977: Non-blocking: this recheck only confirms the map entry wasn't swapped for a different pointer — it doesn't re-read Status/SnapshotPath/MemFilePath. PauseVM mutates the same *VMInstance in place rather than swapping the pointer, so a concurrent pause completing during the 3s probe above leaves this check passing while existing.Status is now Paused with different artifacts. Narrow window, but worth re-validating Status/artifacts here too?

internal/vm/reconciler.go
```diff
@@ -348,6 +348,36 @@ func (r *Reconciler) runOnce(ctx context.Context) {
...
 	}
 
+	// Drift 5: systemd unit active, DB says paused — an interrupted pause
+	// stop left the old firecracker pinning guest RAM; stop it. DB rows
+	// only: a mid-resume sandbox reads 'resuming' there, clearing the
+	// drift, while its BoltDB record stays Paused until the resume
+	// persists — keying off records could kill a just-launched resume.
+	if dbSandboxes != nil {
+		for id := range active {
+			sb, known := dbSandboxes[id]
```
pavitrabhalla, line 358: Non-blocking: dbSandboxes is a single snapshot taken at the top of runOnce, but this loop calls stopUnit later without re-checking DB status, and stopUnit has no generation/PID check. If a resume claims and relaunches this exact stale unit after the snapshot but before this id's turn in the loop (bounded by the 25s runTimeout), this could kill the freshly-relaunched process. Since this rule specifically targets leftover units from interrupted pauses, is a per-id status recheck right before stopUnit worth the extra query?

internal/vm/reconciler.go
```diff
@@ -348,6 +348,36 @@ func (r *Reconciler) runOnce(ctx context.Context) {
 		}
 	}
 
+	// Drift 5: systemd unit active, DB says paused — an interrupted pause
```
pavitrabhalla: Nit: this is labeled Drift 5, same as the existing BoltDB-orphan block further down (~line 414). Two blocks sharing a number will make "Drift 5" ambiguous in logs/comments — might be worth renumbering.

pavitrabhalla (changes requested):
Idempotency guards here each define "this is a retry" more narrowly than what the surrounding code already tolerates elsewhere (30s boxd warmup, in-flight StatusCreating restores, systemd's own async transitions) — most of the blocking comments below are variations on that same gap. Also flagged two narrower/lower-probability races as non-blocking, plus a cosmetic label collision.

awille-ai (changes requested):
Requesting changes for two high-severity retry/idempotency races that can still kill a healthy first attempt:

1. `retriedLaunchTarget` only recognizes `StatusRunning`. During restore, the tracked instance remains `StatusCreating` until the later boxd readiness phase completes. A duplicate restore arriving in that window skips the retry guard and falls into the in-place replacement path, deleting the record/stopping the unit for the still-running first restore. The retry path needs to recognize and safely join/re

meAmitPatil:
@pavitrabhalla @awille-ai — addressed in b5afbca (you both reviewed 961463d, one commit earlier). All the blocking comments were the same theme: the guards defined "retry" too narrowly for the in-between states around them. Fixed by serializing per-vmID lifecycle ops with a lock instead of widening each guard:

- **StatusCreating / boxd-warmup stomps (①②):** restore/resume/verify hold a per-vmID lock across the whole attempt, so a duplicate waits and is then recognized as a retry — it can't reach the inPlace block or false-negative on a still-booting VM. The boxd probe is gone (the lock makes StatusRunning trustworthy).
- **unitLingering→restartUnit TOCTOU (④):** both run under the lock; reconciler Drift 5 now TryLock-skips instead of blocking or stomping.
- **shared 15s stop budget (③):** each stop attempt gets its own; dead-check gets a fresh ctx.
- Non-blocking ⑤⑥ moot under the lock; ⑦ renumbered to Drift 6.

DestroyVM deliberately bypasses the lock (it's the only rescue for a wedged op), and reads/proxy stay off it — safety analysis in the commit. Re-validating the staging landmine tests on the new commit; will re-request review after.

awille-ai (changes requested):
Re-reviewed the current head and verified the two blockers from my previous review:

- A duplicate restore can no longer race a `StatusCreating` restore and kill it: same-VM lifecycle operations are serialized with `vmOpLocks`.
- The 3s boxd-probe false negative is gone: completed retries are recognized under the lifecycle lock and use process-level unit liveness, so a warming VM is not restarted.

There are still two blocking correctness issues
Cancellation race after acquiring the lifecycle lock
lockVMOp() honors cancellation while waiting, but there is still a small race where both the lock becomes available and the context is already canceled.
In Go, either select branch may win.
That can allow an already-abandoned pause/restore/resume to continue executing.
I recommended immediately checking ctx.Err() after acquiring the lock, releasing it, and returning if cancellation has already occurred.
Completed restore retries still consume global restore capacity
After a duplicate restore waits on the per-VM lock and the original restore completes, it still acquires restoreSem before recognizing it's a no-op retry.
A retry that should immediately return the existing VM can therefore block behind unrelated restores or even fail due to semaphore pressure.
Since it is logically a completed retry, it shouldn't require scarce restore capacity.

meAmitPatil:
The stale socket is removed before the restart — startFirecrackerViaSystemd does `os.Remove(socketPath)` (unconditional, same function, before restartUnit), and firecracker never recreates an unlinked socket. So the path is empty when waitForSocket runs: it waits for the FRESH process to bind, not the old one. That's already the "wait for the stale socket to disappear/reappear" this suggests — os.Remove is the disappear, waitForSocket the reappear. The residual bind-vs-listen readiness (socket exists at bind, connectable at listen) is pre-existing and handled by the separate socket-readiness change, not this PR.

internal/vm/manager.go
```diff
@@ -256,6 +256,77 @@ type Manager struct {
...
+	// in-flight launch (kill a still-booting VM, re-diff a consumed dirty
+	// bitmap). Keyed by vmID → chan struct{}{} used as a capacity-1
+	// semaphore, so acquisition can honor the caller's context. Entries are
+	// never deleted: vmIDs are unique UUIDs so they don't collide, and
+	// deleting on destroy would let a same-vmID retry acquire a fresh lock
+	// and lose mutual exclusion. DestroyVM deliberately does NOT take these
+	// locks — it must interrupt a wedged op (a hung firecracker socket call
+	// holds the lock until destroy SIGKILLs the process), so blocking it
+	// would turn a recoverable wedge into a permanent hang.
+	vmOpLocks sync.Map
```
pavitrabhalla, line 271: Non-blocking: vmOpLocks entries are never deleted, so this grows by one entry per vmID for the life of the process. Understood why (deleting on destroy would let a same-vmID retry lose mutual exclusion), but on a sandbox platform with real VM churn this is unbounded over a long-running vmd. Is this bounded some other way (periodic restarts?) or worth a reap/cap strategy down the line?

internal/vm/reconciler.go
```diff
@@ -348,6 +348,62 @@ func (r *Reconciler) runOnce(ctx context.Context) {
 		}
 	}
 
+	// Drift 6: systemd unit active, DB says paused — an interrupted pause
```
pavitrabhalla: Non-blocking: this is now "Drift 6", but there's already a "Drift 6 (detect-only)" block further down for disk orphans — same collision as before, just with the next number. Might be worth a pass to make the numbering unique.

pavitrabhalla (approved):
The vmOpLocks serialization closes all five blocking issues from the last round at the root rather than patching each one — retriedLaunchTarget no longer needs a boxd probe since Status is trustworthy under the lock, restoreVMSnapshot's in-flight window and the unitLingering/restartUnit race are both closed by the same lock, and the reconciler now rechecks the live record instead of trusting a stale snapshot before stopping a unit. pauseWithRetry closes the loop on the pause-stop-budget concern too. Two non-blocking items below, otherwise good to merge.

meAmitPatil:
**vmOpLocks growth:** bounded in practice by vmd restarts — every deploy touching the vmd paths restarts the process, and those are frequent, so the map resets regularly. Each entry is tiny (a channel + map slot, ~tens of bytes), so even millions of creates between restarts is a few MB. Agreed it's worth a reap/cap if vmd ever runs for weeks; a safe reaper would delete an entry only when m.vms[vmID] is absent AND the lock is free, to avoid the lost-mutual-exclusion race. Tracking as a down-the-line follow-up.

**Drift numbering:** fixed in 3778494 — the paused-unit rule is now Drift 7, so it no longer collides with the disk-orphan "Drift 6 (detect-only)".

awille-ai (approved):

PR #212, opened 2026-07-14
vmd: persistent systemd D-Bus connection for unit operations

meAmitPatil:
Every unit operation (start, stop, state reads, unit listing, MainPID) forked a `systemctl` process. Each fork pays exec + a D-Bus connect/handshake whose cost grows with the number of loaded units, and under a concurrent launch burst the spawns also compete with systemd and the starting units for the scheduler — the launch enqueue alone inflated ~4x at 100-way concurrency.

Replace the transport: one persistent connection to systemd's private socket (root, no dbus-daemon), with every helper falling back to exec'd `systemctl` per call when the connection is unavailable — worst case is exactly the previous behavior.

Design points:
- Fire-and-forget starts (`RestartUnit/StartUnit` with a nil result channel) return at job-enqueue, matching `--no-block`, and skip the library's job-tracking lock so concurrent launches pipeline on the socket.
- Blocking stops use a buffered result channel — the library's dispatcher does one blocking send, so an abandoned wait cannot wedge the connection.
- A closed connection (systemd re-exec kills the socket; the library never reconnects) is redialed once per call; a second failure falls back to exec.
- D-Bus method errors (NoSuchUnit, start-limit) surface to callers as before; only transport-shaped failures trigger the exec fallback.
- Gated by `VMD_SYSTEMD_DBUS` (default off) for a staged rollout.

### DIFF
--- a/internal/vm/sdbus.go
+++ b/internal/vm/sdbus.go
@@ -14,10 +14,10 @@ package vm
 //     ends, so a per-request context would kill it seconds after birth.
 //   - No caller ever blocks on connection state. Acquisition is a
 //     non-blocking check; if there is no healthy connection, one background
-//     dial is kicked off (single-flight, bounded by its own timer since
-//     dial+auth I/O ignores context deadlines) and the caller falls back to
-//     exec immediately. A wedged PID1 can therefore never stall helpers
-//     behind a mutex.
+//     dial is kicked off (single-flight, held until it truly returns) and
+//     the caller falls back to exec immediately. A wedged PID1 can
+//     therefore never stall helpers behind a mutex, and never costs more
+//     than one blocked dial goroutine.
 //   - Connected() is checked on every acquisition: the library's Conn is
 //     two sockets, and the signal socket can die alone, which would
 //     otherwise strand every job-completion wait while method calls keep
@@ -49,10 +49,6 @@ func SetSystemdDBusEnabled(v bool) {
 		sdbusAcquire()
 	}
 }
-
-// sdbusDialTimeout bounds a background dial attempt; an attempt that
-// overruns is abandoned (and closed if it ever completes).
-const sdbusDialTimeout = 5 * time.Second
 
 // sdbusRedialCooldown is the minimum gap between dial attempts, so a host
 // where systemd is unreachable doesn't burn a dial per helper call.
@@ -86,42 +82,23 @@ func sdbusAcquire() *sddbus.Conn {
 }
 
 // sdbusDial runs one background dial. The connection context is detached —
-// its lifetime is the daemon's, not any request's. The dial itself is
-// bounded by a timer, not a context deadline: the library's dial+auth I/O
-// ignores deadlines, and a deadline on the connection context would close
-// the connection when it fired.
+// its lifetime is the daemon's, not any request's (a deadline on it would
+// close the connection when it fired). The attempt stays marked in flight
+// until it truly returns: the library's dial+auth I/O has no deadline, so
+// abandoning a hung attempt and retrying would leak a goroutine per retry;
+// held single-flight, a wedged PID1 costs at most one blocked goroutine
+// while every caller keeps exec-falling-back.
 func sdbusDial() {
-	type result struct{ conn *sddbus.Conn }
-	ch := make(chan result, 1)
-	go func() {
-		conn, err := sddbus.NewSystemdConnectionContext(context.Background())
-		if err != nil {
-			ch <- result{nil}
-			return
-		}
-		ch <- result{conn}
-	}()
-
-	var conn *sddbus.Conn
-	select {
-	case r := <-ch:
-		conn = r.conn
-	case <-time.After(sdbusDialTimeout):
-		// Abandoned: if the dial ever completes, close the orphan.
-		go func() {
-			if r := <-ch; r.conn != nil {
-				r.conn.Close()
-			}
-		}()
-	}
+	conn, err := sddbus.NewSystemdConnectionContext(context.Background())
 
 	sdbus.mu.Lock()
 	sdbus.dialing = false
-	if conn != nil {
+	sdbus.lastDial = time.Now() // cooldown runs from completion, not start
+	if err == nil {
 		if sdbus.conn == nil {
 			sdbus.conn = conn
 		} else {
-			// Lost a race with another successful dial.
+			// Lost a race with a connection cached by someone else.
 			defer conn.Close()
 		}
 	}
--- a/internal/vm/sdbus.go
+++ b/internal/vm/sdbus.go
@@ -157,10 +157,22 @@ func sdbusDo(fn func(*sddbus.Conn) error) (err error, handled bool) {
 	if merr, answered := sdbusAnswer(err); answered {
 		return merr, true
 	}
-	// Transport-shaped failure: drop the corpse so a later call redials,
-	// and let the exec path handle this call with today's semantics.
-	sdbusDrop(conn)
+	// Transport-shaped failure: drop the corpse so a later call redials —
+	// unless it's the CALLER's cancellation, which is this call's result,
+	// not evidence the shared bus is dead. Either way the exec path
+	// handles this call with today's semantics.
+	if sdbusCondemns(err) {
+		sdbusDrop(conn)
+	}
 	return nil, false
+}
+
+// sdbusCondemns reports whether a non-answer error condemns the shared
+// connection. Caller ctx expiry doesn't: the library surfaces it without
+// the connection being unhealthy, and dropping would tear the bus down
+// under every other in-flight operation.
+func sdbusCondemns(err error) bool {
+	return !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded)
 }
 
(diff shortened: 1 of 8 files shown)


meAmitPatil:
Real — this was the top finding of the internal review pass as well (same root cause: godbus closes the connection when the dialing ctx ends). Fixed in 13ece9e, which reworked the whole acquisition model: the dial runs in a single-flight background goroutine on a daemon-lifetime context (bounded by a timer, since dial/auth I/O ignores ctx deadlines), callers never block on connection state (no healthy conn → exec fallback immediately), Connected() is checked on every acquisition to catch a half-dead signal socket, and per-call contexts now bound only individual method calls. Same commit also stops re-driving a stop via exec once systemd has accepted the job, and treats NoSuchUnit as the definitive dead-unit answer.

meAmitPatil:
Real — the rework's drop-on-transport logic was too eager: godbus surfaces a per-call ctx expiry without the connection being unhealthy, and the catch-all condemned the shared bus for it. Fixed in b25825e: sdbusCondemns excludes context.Canceled/DeadlineExceeded — caller cancellation is that call's result (exec fallback handles it with today's semantics), never grounds to drop the connection under other in-flight operations. Test pins the distinction.

meAmitPatil:
Real — the abandon-and-retry machinery was the leak: each abandoned attempt strands a goroutine pair on unbounded dial/auth I/O, and the cooldown starts another. Fixed in the latest commit by deleting that machinery: the attempt stays marked in flight until it truly returns (single-flight), so a wedged PID1 costs at most ONE blocked goroutine total while every caller keeps exec-falling-back; if the dial eventually completes, the connection is cached (or closed if raced). The timer bought nothing — callers never wait on the dial — so this is also net-simpler.

awille-ai (approved):
Reviewed with the same security/correctness/scalability bar. I did not find any critical or high-severity issues, so I’m approving.

The current head addresses the important persistence and failure-isolation cases:
- the cached D-Bus connection is created with daemon lifetime rather than a request context;
- caller cancellation does not tear down the shared connection underneath unrelated operations;
- a hung dial remains single-flight instead of leaking a new blocked goroutine every cooldown;
- connection loss

internal/vm/systemd.go
```diff
@@ -43,31 +120,44 @@ func unitFailureSummary(ctx context.Context, unit string) string {
...
+	s, _ := val.(string)
+	if s == "" {
+		return false, false
+	}
+	return s == "active" || s == "reloading", true
 }
 
 // isUnitActive checks if a systemd unit is currently active (running).
 func isUnitActive(ctx context.Context, unit string) bool {
+	if active, ok := unitActiveState(ctx, unit); ok {
```
pavitrabhalla: **Blocker.** isUnitActive can spend the caller's whole ctx budget on the D-Bus attempt, then falls through to `exec.CommandContext(ctx, ...)` with that same expired ctx — the exec call fails instantly and reads as "not active". handleVMError (manager.go:2107-2114) passes a 3s checkCtx and treats `false` as "VM is gone": deletes it from `m.vms`, marks StatusStopped, returns NotFound. unitDefinitelyDead has an explicit `ctx.Err() != nil` guard for this exact case; isUnitActive doesn't. Under the concurrent-launch-burst load this PR targets, a slow/congested shared connection could genuinely eat the whole budget before the D-Bus call returns.

Should isUnitActive get the same "ctx already spent → inconclusive, not dead" treatment as unitDefinitelyDead, e.g. by reserving a minimum slice of ctx for the exec fallback or treating a D-Bus ctx-expiry as "can't tell" rather than falling through?

internal/vm/systemd.go
```diff
@@ -26,10 +41,72 @@ func startUnit(ctx context.Context, unit string) error {
...
+const stopJobWaitCap = 15 * time.Second
+
+// stopUnit stops a systemd unit. Idempotent — stopping an already-stopped
+// unit is a no-op. Blocks until the stop job completes.
+func stopUnit(ctx context.Context, unit string) error {
+	// Buffered channel: the library's dispatcher does one blocking send;
+	// buffer 1 keeps an abandoned wait from wedging every job on the
+	// connection.
+	ch := make(chan string, 1)
+	err, enqueued := sdbusDo(func(c *sddbus.Conn) error {
```
pavitrabhalla, line 74: **Blocker (or at least needs a call).** stopUnit passes a non-nil channel to `StopUnitContext`. In go-systemd v22.5.0, `startJob` does `if ch != nil { c.jobListener.Lock(); defer c.jobListener.Unlock() }` around the blocking `CallWithContext` — the lock is held for the full network round trip, not just the map insert. startUnit avoids this entirely by passing `ch=nil` (per its own comment: "skips the library's job-tracking lock, so concurrent launches pipeline on the socket"). stopUnit can't do the same since it needs the completion channel.

Since vmd uses one process-wide shared `*sddbus.Conn`, a host stopping/draining many VMs at once has every concurrent stopUnit call serialize on this one mutex for the full round trip of whichever stop currently holds it. Isn't that the same per-call bottleneck this PR set out to eliminate, just relocated from starts to stops? Worth confirming whether that's an accepted tradeoff (stops are rarer/less bursty than starts) or something to design around.

internal/vm/systemd.go
```diff
@@ -26,10 +41,72 @@ func startUnit(ctx context.Context, unit string) error {
...
 // unitFailureSummary returns a one-line summary of a unit's
 // Result/SubState for embedding in error messages. Best-effort —
 // returns "unknown" on any error.
 func unitFailureSummary(ctx context.Context, unit string) string {
+	// Result lives on the Service interface, SubState on Unit.
+	if res, notLoaded, ok := sdbusUnitProperty(ctx, unit, "Service", "Result"); ok {
+		if notLoaded {
+			return "not-loaded"
+		}
+		if sub, _, sok := sdbusUnitProperty(ctx, unit, "", "SubState"); sok {
```
pavitrabhalla, line 149: Non-blocking. unitFailureSummary now does two independent D-Bus round trips (Result, then SubState) instead of the old atomic `systemctl show --property=Result,SubState`, and the second call's `notLoaded` is discarded via `sub, _, sok := ...`. A unit that transitions between the two calls could yield a Result/SubState pair that never coexisted, or lose the not-loaded signal if collected between calls. Low impact today — manager.go:2631 only embeds this in a log/error string with nothing branching on the exact content — but was the split intentional, or would a single combined read (e.g. `GetAllPropertiesContext`) be cleaner here?

internal/vm/sdbus.go
```diff
@@ -0,0 +1,230 @@
...
+	"time"
+
+	sddbus "github.com/coreos/go-systemd/v22/dbus"
+	"github.com/godbus/dbus/v5"
+	zlog "github.com/rs/zerolog/log"
+)
+
+// systemdDBusEnabled gates the D-Bus path; set once at startup from
+// VMD_SYSTEMD_DBUS. Off → helpers exec systemctl as before.
+var systemdDBusEnabled atomic.Bool
```
pavitrabhalla, line 42: Non-blocking design question. `systemdDBusEnabled` and the `sdbus{conn,dialing,lastDial}` state are package-level globals set via a free function called from main.go, while every other feature flag in this package (UffdEnabled, VerifySnapshotEnabled, IncrementalSnapshotEnabled, HandlerDeathAbortEnabled) is threaded through `ManagerConfig`. `sdbus_test.go` also has no cleanup/reset for this state — the moment any test enables the flag it leaks a live dial goroutine + connection for the rest of the test binary's process lifetime. `manager_test.go` builds 28+ `Manager` literals directly, so Manager isn't strictly a process-wide singleton even in this codebase.

Was the global intentional (e.g. because the underlying `*sddbus.Conn` really is a one-per-host resource), or would threading a bool through `ManagerConfig` (with the connection singleton still living at package scope underneath) be worth it for testability?

internal/vm/sdbus.go
```diff
@@ -0,0 +1,230 @@
...
+	if v {
+		sdbusAcquire()
+	}
+}
+
+// sdbusRedialCooldown is the minimum gap between dial attempts, so a host
+// where systemd is unreachable doesn't burn a dial per helper call.
+const sdbusRedialCooldown = 3 * time.Second
+
+var sdbus struct {
```
pavitrabhalla, line 58: Minor cleanup. sdbusAcquire hand-rolls single-flight dial dedup (`dialing bool` + mutex + background goroutine). `golang.org/x/sync/singleflight` is already a dependency and already used in this exact package for an analogous pattern (`reattachSF` in manager.go). `Group.DoChan` (unlike `Do`) doesn't block the caller, so it should fit sdbusAcquire's "never block, at most one dial in flight" contract — worth reusing instead of a second hand-rolled implementation of the same idea?

pavitrabhalla (changes requested):
Two blockers below (isUnitActive's ctx reuse, stopUnit's jobListener lock); the rest are minor/non-blocking. Also flagging PauseVM (internal/vm/manager.go:763) as worth a look even though it's outside this diff's hunks — it has no PID-kill fallback if stopUnit stalls/fails, unlike DestroyVM, and stopUnit's blocking semantics/failure modes just changed underneath it.

pavitrabhalla (approved):

PR #213, opened 2026-07-14
vmd: treat a socket as ready only once it accepts connections

meAmitPatil:
A unix socket file appears at bind(), but connect() is refused until listen(). The launch readiness check treated file existence as readiness, so under a concurrent burst — where scheduler pressure stretches the normally sub-microsecond bind→listen gap to milliseconds — the very first API call could land in the gap, get connection refused, and fail the whole launch. A small fraction of burst creates failed this way.

The fix: waitForSocket now has a connectability phase after file existence, and the same policy covers every spawn site — template-builder's stat-only copy is deleted in favor of the exported WaitForAPISocket.

Shape of the connect phase, each rule load-bearing:
- ONLY a refused connection is waited out (the gap, or a dead process's lingering socket). Any other error — ENOENT from a teardown race, ENOTSOCK — surfaces immediately with the true errno, preserving fast failure.
- The wait is capped at 1s past file-appearance: a firecracker that bound and then died also refuses, and must fail in bounded time rather than spinning the caller's full budget.
- It is floored at 250ms so a socket appearing at the edge of the file budget still gets a real chance to reach listen() — a bounded overrun of the nominal timeout, never a cliff.
- Fast path: one successful connect+close (tens of microseconds). Error texts keep the actual budget for log greppability.

Tests reproduce the exact race with bind()-without-listen() sockets: the gap is waited out, a never-listening socket fails within the cap, and a non-refused error fails fast.

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -2692,27 +2692,41 @@ func (m *Manager) resolveAndSetPID(vmID string) {
 // at bind(), and a connect() in the bind()→listen() gap is refused.
 func waitForSocket(path string, timeout time.Duration) error {
 	deadline := time.Now().Add(timeout)
-	if err := waitForSocketFile(path, deadline); err != nil {
+	if err := waitForSocketFile(path, deadline, timeout); err != nil {
 		return err
 	}
-	return waitForSocketConnectable(path, deadline)
+	// The connect phase gets its own bounded window from file-appearance:
+	// capped at 1s so a process that bound and then died fails fast (its
+	// socket refuses just like the gap does), floored at 250ms so a socket
+	// appearing at the edge of the file budget still gets a real chance to
+	// reach listen() — a bounded overrun of the nominal timeout.
+	connWindow := min(time.Until(deadline), time.Second)
+	connWindow = max(connWindow, 250*time.Millisecond)
+	return waitForSocketConnectable(path, time.Now().Add(connWindow))
+}
+
+// WaitForAPISocket is the exported readiness wait for a firecracker API
+// socket: file existence AND accepting connections. For use by every
+// spawn site, in and out of this package.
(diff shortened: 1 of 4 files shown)


awille-ai (approved):
Reviewed with the same security/correctness/scalability bar. I did not find any critical or high-severity issues, so I’m approving.

The implementation closes the actual bind-before-listen race rather than treating socket-file existence as readiness, applies the same readiness helper to the template-builder path, retries only ECONNREFUSED, bounds the post-appearance wait, and preserves fast failure for teardown/type/permission errors. The tests exercise the listening fast path, delayed listen, never-listening cap, and non-refused fast-failure behavior. CI is green.

One minor residual consideration: `net.Dial("unix", path)` itself has no explicit per-attempt deadline

pavitrabhalla (approved):

internal/vm/manager.go
```diff
@@ -2686,21 +2687,46 @@ func (m *Manager) resolveAndSetPID(vmID string) {
...
+	if err := waitForSocketFile(path, deadline, timeout); err != nil {
+		return err
+	}
+	// The connect phase gets its own bounded window from file-appearance:
+	// capped at 1s so a process that bound and then died fails fast (its
+	// socket refuses just like the gap does), floored at 250ms so a socket
+	// appearing at the edge of the file budget still gets a real chance to
+	// reach listen() — a bounded overrun of the nominal timeout.
+	connWindow := min(time.Until(deadline), time.Second)
+	connWindow = max(connWindow, 250*time.Millisecond)
```
pavitrabhalla, line 2887: We have too many values now that are hardcoded numbers like this. Let's consolidate all of these into a shared constants file, so we can take a holistic view at how different numbers are playing together.

meAmitPatil: Sure, will do this as a follow up.

PR #215, opened 2026-07-14
vmd: inherit the gRPC listener from a systemd socket unit

meAmitPatil:
A vmd restart (deploys, mostly) closes its gRPC listener, so connections arriving during the swap get connection refused until the new process binds. The control plane retries through that window, but requests arriving early in a slow drain can outlive the retry budget, and non-retrying clients see hard failures.

Socket activation removes the refused window at the source: a systemd socket unit owns the listening fd, so it survives vmd restarts — connections arriving mid-deploy queue in the kernel accept backlog and are accepted by the fresh vmd within a second (it serves before reattach, behind the retriable startup gate). The canonical graceful-restart pattern.

Mechanics:
- vmd inherits the listener via sd_listen_fds when present; with no socket unit (dev, tests, unmigrated hosts) it binds directly, exactly as before — the code is inert until the unit is installed, which makes rollout and rollback a unit-file operation per host.
- The service unit Requires/After the socket unit, so a plain systemctl restart handles even the first-time migration: the stop releases the directly-bound port, the dependency start binds the socket, the service inherits it.
- The deploy script ships and enables the socket unit; the restart flow is unchanged.

In-flight gRPC streams still terminate on restart (retried by their callers); this makes new connections seamless, not existing ones.

### DIFF
--- a/.github/workflows/deploy-vmd.yml
+++ b/.github/workflows/deploy-vmd.yml
@@ -31,6 +31,7 @@ on:
       - 'internal/db/**'
       - 'internal/secretsproxy/**'
       - 'deploy/superserve-vmd.service'
+      - 'deploy/superserve-vmd.socket'
       - 'deploy/superserve-secretsproxy.service'
       - 'deploy/firecracker@.service'
       - 'deploy/firecracker-netns@.service'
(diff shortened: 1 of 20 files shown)


meAmitPatil:
Right — added deploy/superserve-vmd.socket to the workflow path filter, so socket-unit-only changes trigger the deploy instead of silently drifting until the next unrelated vmd change.

pavitrabhalla (approved):

PR #216, opened 2026-07-15
Aw/metrics fix

awille-ai:
## Summary
Add the production/us-central1 infrastructure needed for OpenTelemetry metrics collection without enabling API OTLP export yet.

## Resources changed
- Production central firewall rule allowing Cloud Run connector traffic to the VMD host on OTLP ports `4317` and `4318`
- `superserve-vmd` network tag on the production central sandbox host
- `roles/monitoring.metricWriter` bound to the existing production host service account
- Production Cloud Monitoring dashboards rendered from shared `templatefile()` sources
- OTEL collector workflow support for manual deployment to production/us-central1

## Production host selector used by the workflow
- `labels.component=vmd AND labels.environment=production AND labels.region=us-central1`

## Expected Terraform changes
- Create the OTLP ingress firewall rule
- Update the production sandbox host tags
- Bind `roles/monitoring.metricWriter` to the existing production host service account
- Create/update the production observability dashboards from the checked-in templates

## Collector deployment behavior
- Uses the GitHub production environment for the production job
- Still downloads the pinned collector binary on the GitHub runner
- Selects the remote architecture, SCPs the matching artifact, installs the config and systemd unit, and validates health/self-metrics
- Writes the production project and zone into `/etc/sandbox/otel/collector.env`
- Keeps staging deployment behavior working

## Manual verification
- `terraform fmt -recursive`
- `cd infra/envs/production/us-central1 && terraform init`
- `terraform validate`
- `terraform plan`
- `python3 -m py_compile .github/workflows/scripts/deploy-otel-collector.py`
- Confirm collector health with `curl -sf http://127.0.0.1:13133/`
- Confirm collector self-metrics with `curl -sf http://127.0.0.1:8888/metrics | grep '^otelcol_'`

## Notes
- Production application metrics remain disabled; this PR only adds the infrastructure and collector path needed for metrics collection later.


infra/envs/production/us-central1/main.tf
```diff
@@ -97,6 +109,18 @@ module "iam" {
       account_id   = "superserve-api-runner"
       display_name = "Superserve API Cloud Run Service Account"
     }
+    (local.vmd_sa_key) = {
+      account_id   = "superserve-vmd-host"
+      display_name = "Superserve VMD Host Service Account"
+    }
+  }
+  project_bindings = {
+    production_host_metric_writer = {
```
awille-ai, line 114: These production JSON files are near-duplicates of the staging dashboards with only the display name and `environment` matcher changed. The `environment` module input does not automatically interpolate inside `file(...)`, but we can avoid maintaining two copies by making the dashboard files templates and loading them with `templatefile(..., { environment = local.environment, environment_title = title(local.environment) })`. Otherwise every panel/query/layout change has to be kept synchronized across two files, and they are likely to drift. Could we

awille-ai (commented):
I found two issues worth addressing before this is merged: the production dashboard copies will drift from staging, and the production workflow currently deploys staging first.

.github/workflows/deploy-otel-collector.yml
```diff
@@ -40,3 +49,33 @@ jobs:
           OTEL_COLLECTOR_PROJECT: ${{ vars.GCP_PROJECT }}
         run: |
           python3 .github/workflows/scripts/deploy-otel-collector.py
+
+  deploy-production:
+    if: github.event_name == 'workflow_dispatch' && github.event.inputs.environment == 'production'
+    needs: [deploy-staging]
+    runs-on: ubuntu-latest
+    environment: production
```
awille-ai, line 57: This still makes a production dispatch run the staging deployment first. `deploy-staging` has no condition, so choosing `environment=production` executes staging, then production waits

awille-ai (commented):
The dashboard duplication issue is fixed cleanly with `templatefile()`. I found two remaining deployment issues and one service-account compatibility concern to address before applying this to production.

.github/workflows/deploy-otel-collector.yml
```diff
@@ -40,3 +50,32 @@ jobs:
           OTEL_COLLECTOR_PROJECT: ${{ vars.GCP_PROJECT }}
         run: |
           python3 .github/workflows/scripts/deploy-otel-collector.py
+
+  deploy-production:
+    if: github.event_name == 'workflow_dispatch' && github.event.inputs.environment == 'production'
```
meAmitPatil, line 54: This can deploy straight to production without staging ever running: the two `if`s are mutually exclusive, so a production dispatch skips `deploy-staging` entirely, and there's no `needs: [deploy-staging]` to gate this job.

That's a departure from every other deploy workflow in the repo — `deploy-vmd.yml` and `deploy-proxy.yml` both have `needs: [deploy-staging]` on their production jobs, and `deploy-api.yml` documents the policy in its header:

> Production NEVER deploys without staging first — whether the trigger is a push, a dispatch, or a forced production run. Selecting "production" does not skip staging; it just promises to run production afterward.

As written, a production dispatch also actively excludes staging, so the staging and production collector versions can drift — exactly the hazard that header calls out.

Suggestion: drop the `environment == 'staging'` arm from the staging job's `if` (let it always run) and add `needs: [deploy-staging]` here, matching the other workflows. If a direct-to-production escape hatch is deliberate for the collector, a header comment saying why would save the next person from "fixing" it.

.github/workflows/deploy-otel-collector.yml
```diff
@@ -14,6 +23,7 @@ on:
 
 jobs:
   deploy-staging:
+    if: github.event_name == 'push' || github.event.inputs.environment == 'staging'
```
meAmitPatil: Not new in this PR, but since you're touching this job: the bot flagged on #185 that `VMD_LABEL: ${{ vars.VMD_LABEL }}` exports an empty string when the Actions variable is unset, which defeats the `component=vmd` default in the deploy script and produces a malformed `labels.` filter. The new production job dodges this by hardcoding `VMD_FILTER`, but staging still carries it. Cheap to fix while you're in here — either drop the env line or have the script treat empty as unset. Fine as a follow-up too.

meAmitPatil (commented):
Dashboards and the Terraform wiring look good — the SA/tag/firewall pieces line up. One real concern on the workflow gating, plus an optional drive-by.

infra/envs/production/us-central1/main.tf
```diff
@@ -98,6 +110,14 @@ module "iam" {
       display_name = "Superserve API Cloud Run Service Account"
     }
   }
+  project_bindings = {
+    production_host_metric_writer = {
+      role = "roles/monitoring.metricWriter"
```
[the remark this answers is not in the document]
awille-ai: intentionally not collecting for production first.

meAmitPatil (approved):

PR #217, opened 2026-07-15
Aw/metrics phase2

awille-ai:
## Summary
Enable production application metrics for `production/us-central1` using OTEL environment variables only.

This updates the production API service to export application metrics to the host-local collector at `http://10.0.0.3:4318` with a 15-second export interval. No telemetry-package production branching was added.

## What changed
- Set `OTEL_ENVIRONMENT=local.environment`
- Set `OTEL_EXPORTER_OTLP_ENDPOINT=http://10.0.0.3:4318`
- Set `OTEL_EXPORT_INTERVAL=15s`
- Set `OTEL_METRICS_ENABLED=true`
- Set `OTEL_SERVICE_NAME=sandbox-controlplane`

## Safety checks
- Application metrics stay bounded to existing low-cardinality labels.
- No sandbox, team, user, API key, request, URL, or raw-error identifiers are introduced as metric labels.
- The collector high-cardinality deletion processor remains unchanged.
- Existing tests for VMD `host_id` propagation and OTel datapoint attributes remain present.

## Validation
- `terraform fmt -recursive`
- `terraform init`
- `terraform validate`
- `terraform plan`
- `go test ./internal/telemetry -v`

## Production rollout checklist
- [ ] Use the `deploy-otel-collector` workflow with `environment=production` to start or repair the production collector on the existing VMD host.
- [ ] Confirm the collector workflow succeeded and the collector is listening on `4318`.
- [ ] Confirm collector health and zero exporter failures.
- [ ] Deploy the API revision.
- [ ] Verify these queries:
  - `db_pool_total_conns{environment="production"}`
  - `host_capacity_running_sandboxes{environment="production"}`
  - `sandbox_transition_total{environment="production"}`
  - `vmd_call_total{environment="production"}`
- [ ] Confirm staging metrics remain unaffected.
- [ ] Roll back by setting `OTEL_METRICS_ENABLED=false` if exporter behavior causes operational issues.

### DIFF
--- a/infra/dashboards/cloud-monitoring/sandbox-telemetry-fleet.json.tftpl
+++ b/infra/dashboards/cloud-monitoring/sandbox-telemetry-fleet.json.tftpl
@@ -217,7 +217,7 @@
             "dataSets": [
               {
                 "timeSeriesQuery": {
-                  "prometheusQuery": "clamp_min(sum(rate(sandbox_transition_total{environment=\"${environment}\",result=\"success\"}[30m])) or vector(0), 0) / clamp_min(sum(rate(sandbox_transition_total{environment=\"${environment}\"}[30m])), 1e-9)"
+                  "prometheusQuery": "(sum(rate(sandbox_transition_total{environment=\"${environment}\",result=\"success\"}[30m])) or vector(0)) / clamp_min(sum(rate(sandbox_transition_total{environment=\"${environment}\"}[30m])), 1e-9)"
                 },
                 "plotType": "LINE",
                 "targetAxis": "Y1",
--- a/infra/envs/production/us-central1/main.tf
+++ b/infra/envs/production/us-central1/main.tf
@@ -150,14 +150,19 @@ module "api" {
   image                 = "${local.region}-docker.pkg.dev/${local.project_id}/superserve/controlplane:replace-me"
 
   env = {
-    API_PORT               = "8080"
-    SUPABASE_URL           = var.supabase_url
-    SECRETS_SIGNING_KEY_ID = "v1"
-    VMD_GRPC_ADDRESS       = format("%s:50051", module.sandbox_host.internal_ip)
-    ALLOW_EPHEMERAL_SEED   = "0"
-    DB_MAX_CONNS           = "12"
-    EDGE_PROXY_DOMAIN      = "sandbox.superserve.ai"
-    KMS_KEY_RESOURCE       = "projects/rayai-prod/locations/us-central1/keyRings/superserve/cryptoKeys/credentials-kek"
+    API_PORT                    = "8080"
+    SUPABASE_URL                = var.supabase_url
+    SECRETS_SIGNING_KEY_ID      = "v1"
+    VMD_GRPC_ADDRESS            = format("%s:50051", module.sandbox_host.internal_ip)
+    ALLOW_EPHEMERAL_SEED        = "0"
+    DB_MAX_CONNS                = "12"
+    EDGE_PROXY_DOMAIN           = "sandbox.superserve.ai"
+    KMS_KEY_RESOURCE            = "projects/rayai-prod/locations/us-central1/keyRings/superserve/cryptoKeys/credentials-kek"
+    OTEL_ENVIRONMENT            = local.environment
+    OTEL_EXPORTER_OTLP_ENDPOINT = "http://${module.sandbox_host.internal_ip}:4318"
+    OTEL_EXPORT_INTERVAL        = "15s"
+    OTEL_METRICS_ENABLED        = "true"
+    OTEL_SERVICE_NAME           = "sandbox-controlplane"
   }
 
   secrets = {
--- a/.github/workflows/deploy-otel-collector.yml
+++ b/.github/workflows/deploy-otel-collector.yml
@@ -23,6 +23,7 @@ on:
 
 jobs:
   deploy-staging:
+    if: github.event_name == 'push' || github.event.inputs.environment == 'staging'
     runs-on: ubuntu-latest
     environment: staging
     permissions:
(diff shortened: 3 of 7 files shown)


awille-ai (commented):
Phase 2 itself is small and looks directionally correct: it enables the existing bounded OTEL configuration for production central and adds a config-loading test. I found two things to fix before treating this as the clean second PR in the stack:

1. The stack is currently out of sync. This PR targets `aw/metrics-fix`, but its recorded base and head have diverged, so GitHub is showing essentially all of Phase 1 again (dashboards, collector workflow, IAM, firewall, service-account changes, tests) instead of only the Phase 2 delta

awille-ai (commented):
I re-reviewed the current version. The Phase 2 goal (enabling production OTEL application metrics) looks right, but I have four concerns worth addressing:

- The production VM service account is still being changed to a new identity. That is a larger operational change than telemetry enablement and I would keep it in a separate least-privilege PR.
- `allow_stopping_for_update = true` allows Terraform to stop the production VMD host for updates. On a single-host production cell this is an outage-capable behavior change and should be introduced intentionally, not as part of Phase 2.
- The DB acquire latency dashboard now clamps the rate denominator to `1`, which under-reports latency whenever acquire rate is below 1/sec.
- The fleet success-ratio query has the same issue: clamping the denominator to `1` causes incorrect ratios at low traffic rates. A small epsilon (or explicit zero handling) preserves the correct ratio while avoiding divide-by-zero.

infra/dashboards/cloud-monitoring/sandbox-telemetry-fleet.json.tftpl
```diff
@@ -217,7 +217,7 @@
             "dataSets": [
               {
                 "timeSeriesQuery": {
-                  "prometheusQuery": "clamp_min(sum(rate(sandbox_transition_total{environment=\"${environment}\",result=\"success\"}[30m])) or vector(0), 0) / clamp_min(sum(rate(sandbox_transition_total{environment=\"${environment}\"}[30m])), 1e-9)"
+                  "prometheusQuery": "sum(rate(sandbox_transition_total{environment=\"${environment}\",result=\"success\"}[30m])) / clamp_min(sum(rate(sandbox_transition_total{environment=\"${environment}\"}[30m])), 1e-9)"
```
awille-ai: This now preserves the ratio correctly, but removing `or vector(0)` from the numerator creates a misleading all-failure edge case: when transitions exist but none have `result="success"`, Prometheus has no success series, so the whole expression returns no data instead of `0`. Please retain zero-filling for the numerator, e.g. `(sum(rate(...result="success"...)) or vector(0)) / clamp_min(sum(rate(...)), 1e-9)`, so a 0% success window is displayed as zero rather than a gap.

awille-ai (commented):
Re-reviewed the latest commit. The two major concerns from the prior pass are fixed: the production VM no longer changes service accounts or opts into being stopped, and the rate denominators are back to using an epsilon rather than `1`. The core Phase 2 change now looks appropriate. I found one remaining dashboard edge case worth fixing; otherwise this is in line with the intended Phase 2 rollout. The generic `allow_stopping_for_update` module plumbing is now inert because its default is false, though it is still unrelated cleanup that could be dropped to keep the PR narrower.

awille-ai (commented):
Re-reviewed the latest commit. The remaining PromQL issue is fixed: the success-rate numerator now preserves `or vector(0)`, so a window with transitions but zero successes renders as 0% instead of no data. I don't see any remaining correctness issues in the Phase 2 OTEL rollout itself.

The only non-blocking cleanup I still notice is the generic `allow_stopping_for_update` module plumbing. Because it defaults to `false` and production no longer enables it, it is inert and safe, but it is unrelated to this PR's stated scope and could be removed to keep the diff tighter. I would not block the Phase 2 rollout on that alone.

infra/envs/production/us-central1/main.tf
```diff
@@ -224,7 +236,9 @@ module "sandbox_host" {
   boot_disk_image       = "projects/ubuntu-os-cloud/global/images/family/ubuntu-2404-lts"
   can_ip_forward        = false
   on_host_maintenance   = "TERMINATE"
-  metadata              = {}
+  metadata = {
+    startup-script = local.otel_collector_bootstrap
+  }
```
awille-ai: This metadata change will be present for future boots, but Compute Engine does not automatically execute a newly-added `startup-script` on an instance that is already running. Terraform can update this metadata in place, so the production collector may remain absent until the VM is rebooted while the API starts exporting immediately. Please handle the initial rollout explicitly—for example, run the bootstrap through the deployment workflow/SSH and verify port 4318 before deploying the API, or plan a controlled restart. The PR checklist should make that ordering unambiguous.

awille-ai (commented):
Re-reviewed the latest commit. The new bootstrap closes the biggest deployment gap: production now has a Terraform-managed way to install and start the collector, and the API endpoint follows the actual host IP rather than assuming `10.0.0.3`.

I found one rollout-blocking issue for the existing production VM: adding or changing the Compute Engine `startup-script` metadata does not itself run the script on an already-running instance. Unless the apply also restarts/recreates the host, the collector will not be installed or started when this PR is applied, even though the API will immediately begin exporting to port 4318. Please add an explicit first-deployment mechanism or rollout step that executes the bootstrap on the existing host, and verify the collector is listening before enabling the API exporter. After that, the architecture looks consistent with the Phase 2 goal.

awille-ai (commented):
Re-reviewed the latest version. I think I found one blocking issue in the GitHub Actions workflow.

The production job is configured with `needs: [deploy-staging]`, but `deploy-staging` is skipped whenever the workflow is manually dispatched for `environment=production`. In GitHub Actions, a job whose dependency is skipped is itself skipped unless explicitly written to handle that case.

As a result, manually dispatching the workflow for production appears to skip the production deployment entirely.

I recommend either:
- removing the dependency for the production deployment, or
- restructuring the workflow so the production job does not depend on a job that is intentionally skipped.

Aside from that workflow issue, I think the rollout plan now makes sense. The PR body now explicitly calls out starting the collector before deploying the API, which addresses my previous concern.

meAmitPatil (approved):

PR #218, opened 2026-07-16
vmd: tag restore phase log with cold/hot segmentation dimensions

meAmitPatil:
## What

Adds six segmentation fields to the existing `restoring snapshot` phase log so a
restore's latency can be attributed to the conditions it ran under:

| field | signal |
| --- | --- |
| `inflight` | concurrent restores at entry (contention on unit-start / socket-bind) |
| `secs_since_template_restore` | seconds since this template mem file last restored — a page-cache warmth proxy; `-1` on first sighting |
| `inplace` | resume vs fresh restore (separates two populations that otherwise blend) |
| `uffd` | memory backend in use (lazy-fault vs eager) |
| `cpu_psi_avg10` | host CPU pressure-stall at entry |
| `mem_psi_avg10` | host memory pressure-stall at entry |

## Why

The phase timings (`start_unit`, `wait_socket`, `wait_boxd`, …) already exist, but
nothing records *whether a given restore hit warm or cold conditions*. As a result
the percentiles blend cold and warm creates together, so p50 looks like "mostly
warm" and the tail looks like "unlucky cold," and neither is separable. These tags
make cold-path vs warm-path latency a first-class, groupable dimension instead of
something inferred by hand from timestamp clustering — the prerequisite for
measuring (and then flattening) temperature-driven latency variance.

## Safety

Observability-only. It adds no branch, return, or error path to the restore, and
the six captured values feed **only** the log line — remove them and the restore
is byte-identical.

- New `tplLastRestore` map is gated by `isTemplateMemPath`, so it keys on the
  finite set of template mem files only; per-VM resume snapshots short-circuit to
  `-1` and are never inserted (the map cannot grow unbounded on vmIDs).
- Guarded by a dedicated leaf mutex (`tplMu`) held only for the map op — no
  nesting with the manager lock, no channel/IO under it.
- PSI reads are best-effort `os.ReadFile` of `/proc/pressure/*`; missing or
  unparseable (older kernel) yields `-1`, never an error or panic.

Unit tests cover the template-age helper (first-seen, delta, resume-not-tracked)
and the PSI parser (some-vs-full line, missing file).

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -252,6 +253,11 @@ type Manager struct {
 	// channel; capacity = effective MaxConcurrentRestores.
 	restoreSem chan struct{}
 
+	// tplLastRestore: last restore time per template mem file, for the
+	// secs_since_template_restore phase tag (a page-cache warmth proxy).
+	tplMu          sync.Mutex
+	tplLastRestore map[string]time.Time
+
 	// builds tracks in-flight and completed template builds. Keyed by
 	// build VM id (which is also "build-" + templateID). Entries survive
 	// until process exit so late pollers can read terminal outcomes.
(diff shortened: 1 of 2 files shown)


awille-ai (approved):
Reviewed at the same security, correctness, reliability, and scalability bar. The previously raised observability-correctness issues are addressed: warmth is recorded only after a successful boot, layered restores use the template base path, UFFD restores do not mark warm at load time, and inflight excludes the current restore. The change remains observability-only, uses bounded template keys and a leaf mutex, PSI reads fail closed to sentinel values, and CI is green. I found no critical or high-severity issue.

PR #220, opened 2026-07-16
Authorize impersonated activity log reads

mohamedsobhi777:
## Summary

- authorize console impersonation keys with `platform:activity:read` for `GET /activity`
- preserve the existing `settings:read` RBAC path for normal customer API keys
- document the endpoint's 403 response in OpenAPI
- add integration coverage for target-team isolation, missing scopes, scope narrowness, and key-name enforcement

## Why

Console impersonation keys do not carry a customer actor ID, so the normal team-settings RBAC check always rejected audit-log requests. Sandbox and template reads already have narrowly scoped impersonation exceptions, but activity reads did not.

## Impact

The control plane can return the impersonated team's activity only when the temporary console key has the explicit activity-read scope. Sandbox-only and template-only keys remain unable to read activity.

## Validation

- `go test ./internal/api/...`
- integration test package compiled successfully with the `integration` build tag
- Docker-backed integration execution was unavailable because the local Docker daemon was not running


awille-ai (commented):
Reviewed together with superserve PR #285 against the existing impersonation model.

I did not find a new audit endpoint or a new cross-team access path. This change reuses the existing `GET /activity` handler and only adds a narrow impersonation exception for temporary console impersonation keys carrying `platform:activity:read`.

Security properties checked:
- Normal customer API keys still use the existing team-scoped `settings:read` RBAC path.
- The impersonation bypass requires both the canonical key name `__console_impersonation__` and the explicit `platform:activity:read` scope.
- A non-impersonation API key with the same scope is still denied.
- Sandbox-only and template-only impersonation keys remain unable to read activity.
- Activity-only impersonation keys remain unable to read sandboxes or templates.
- The handler resolves `team_id` from the authenticated API-key context and queries activity with that team ID, so it does not accept an arbitrary team selector from the request.
- No write path is introduced.
- The existing pagination cap remains in place, so this does not add an unbounded audit-log export path.

The integration coverage for target-team isolation, missing scope, scope narrowness, and temporary-key-name enforcement is the right set of tests. I did not find a blocking security issue in this PR.

meAmitPatil (approved):

PR #221, opened 2026-07-16
Sandboxes: add saved snapshots and independent forks

mohamedsobhi777:
## Summary

Adds Saved Snapshots and Forks V1 for SuperServe sandboxes.

A team can capture an immutable point-in-time snapshot of an active or paused sandbox—including memory, process, and disk state—and create multiple independent sandboxes from it with `POST /sandboxes` using `from_snapshot`.

The feature is disabled globally by default and is intended for a team-scoped staging canary before SDK or Console integration.

## Why

The existing snapshot model is a replaceable runtime checkpoint used to resume the same paused sandbox. It is mutable, tied to one sandbox lifecycle, and cleaned up with that sandbox, so it is not safe to expose as a reusable snapshot.

Reusable forks need immutable artifacts, fresh writable layers, durable lineage, crash-safe lifecycle handling, deduplicated storage accounting, and fresh child identity/security state.

## Public API

Adds:

- `POST /sandboxes/{sandbox_id}/snapshots`
- `GET /snapshots`
- `GET /snapshots/{snapshot_id}`
- `DELETE /snapshots/{snapshot_id}`
- `POST /sandboxes` with `from_snapshot`
- `source_snapshot_id` on sandbox responses

Capture is synchronous and supports `Idempotency-Key`. A successful `201` means the immutable manifest and artifacts are committed and the source has returned to its prior active or paused state. `from_template` and `from_snapshot` are mutually exclusive.

### Immutable capture and restore

- Captures active sandboxes with a bounded freeze and restores the source before returning.
- Captures paused sandboxes offline without waking them or consuming their runtime checkpoint.
- Commits versioned manifests atomically after fsync and integrity hashing.
- Uses mandatory filesystem reflinks and fresh writable layers so sources and sibling forks diverge independently.
- Negotiates the VMD `saved_snapshots_v1` capability and fails closed on unsupported hosts.
- Routes V1 capture, restore, and cleanup to the exact host that owns the artifacts.

### Lineage, recovery, and cleanup

- Separates replaceable runtime checkpoints from immutable saved snapshots.
- Records parent snapshots, source sandboxes, template pins, resource shape, and artifact identity.
- Serializes capture against pause, resume, deletion, configuration mutation, and competing captures.
- Prevents deletion while descendant snapshots or sandboxes still depend on an artifact.
- Adds a supervisor for stale `creating` and `deleting` operations.
- Makes capture and cleanup idempotent across lost responses and process restarts.
- Refuses `migrate-team` for teams with V1 host-local snapshot state.

### Policy and security

- Inherits or explicitly replaces snapshot-time network and secret-binding policy.
- Mints fresh child access credentials, secret proxy tokens, and JWTs.
- Scrubs captured source and removed-binding credentials before child exposure.
- Preserves team-scoped lookup behavior and does not expose host paths or secret material.

### Quotas and shadow accounting

- Defaults to 100 saved snapshots per team and 20 per source sandbox.
- Requires an explicit per-team snapshot-storage quota.
- Tracks copy-on-write layer ownership and references.
- Uses FIEMAP-based exclusive extent measurement and a team shadow counter.
- Does not affect invoice/export behavior in this PR.

### Observability and canary

Adds bounded snapshot outcome and VMD RPC telemetry, plus `scripts/smoke-test-snapshots.sh`. The canary covers active and paused capture, concurrent and nested forks, memory/process/file preservation, divergence, credential isolation, inherited restricted egress, pause/resume, restart recovery, storage-ledger consistency, and dependency-safe deletion.

## Database and rollout

The migration is additive but substantial. It adds lifecycle and lineage fields, team quota fields, sandbox references, storage layer/reference tables, indexes, constraints, and transactional quota/accounting guards.

Staging rollout order:

1. Apply the database migration.
2. Deploy VMD from the same release SHA.
3. Verify `GetCapabilities.saved_snapshots_v1 == true`.
4. Deploy the control plane.
5. Keep the global feature default disabled.
6. Enable `sandbox_snapshots_v1` only for a dedicated canary team.
7. Provision an explicit storage quota; recommended staging value is 100 GiB.
8. Run the snapshot canary with the restart hook and shadow-ledger verification.

VMD restore and cleanup support must remain deployed while saved artifacts or descendants exist.


awille-ai (changes requested):
Blocking: migration is not backward-compatible with the currently deployed control plane

I think this PR introduces a deployment compatibility issue.

The new schema changes runtime checkpoints from a global uniqueness constraint on snapshot.sandbox_id to a partial unique index for runtime checkpoints only:

UNIQUE (sandbox_id) WHERE kind = 'runtime_checkpoint'

However, the currently deployed control plane (pre-PR) still executes:

INSERT ...
ON CONFLICT (sandbox_id)

Postgres cannot match that conflict target to a partial unique index, so after the migration is applied every FinalizePause fails with:

ERROR: there is no unique or exclusion constraint matching the ON CONFLICT specification
SQLSTATE 42P10

I was able to reproduce this in staging:

VMD successfully created the runtime checkpoint.
FinalizePause failed with 42P10.
The sandbox remained stuck in pausing.
Resume/delete then failed because the lifecycle transition never completed.

This PR correctly updates the query to:

INSERT INTO snapshot (
  ...
  kind,
  status
)
...
ON CONFLICT (sandbox_id)
WHERE kind = 'runtime_checkpoint'

so the code itself is correct.

The problem is rollout.

Today the deployment order is:

Apply migration.
Deploy VMD.
Deploy control plane.

That means there is a period where the old control plane is running against the new schema. During that period, ordinary pause/resume is broken even though saved snapshots are feature-gated.

This is a deployment compatibility issue rather than a feature issue.

I don't think we should merge this until we have a rollout strategy that guarantees compatibility, for example:

an expand/contract migration,
a compatibility release deployed before the schema change,
or another approach that avoids old binaries executing ON CONFLICT (sandbox_id) after the migration.

Otherwise a production rollout has a window (or longer if deployment fails after migration) where every pause operation will fail.

awille-ai:
Staging currently has a temporary snapshot_sandbox_unique index restored manually to keep the pre-PR control plane compatible with the migrated schema. Please preserve DROP INDEX IF EXISTS snapshot_sandbox_unique in the final migration so the temporary index is removed during the corrected rollout.

pavitrabhalla:
Closing — superseded by the split stack (#284–#287), which is itself being closed: we're redoing snapshot/restore from scratch around a filesystem-only, GCS-durable contract.

PR #224, opened 2026-07-16
Misc terraform fixes

awille-ai:
Summary of the PR changes:

  - Split Terraform PR CI into two workflows:
      - .github/workflows/terraform-checks.yml runs on every PR.
      - .github/workflows/terraform-plans.yml runs only when Terraform-related paths change.

  - Made Terraform formatting an always-on required check:
      - Terraform Fmt now runs on every PR.
      - It does terraform fmt -check -recursive infra/ and does not require GCP auth.

  - Added always-on Terraform validation without production state:
      - New helper scripts/terraform-validate.sh runs terraform init -backend=false and terraform validate for the supported roots.
      - It now covers:
          - infra/envs/staging/us-central1
          - infra/envs/production/us-central1
          - infra/envs/production/us-west2
          - infra/envs/production/us-east4

  - Moved authenticated Terraform plans behind native PR path filtering:
      - Plans now run only for:
          - infra/**
          - scripts/terraform-*.sh
          - .github/workflows/terraform-*.yml

      - The workflow no longer depends on a PR changed-files lookup or GitHub listFiles.

  - Restored and fixed the consolidated plan summary:
      - Each matrix plan writes its own summary file.
      - The consolidation job now collects all per-environment summaries and publishes one combined PR summary.
      - It no longer loses files due to overwrite/merge issues.

  - Added us-east4 to PR plan coverage:
      - The plan matrix now includes production-us-east4.
      - This environment already existed in the repo; the PR just adds CI coverage for it.

  - Kept plan behavior read-only:
      - Plans do init, validate, and plan.
      - No workflow introduced terraform apply.

  Verification:

  - terraform fmt -check -recursive infra/ passed.
  - scripts/terraform-validate.sh passed.


infra/envs/production/us-west2/main.tf
```diff
@@ -118,6 +118,7 @@ module "api" {
     SANDBOX_ID_REGION = "usw"
     SUPABASE_URL      = var.supabase_url
     VMD_GRPC_ADDRESS  = "10.1.0.2:50051"
+    SYSTEM_TEAM_ID    = "258e290e-d30d-4d2a-b751-07da118248c0"
```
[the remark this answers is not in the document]
awille-ai: they have the same SYSTEM_ID in the DB

infra/envs/production/us-central1/main.tf
```diff
@@ -132,10 +142,17 @@ module "api" {
...
+    API_PORT               = "8080"
+    SUPABASE_URL           = var.supabase_url
+    SENTRY_DSN             = "https://4497bf1fc61a2388b355489982086ac7@o4511499764760576.ingest.us.sentry.io/4511499780620288"
+    SECRETS_SIGNING_KEY_ID = "v1"
+    VMD_GRPC_ADDRESS       = "10.0.0.3:50051"
+    SENTRY_DSN             = "https://4497bf1fc61a2388b355489982086ac7@o4511499764760576.ingest.us.sentry.io/4511499780620288"
+    SYSTEM_TEAM_ID         = "258e290e-d30d-4d2a-b751-07da118248c0"
+    ALLOW_EPHEMERAL_SEED   = "0"
+    DB_MAX_CONNS           = "12"
+    EDGE_PROXY_DOMAIN      = "sandbox.superserve.ai"
```
pavitrabhalla, line 160: Revising my earlier note here — this flip is defensible, but worth confirming intent.

`EDGE_PROXY_DOMAIN` goes from `usc-sandbox.superserve.ai` to the legacy apex `sandbox.superserve.ai`. That actually lines up better than the old value: the SDKs derive the data-plane host client-side from the API-key region token, and both the `use` region and all legacy keys resolve to `sandbox.superserve.ai` — no client ever derives `usc-sandbox`. So this makes server-built URLs (previews etc.) match what clients hit.

Two things to confirm:
1. This is the *legacy* apex, not the regional name the west cell uses (`usw-sandbox`). If the intent is symmetric per-cell regional domains, the target here is `use-sandbox.superserve.ai` — but that only works once the proxy accepts a multi-domain set and the SDKs map `use` → `use-sandbox`. Until then `sandbox.superserve.ai` is the right (and permanent, back-compat) choice.
2. Is the proxy in front of this cell actually serving `sandbox.superserve.ai` today, so URLs built from this env var resolve end-to-end?

infra/envs/production/us-central1/main.tf
```diff
@@ -132,10 +150,15 @@ module "api" {
...
   env = {
-    API_PORT          = "8080"
-    EDGE_PROXY_DOMAIN = "usc-sandbox.superserve.ai"
-    SUPABASE_URL      = var.supabase_url
-    VMD_GRPC_ADDRESS  = "10.0.0.3:50051"
+    API_PORT               = "8080"
+    SUPABASE_URL           = var.supabase_url
+    SECRETS_SIGNING_KEY_ID = "v1"
+    VMD_GRPC_ADDRESS       = "10.0.0.3:50051"
+    SYSTEM_TEAM_ID         = "258e290e-d30d-4d2a-b751-07da118248c0"
```
pavitrabhalla: this should be set as a secret. Let's rotate it in the DB and set it as a secret.

infra/envs/production/us-west2/main.tf
```diff
@@ -113,11 +113,16 @@ module "api" {
...
-    VMD_GRPC_ADDRESS  = "10.1.0.2:50051"
+    API_PORT               = "8080"
+    EDGE_PROXY_DOMAIN      = "usw-sandbox.superserve.ai"
+    SANDBOX_ID_REGION      = "usw"
+    SUPABASE_URL           = var.supabase_url
+    SECRETS_SIGNING_KEY_ID = "v1"
+    ALLOW_EPHEMERAL_SEED   = "0"
+    DB_MAX_CONNS           = "12"
+    VMD_GRPC_ADDRESS       = "10.1.0.2:50051"
+    SYSTEM_TEAM_ID         = "258e290e-d30d-4d2a-b751-07da118248c0"
```
pavitrabhalla: need to turn this into a secret as well

infra/envs/production/us-west2/main.tf
```diff
@@ -113,11 +113,16 @@ module "api" {
...
-    SUPABASE_URL      = var.supabase_url
-    VMD_GRPC_ADDRESS  = "10.1.0.2:50051"
+    API_PORT               = "8080"
+    EDGE_PROXY_DOMAIN      = "usw-sandbox.superserve.ai"
+    SANDBOX_ID_REGION      = "usw"
+    SUPABASE_URL           = var.supabase_url
+    SECRETS_SIGNING_KEY_ID = "v1"
+    ALLOW_EPHEMERAL_SEED   = "0"
+    DB_MAX_CONNS           = "12"
+    VMD_GRPC_ADDRESS       = "10.1.0.2:50051"
```
pavitrabhalla: should also be a secret imo

infra/envs/staging/us-central1/main.tf
```diff
@@ -150,6 +150,7 @@ module "api" {
     OTEL_SERVICE_NAME           = "sandbox-controlplane"
     SUPABASE_URL                = var.supabase_url
     VMD_GRPC_ADDRESS            = "10.0.0.2:50051"
+    SYSTEM_TEAM_ID              = "a79a37c6-8622-4caa-b0e8-008ab8e4b78e"
```
pavitrabhalla: should rotate these and move to secrets

infra/envs/production/us-central1/main.tf
```diff
@@ -132,10 +150,15 @@ module "api" {
...
 
   env = {
-    API_PORT          = "8080"
-    EDGE_PROXY_DOMAIN = "usc-sandbox.superserve.ai"
-    SUPABASE_URL      = var.supabase_url
-    VMD_GRPC_ADDRESS  = "10.0.0.3:50051"
+    API_PORT               = "8080"
+    SUPABASE_URL           = var.supabase_url
+    SECRETS_SIGNING_KEY_ID = "v1"
+    VMD_GRPC_ADDRESS       = "10.0.0.3:50051"
```
pavitrabhalla: this should also be a secret. Will need to rotate the grpc address though. Is that safe?

meAmitPatil (approved):

PR #230, opened 2026-07-17
api: resolve sandbox secret bindings in one batched query

meAmitPatil:
Creating a sandbox with attached secrets resolved each reference with its own `GetSecretByName` round trip — N queries for N secrets, run sequentially before the create proceeds. Sandboxes with several secrets paid a query per secret on the create path.

This fetches all referenced secrets in a single `GetSecretsByNames` query (`WHERE name = ANY($1)`), builds a name→row map, then assembles the bindings and mints proxy tokens from it. Distinct names only — multiple env keys mapping to the same secret now cost one lookup, not one each. Behavior is unchanged: a reference to a secret the team does not own still returns the same `secret_not_found` 400, and creates with no secrets still short-circuit before any query.

Tests cover the batched happy path (one query for multiple secrets), de-duplication of a shared secret name across env keys, and the not-found error.

### DIFF
--- a/db/queries/secrets.sql
+++ b/db/queries/secrets.sql
@@ -11,6 +11,10 @@ RETURNING *;
 SELECT * FROM secret
 WHERE team_id = $1 AND name = $2 AND deleted_at IS NULL;
 
+-- name: GetSecretsByNames :many
+SELECT * FROM secret
+WHERE team_id = $1 AND name = ANY($2::text[]) AND deleted_at IS NULL;
+
 -- name: GetSecretByID :one
 SELECT * FROM secret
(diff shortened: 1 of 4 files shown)


awille-ai (approved):
Reviewed with the same security, correctness, reliability, and scalability bar. The batched lookup preserves the existing team_id and deleted_at authorization constraints, de-duplicates only database reads while retaining one binding per environment key, and keeps the existing secret_not_found behavior. The no-secret path still avoids a query, tests cover batching/de-duplication/missing references, and CI is green. I found no critical or high-severity issues.

PR #239, opened 2026-07-20
Preview URLs phase 1: require explicit port publication

mohamedsobhi777:
## Summary

- Require explicit publication before a numeric preview port is routable on new sandboxes.
- Preserve legacy all-port behavior only for sandboxes created by older control planes.
- Gate placement and enablement on a live `preview_ports_v1` host capability.
- Persist monotonic preview-policy snapshots in VMD and remove positive resolver caching so revocations are observed on the next request.
- Keep rolling database compatibility by storing policy, publication, and capability state in RLS-protected side tables.
- Require VMD response-protocol attestation so a new proxy fails closed if VMD is rolled back.
- Reject reserved boxd port `49983` from the publication contract.

## Why

This is the first independently deployable layer of the preview URL security rollout. It establishes deny-by-default port publication without introducing private access or credentials, and it fails closed across single-component control-plane, VMD, and proxy version skew.

No product feature flags are included; the PR/deploy boundary is the rollout boundary. The split follows [the staged review on #201](https://github.com/superserve-ai/sandbox/pull/201#issuecomment-5007470378).

## Stack

1. **This PR:** explicit port publication
2. [#240](https://github.com/superserve-ai/sandbox/pull/240) — per-port public/private policy
3. [#242](https://github.com/superserve-ai/sandbox/pull/242) — per-port header-token authentication
4. [#243](https://github.com/superserve-ai/sandbox/pull/243) — browser signed-link authentication

Client/SDK/docs companion: [superserve#287](https://github.com/superserve-ai/superserve/pull/287), which deploys after Phase 4.

## Rollout / rollback

The proxy and VMD have bidirectional mixed-version guards: an old proxy cannot consume strict policy from new VMD, and a new proxy rejects old-VMD responses that lack protocol attestation. Once strict sandboxes exist, rolling **both** data-plane binaries back requires draining or moving those sandboxes from the host first; `deploy/README.md` records this operational boundary.

## Validation

- All GitHub checks pass, including integration, unit, build, vet, sqlc, migration order, and Terraform checks.
- Focused API, database, scheduler, proxy, VMD-state, and proxy-command tests
- Race tests for API, proxy, preview, DB, and proxy command packages
- Exact equal-revision policy attestation tests
- Linux/amd64 `go build ./...` and `go vet ./...`
- Linux VM test binary compilation

Native macOS execution of Linux-only VM networking tests remains unavailable because those packages depend on nftables constants; their Linux compilation was validated separately.


awille-ai (approved):
Reviewed as Phase 1 and against the complete preview-auth stack. I did not find a blocking correctness or security issue. Explicit publication is enforced at the proxy, legacy behavior is represented explicitly, capability attestation covers the VMD/proxy split, policy updates are monotonic and retryable, and the reserved boxd port is consistently excluded. CI and Terraform checks are green.

Non-blocking simplification note: this phase is independently valuable, but it is still a very large change (46 files / ~5k additions). Much of that is generated code and tests, but a substantial part is mixed-version and rollback machinery. I would keep Phase 1 separate,

PR #241, opened 2026-07-20
api, authz, scheduler: serve stale cache entries while refreshing in the background

meAmitPatil:
A burst arriving on an expired cache entry used to queue behind the refresh: singleflight kept it to one query, but every request arriving during that flight carried the wait. This applies stale-while-revalidate to the four control-plane caches — a just-expired entry is served immediately while one coalesced background flight refreshes it, so bursts never queue on cache expiry.

Per cache, with the staleness each can afford:

- **API key** (grace 2s): worst-case revocation widens ttl→ttl+2s; an expired `expires_at` on the key itself is still checked exactly and never stale-served.
- **Team permissions** (grace 2s): an invalidated (generation-bumped) entry is never stale-served, so same-instance revocation stays immediate; a background refresh that discovers a denial drops the entry, ending stale serves for it at once.
- **Template** (grace 30s): bounds how long a rebuilt/deleted template's old snapshot paths can be handed out; a refresh that finds the template gone drops the entry.
- **Host list** (unbounded): the host set changes via ops actions that call Invalidate for an immediate blocking reload; otherwise any cached list is servable while a CAS-guarded background refresh runs.

The only requests that ever block on a lookup are true first uses (nothing cached) and entries past their grace window.

Tests cover, per cache: stale-serve inside grace, background refresh landing (single coalesced flight), past-grace blocking, and the safety invariants — generation bump beats stale-serve, denial refresh drops the entry, key expiry never stale-served, Invalidate forces a blocking reload.

### DIFF
--- a/cmd/controlplane/main.go
+++ b/cmd/controlplane/main.go
@@ -217,7 +217,8 @@ func run() error {
 		return telemetry.WrapVMDClient(newGRPCVMDClient(conn), recorder, api.SandboxIDRegion(), hostID), nil
 	}
 	handlers.Hosts = hostreg.New(queries, dialVMD)
-	handlers.Scheduler = &scheduler.LeastLoaded{DB: queries, DefaultHostID: cfg.DefaultHostID}
+	sched := &scheduler.LeastLoaded{DB: queries, DefaultHostID: cfg.DefaultHostID}
+	handlers.Scheduler = sched
 
 	router := api.SetupRouter(ctx, handlers, dbPool)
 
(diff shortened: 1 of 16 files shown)


awille-ai (changes requested):
Re-review of the current head. The earlier cache-consistency findings are addressed: invalidation generations prevent pre-invalidation refreshes from overwriting fresh state; API-key refreshes evict on definitive revocation; permission denials remove stale grants; and template refreshes remove stale system entries when a team override appears.

CI is green. I found one remaining high-reliability issue:

## High: stale host health can still be used for scheduling for up to

awille-ai (approved):
Re-review of the current head: the remaining blocking scheduler concern is addressed.

`StartHostDetector` now accepts an invalidation callback, `cmd/controlplane` wires it to `sched.Invalidate`, and the callback runs after at least one stale host is successfully marked unhealthy. Combined with the scheduler generation check, this prevents an in-flight pre-invalidation refresh from restoring the retired host list.

The earlier API-key, permission, template, and stale-refresh race findings also remain addressed, and CI/Terraform checks are green.

One non-blocking operational caveat: invalidation is process-local. In a multi-instance control

PR #243, opened 2026-07-20
Preview URLs phase 4: add browser signed-link authentication

mohamedsobhi777:
## Stack

1. #239 — explicit preview-port publication
2. #240 — per-port public/private policy
3. #242 — per-port header-token authentication
4. **This PR** — browser signed-link authentication

The cross-repository client/SDK/docs companion is [superserve#287](https://github.com/superserve-ai/superserve/pull/287) and must deploy after this backend layer. Review this PR against `agent/preview-port-header-auth`; it contains exactly one Phase 4 commit. This split follows [the staged review on #201](https://github.com/superserve-ai/sandbox/pull/201#issuecomment-5007470378).

## What this adds

- Adds the distinct wire-only `private_browser_v1` policy and `preview_port_browser_auth_v1` host capability, preserving Phase 3's exact header-only contract.
- Returns the reserved `superserve_preview_token` query-parameter name with minted and rotated per-port credentials.
- Exchanges a valid ordinary-GET signed link for a host-only `__Host-` cookie and a clean, absolute same-origin HTTPS 302.
- Sets `Secure`, `HttpOnly`, `SameSite=None`, `Partitioned`, and `Path=/`, with no `Domain`; bootstrap responses are `no-store` / `no-referrer`.
- Accepts a signed query directly for non-GET requests and complete WebSocket upgrade handshakes, while allowing only genuine CORS preflight through without credentials.
- Parses repeated/coalesced header, cookie, and query candidates independently so a stale carrier cannot mask a fresh credential.
- Scrubs every reserved header, cookie, and query occurrence before public, legacy, or private upstream applications receive the request.
- Requires the complete current-heartbeat publication → access → token → browser capability chain on every private activation, reapply, resume, mint, and rotate path.
- Preserves mint rollback and durable rotation-revocation/audit semantics under the sandbox/host transaction locks.
- Persists and compares the Phase 4 sentinel through VMD state, restore, local protocol, policy watermarks, and exact equal-revision checks.

No product feature flags are included; each stacked PR and deploy boundary is the rollout boundary.

## Rollout / rollback

Deploy the browser-aware proxy and VMD first, wait for current heartbeats to attest all four capabilities, then deploy the API, then deploy [superserve#287](https://github.com/superserve-ai/superserve/pull/287). VMD withholds browser policies from a Phase 3 proxy, an older proxy treats the new sentinel as closed, and the API refuses new browser-private mutations when the live chain is incomplete.

## Validation

- Focused normal and race suites for auth, preview, proxy, API, scheduler, control-plane, and VMD state/restore
- Full PostgreSQL integration and `migrate-team` suites under race
- Browser-capability mint rollback and rotation durable-revocation/audit tests
- Deterministic blocked-mint/rotation ordering test with exact `private_browser_v1` revisions
- Adversarial repeated/coalesced carrier, redirect, WebSocket, CORS, raw-query, cookie, and upstream-scrubbing coverage
- Independent security review: no blockers
- OpenAPI synchronization, sqlc drift, Linux/amd64 build/vet, and Linux test compilation
- `git diff --check`

Native execution of Linux-only VM networking tests remains unavailable on macOS because of the repository's existing nftables constant incompatibility; their Linux binaries compile.


awille-ai (commented):
Non-blocking architectural thought: the phased rollout uses several independent capability bits (preview_ports, preview_port_access, preview_port_tokens, preview_port_browser_auth). The implementation is correct, but it does increase the number of compatibility states that have to be reasoned about.

If we add more preview features in the future, it may be worth considering whether a versioned preview capability would be simpler than accumulating additional independent capability flags. I don't think this PR should change, but it's something I'd keep in mind as the feature evolves.

PR #244, opened 2026-07-20
proxy: give exec WebSocket frames their own size limit

meAmitPatil:
The exec WebSocket reused the terminal bridge's 64 KiB per-frame read limit, which is sized for keystrokes. Exec frames legitimately carry a whole command string or a stdin chunk, so an oversized frame tore down the session with only a debug log and no clear signal to the client.

- Exec sockets now use a dedicated 4 MiB frame limit; the terminal bridge keeps its 64 KiB keystroke limit.
- A frame exceeding the limit is logged at warn with the limit value; the client receives a 1009 (message too big) close as before.
- New tests pin both behaviors: a 128 KiB stdin frame reaches boxd intact, and an over-limit frame closes with 1009.

### DIFF
--- a/internal/proxy/exec_ws.go
+++ b/internal/proxy/exec_ws.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


internal/proxy/exec_ws.go
```diff
@@ -377,6 +383,22 @@ func (h *Handler) handleExecControl(ctx context.Context, client boxdpbconnect.Pr
 	}
 }
 
+// logExecReadEnd separates a frame that blew the read limit — a client-visible
+// failure worth a warn — from routine socket teardown. The library reports the
+// limit breach only via error text, and has already closed the connection with
+// StatusMessageTooBig by the time Read returns.
+func logExecReadEnd(l zerolog.Logger, err error) {
+	if strings.Contains(err.Error(), "read limited at") {
```
pavitrabhalla: This matches on err.Error() containing "read limited at" to detect the size-limit close. coder/websocket wraps this case in the exported sentinel ErrMessageTooBig (read.go:528 in v1.8.14) specifically so callers don't have to parse text. Any reason not to use errors.Is(err, websocket.ErrMessageTooBig) here instead?

meAmitPatil: Good catch — fixed in b9408c4

pavitrabhalla (approved):

PR #246, opened 2026-07-20
secretsproxy: re-mint cached leaf certs before they expire

meAmitPatil:
The proxy mints per-SNI TLS leaf certificates with a 72h validity and caches them in an LRU keyed by SNI. The cache only evicted under capacity pressure, never on age, so a frequently-used hostname's cached leaf was served indefinitely — including long after it expired. Clients then failed the TLS handshake with `CERT_HAS_EXPIRED` roughly 72h after the proxy last minted that host's leaf, and stayed broken until the cache was flushed (a proxy restart).

## Fix
- `leafCache.get` now treats a leaf within `leafRenewBefore` (1h) of its `NotAfter` as a miss, so the caller re-mints instead of serving a stale cert. The entry stores its expiry; the cache carries an injectable clock so minting and expiry read the same time source.
- Test mints a leaf, advances the clock into the renew window, and asserts a re-mint rather than a stale hit (fails without the fix).

## Note
This addresses expired *cached* leaves. A separate concern — a guest whose clock has drifted after repeated pause/resume could still misjudge a valid cert's window — is not covered here and is worth tracking on its own.

### DIFF
--- a/internal/secretsproxy/ca.go
+++ b/internal/secretsproxy/ca.go
@@ -260,6 +260,9 @@ func (c *leafCache) get(sni string) (*tls.Certificate, bool) {
 	}
 	entry := el.Value.(*leafCacheEntry)
 	if !c.now().Add(leafRenewBefore).Before(entry.notAfter) {
+		// Concurrent connections in the renew window each re-mint (set is
+		// last-writer-wins) until the first fresh cert lands — a handful of
+		// extra sub-ms signs once per 72h per SNI, not worth coalescing.
 		return nil, false
 	}
 	c.order.MoveToFront(el)
--- a/internal/secretsproxy/ca.go
+++ b/internal/secretsproxy/ca.go
@@ -32,6 +32,12 @@ type CA struct {
 const (
 	leafLifetime = 72 * time.Hour
 	rootLifetime = 10 * 365 * 24 * time.Hour
+
+	// leafRenewBefore re-mints a cached leaf this long before it expires, so a
+	// hot SNI that never falls out of the LRU can't be served a stale cert (and
+	// so a cert can't expire mid-connection). Comfortably larger than any single
+	// request, negligible waste against the 72h lifetime.
+	leafRenewBefore = 1 * time.Hour
 )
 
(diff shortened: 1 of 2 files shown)


internal/secretsproxy/ca.go
```diff
@@ -232,29 +243,40 @@ func newLeafCache(capacity int) *leafCache {
...
 func (c *leafCache) get(sni string) (*tls.Certificate, bool) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	el, ok := c.items[sni]
 	if !ok {
 		return nil, false
 	}
+	entry := el.Value.(*leafCacheEntry)
+	if !c.now().Add(leafRenewBefore).Before(entry.notAfter) {
+		return nil, false
```
pavitrabhalla, line 266: Once a hot SNI enters the renew window, every concurrent connection for that SNI sees a miss here and independently mints+signs its own leaf (cache.set is last-writer-wins). Each connection still gets a valid cert so it's not a correctness issue, just N redundant keygen+sign ops instead of 1. Given #229 just added singleflight for concurrent API-key lookups, is this worth coalescing the same way, or is the redundant-mint cost negligible enough to skip?

pavitrabhalla (approved):
Fix looks correct: verified the renew-before-expiry boundary math by hand and it matches the test's three clock positions, and `go build` + `go test ./internal/secretsproxy/...` pass on this branch (including the new regression test, which does fail without the fix per a quick check of the logic). No compile-time cache.set() call sites were missed in the test file. One non-blocking observation inline.

internal/secretsproxy/ca.go
meAmitPatil: Good catch, but I think it's cheap enough to leave. The window is really just the mint duration (~1ms), not the renew hour — as soon as the first mint lands it sets a +72h cert, so everyone after that hits the cache. And it's a local ECDSA keygen+sign, not a DB round-trip like #229, happening maybe once every 72h per SNI. I'll drop a one-line comment so it doesn't look accidental.

PR #248, opened 2026-07-21
Add PR loop GitHub Actions workflow

mohamedsobhi777:
## What changed
- Add `.github/workflows/loop-pr-loop.yml` for PR-loop review ticks on opened, synchronized, and reopened pull requests.
- Add manual dispatch support with least-privilege permissions and serialized execution.

## Why
The installed PR loop needs a repository workflow so it runs automatically for PR activity.

## Validation
- `git diff --check origin/main...HEAD` passed.
- Confirmed the diff contains only the workflow file.
- `actionlint` and `yamllint` were not installed in the workspace.

### DIFF
--- a/.github/workflows/loop-pr-loop.yml
+++ b/.github/workflows/loop-pr-loop.yml
@@ -1,0 +1,35 @@
+# Installed by `superserve-loops add pr-loop`. Runs on every PR code change
+# (a commit pushed to a PR) — no idle cron. One warm-sandbox tick per event, then it sleeps.
+# Note: `pull_request` from a forked repo gets a read-only token, so reviews on fork PRs need
+# the cross-repo PAT path (see below) or `pull_request_target` (the loop runs PR code only in
+# the sandbox, never on the runner). Add `schedule:` back if you also want a safety-net sweep.
+name: loop-pr-loop
+on:
+  pull_request:
+    types: [opened, synchronize, reopened] # synchronize = new commits pushed to the PR
+  workflow_dispatch: {}
+concurrency:
+  # Serialize per repo: all PR reviews share one warm sandbox, so don't run two at once.
+  group: loop-pr-loop
+  cancel-in-progress: false
+# Least privilege: clone the repo + post the review and ready-to-merge / needs-human labels.
+permissions:
+  contents: read
+  pull-requests: write
+jobs:
+  tick:
+    # Fork PRs can't read repo secrets or write with the built-in token — skip them.
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    steps:
+      - uses: oven-sh/setup-bun@v2
+      # Runs the PUBLISHED loop — no loop source is vendored into this repo, and no repo
+      # checkout is needed (the sandbox clones the target repo itself). `@stable` is a
+      # Superserve-gated channel, so blessed updates roll out here without editing this file.
+      # --pr focuses the tick on the changed PR; empty on manual dispatch → sweep all.
+      - run: bunx @superserve/loops@stable run pr-loop --repo "${{ github.repository }}" --pr "${{ github.event.pull_request.number }}" --once
+        env:
+          SUPERSERVE_API_KEY: ${{ secrets.SUPERSERVE_API_KEY }}
+          SUPERSERVE_CLAUDE_SECRET: claude-oauth
+          # Reviews post as github-actions[bot] via the workflow's built-in token.
+          GITHUB_TOKEN: ${{ github.token }}
--- a/.github/workflows/loop-pr-loop.yml
+++ b/.github/workflows/loop-pr-loop.yml
@@ -21,15 +21,29 @@ jobs:
     # Fork PRs can't read repo secrets or write with the built-in token — skip them.
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


awille-ai (commented):
I reviewed this against the failure we just observed in `superserve-ai/superserve#295`, where `@superserve/loops@stable` reached `[pr-loop] reviewing PR ...` and then failed in the SDK's `connect()` call with `ServerError`, HTTP 500, `internal_error`.

The workflow is otherwise clean: least-privilege permissions, serialized execution, no checkout required, and the fork guard is appropriate. I found two reliability issues that should be addressed before treating this as production-solid:

1. The loop invocation has no retry/backoff around transient Superserve API failures. A single backend 500 currently turns the entire PR check red even though the repository CI and PR code may be healthy. Either the loops CLI should retry `connect()` internally for retryable 5xx/network failures, or this workflow should wrap the command with bounded exponential backoff. The retry should remain bounded and preserve the final non-zero exit if all attempts fail.

2. On `workflow_dispatch`, `github.event.pull_request.number` is absent, but the command still emits `--pr ""`. The comment says an empty value means “sweep all,” but that depends on the CLI parser accepting an explicitly empty option value. It would be safer to branch the command: include `--pr` only for `pull_request` events, and omit the flag entirely for manual dispatch.

I would also add a small comment clarifying that `SUPERSERVE_CLAUDE_SECRET: claude-oauth` is a remote secret identifier rather than the credential itself, assuming that is the intended contract.

These are reliability/operability concerns rather than security blockers, but the missing retry is directly implicated by a real failure in another repository.

.github/workflows/loop-pr-loop.yml
```diff
@@ -0,0 +1,35 @@
...
+    # Fork PRs can't read repo secrets or write with the built-in token — skip them.
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    steps:
+      - uses: oven-sh/setup-bun@v2
+      # Runs the PUBLISHED loop — no loop source is vendored into this repo, and no repo
+      # checkout is needed (the sandbox clones the target repo itself). `@stable` is a
+      # Superserve-gated channel, so blessed updates roll out here without editing this file.
+      # --pr focuses the tick on the changed PR; empty on manual dispatch → sweep all.
+      - run: bunx @superserve/loops@stable run pr-loop --repo "${{ github.repository }}" --pr "${{ github.event.pull_request.number }}" --once
```
awille-ai: This command currently has no resilience to transient Superserve API failures. In `superserve-ai/superserve#295`, the loop failed immediately in the SDK `connect()` call with HTTP 500 / `internal_error`, while the repository's normal CI passed. Please add bounded retry with exponential backoff for retryable 5xx/network failures, either inside the loops CLI (preferred, so every installation benefits) or around this invocation. It should still fail after the final attempt so persistent outages remain visible.

awille-ai: For `workflow_dispatch`, `github.event.pull_request.number` is absent, so this produces an explicit `--pr ""`. The comment says manual dispatch should sweep all PRs, but that relies on the CLI treating an empty option value exactly like an omitted option. Please use separate event-specific steps/commands so `--pr` is omitted entirely on manual dispatch.

awille-ai (changes requested):
Re-reviewed this against the hardened generator/workflow now implemented in superserve#298 and the updated installer in website#50.

This sandbox workflow is still on the older shape and should be regenerated or updated before merge:

1. `workflow_dispatch` still passes `--pr "${{ github.event.pull_request.number }}"`, which becomes an explicit empty value outside a pull-request event. Use separate event-specific steps and omit `--pr` entirely for manual sweeps.
2. The job has no `timeout-minutes`, so a stuck package download, API request, or sandbox operation can occupy the serialized repository-wide concurrency slot indefinitely. Match the hardened workflow's 30-minute job bound.
3. Clarify that `SUPERSERVE_CLAUDE_SECRET: claude-oauth` is the name of a remote Superserve secret rather than the Claude credential itself.

The transient `Sandbox.connect()` retry belongs in the published loops package and is addressed by superserve#298; this workflow will inherit that once a fixed package is released to `@stable`. The three items above are checked-in workflow behavior and will not be fixed by publishing the package.

The permission model, same-repository fork guard, serialization, and lack of attacker-controlled shell interpolation otherwise look appropriate. Current CI, Terraform Checks, and the loop workflow are green.

awille-ai (approved):
Re-reviewed the latest head. The previously requested workflow hardening is now present:

- pull-request and manual-dispatch runs use separate steps;
- manual dispatch omits `--pr` entirely;
- the job has a 30-minute timeout, preventing a stuck run from holding the serialized repository-wide concurrency slot indefinitely;
- `SUPERSERVE_CLAUDE_SECRET` is explicitly documented as the name of a remote Superserve secret, not the credential itself.

The transient warm-sandbox `connect()` retry is correctly implemented in the shared loops package in superserve#298 rather than duplicated in every installed workflow. Once that release is promoted to `@stable`, this workflow will inherit the bounded retry behavior without another repository change.

The workflow otherwise remains appropriately constrained: least-privilege permissions (`contents: read`, `pull-requests: write`), same-repository fork guard before secrets are exposed, no checkout of PR code on the runner, serialized execution, and no attacker-controlled PR text interpolated into the shell command.

CI, Terraform Checks, and the loop workflow are green. The remaining visible inline threads are outdated and refer to code that has been replaced by the latest commit.

Approved, with the operational dependency that superserve#298 be merged and released to the `stable` channel before relying on transient connection retry.

PR #249, opened 2026-07-21
builder: enable a small guest swap file in the template init

meAmitPatil:
Guests previously ran with no swap, so a memory spike past the RAM ceiling faulted the process (SIGBUS/SIGSEGV) instead of paging out. The template init now creates and enables a swap file at build time, sized as RAM/4 (capped at 4 GiB) so it scales with the guest.

- Best-effort: if fallocate/mkswap/swapon are unavailable or fail, boot continues without swap.
- Skipped when the rootfs lacks room for the swap file plus headroom, so small-disk templates keep their space.
- The swap file and swap-on state are frozen into the template snapshot, so create/pause/resume do no extra work.
- `build.meta.json` records `swap_mode` so snapshot consumers know how a template was built.

Applies to templates built after this change.

Validated on staging: swap comes up correctly and survives pause/resume, create/pause/resume latency is unchanged, and a memory-heavy workload that previously crashed now passes reliably.

### DIFF
--- a/.github/workflows/deploy-otel-collector.yml
+++ b/.github/workflows/deploy-otel-collector.yml
@@ -23,6 +23,7 @@ on:
 
 jobs:
   deploy-staging:
+    if: github.event_name == 'push' || github.event.inputs.environment == 'staging'
     runs-on: ubuntu-latest
     environment: staging
     permissions:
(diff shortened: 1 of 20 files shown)


awille-ai (approved):
Reviewed with the usual correctness/security/rollout bar. The current revision cleans up a failed allocation path, applies the intended floor/cap, and CI is green. I don't see a merge blocker.

Operationally, this changes the failure mode from guest OOM/process failure to potentially heavy disk I/O and up to 4 GiB of sandbox-local swap consumption. That is the intended tradeoff, but after rollout I would watch host disk latency, delta growth, and swap-in/swap-out rates so one memory-heavy workload cannot create unexpected host-wide I/O pressure.

PR #253, opened 2026-07-22
vmd: abandon warm network slots on shutdown and adopt them at boot

meAmitPatil:
## Problem

Stopping vmd tears down the warm network pool slot by slot: \`Pool.Stop\` waits for every in-flight recycle verification, then serially deletes each pooled namespace. On hosts with large fleets this holds the old process's exit for tens of seconds — and since the daemon's socket only hands over to the new process after exit, that entire window is downtime for creates/resumes. The control plane's Unavailable-retry budget covers a normal restart, not a drain of that length, so requests landing in the window fail. The new process then also starts with an empty pool and refills it from scratch.

## Fix

Make shutdown crash-equivalent, behind \`VMD_NET_POOL_ADOPT\` (default off):

- **Shutdown abandons instead of tearing down.** \`Stop\` leaves warm slots in the kernel and returns near-instantly (bounded 2s quiesce). Every stop-path arm that used to clean up a slot now routes through \`stopSlot\`, which abandons under the flag and preserves legacy teardown without it.
- **Boot adopts instead of sweeping.** Slot identity is fully derivable from the index (\`ns-N\`/\`veth-N\`/IPs/MAC), so nothing needs persisting: after startup slot reservation, \`AdoptOrphanSlots\` claims every unowned namespace in one locked pass (so the refill loop can never build over a candidate), validates each one (host veth + in-namespace peer present, nftables handle re-attached, host route healed — validation doubles as the layout/version check), and feeds survivors through the existing \`verifyAndRecycle\` path with a **forced tap rebuild** (unknown provenance; same reasoning as the tap-EBUSY hardening). Failures are torn down. Bounded workers; never blocks the readiness gate.
- Graceful stop and crash now leave the host in the same state and share one recovery path — every deploy exercises crash recovery.
- Stray host-side veths (namespace gone, slot unowned) are still swept after adoption; mid-build slots are protected by the ownership check.

Net effect: restart unavailability drops to roughly the daemon's startup time, and the new process starts with a warm, verified pool instead of a cold one.

## Safety

- Flag off = byte-for-byte legacy behavior (boot sweep + teardown Stop).
- Adopted slots pass the same occupancy verification as recycled slots (stray processes killed, namespace confirmed empty, poisoned slots torn down).
- Adoption only ever claims indexes with no owner, after record reservation — a live VM's slot can never become a pool candidate.

## Tests

- Abandon-on-stop leaves parked and pooled slots owned and intact; legacy stop still tears down.
- Adoption: valid orphan becomes claimable with a forced tap rebuild; invalid orphan is torn down and its index released; owned/unparseable namespaces are skipped and the allocator high-water mark advances.

### DIFF
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -564,7 +564,7 @@ func main() {
 	// Reserve slots held by existing VMs (so the pool can't hand out a colliding
 	// one) and sweep leaked namespaces. The per-VM reattach runs in the background
 	// below; VMs it hasn't reached are loaded on-demand on first request.
-	mgr.ReserveStartupSlots(ctx)
+	slotsReserved := mgr.ReserveStartupSlots(ctx)
 	adoptNetPool := envOrDefault("VMD_NET_POOL_ADOPT", "false") == "true"
 	if !adoptNetPool {
 		// Under adoption, orphan namespaces are warm-pool candidates instead
@@ -600,7 +600,8 @@ func main() {
 		AbandonOnStop:     adoptNetPool,
 	})
 	lc.addCloser("network pool", func(_ context.Context) error { netPool.Stop(); return nil })
-	if adoptNetPool {
+	switch {
+	case adoptNetPool && slotsReserved:
 		// Adopt the slots the previous run abandoned (or crashed out of) in
 		// the background: the pool starts warm within seconds instead of
 		// refilling from scratch, and boot never blocks on the pass.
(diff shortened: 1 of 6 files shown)


pavitrabhalla (approved):
Nice change — careful handling of the ownership handoff and firewall rebuild.

One minor race worth flagging: between `StartPool` starting `refillLoop` and the background `AdoptOrphanSlots` goroutine actually running, `refillLoop` can walk past orphaned indices via `claimSlotIndex`'s `nsExists` discard, advancing `nextSlot` past them before adoption claims them. Not a correctness bug — `claimOrphanSlots` scans the netns directory directly, not gated by `nextSlot` — but on a congested boot it means some orphans sit unclaimed longer and the high-water mark ends up higher than necessary.

Did you consider running `AdoptOrphanSlots` before `StartPool` kicks off the refill loop, or is blocking on that (vs. the readiness gate) not worth it here?

PR #257, opened 2026-07-22
Add Multi-Region Host Monitoring and Terraform Safeguards

awille-ai:
## Summary

- Add Terraform-managed CPU alert policies for monitored Compute Engine hosts across production regions.
- Route alerts through existing Cloud Monitoring notification channels supplied by the production GitHub Environment.
- Pass notification channels consistently through plan and production apply workflows, with validation for full channel resource names.
- Add sanitized Terraform plan evidence and document thresholds, response steps, rollback guidance, and verification.
- Split unrestricted SSH firewall rules into separate IPv4 and IPv6 resources while preserving the existing IPv4 rule state.
- Keep generated firewall names within Google Cloud's length limit while retaining deterministic uniqueness for long names.

## Validation

- `terraform fmt -check -recursive infra`
- `bash -n scripts/sanitize-terraform-plan.sh`
- `git diff --check`

No production Terraform apply is performed by this change.

### DIFF
--- a/.github/workflows/terraform-plans.yml
+++ b/.github/workflows/terraform-plans.yml
@@ -25,15 +25,23 @@ jobs:
         include:
           - name: staging-us-central1
             dir: infra/envs/staging/us-central1
+            environment: staging
           - name: production-us-central1
             dir: infra/envs/production/us-central1
+            environment: production
           - name: production-us-west2
             dir: infra/envs/production/us-west2
+            environment: production
           - name: production-us-east4
             dir: infra/envs/production/us-east4
+            environment: production
+    environment: ${{ matrix.environment }}
     env:
       TF_IN_AUTOMATION: "true"
       HAS_GCP_AUTH: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER != '' && secrets.GCP_SERVICE_ACCOUNT != '' }}
+      # Existing Monitoring channel resource names are non-secret references;
+      # the channel's Slack credential remains owned by Cloud Monitoring.
+      TF_VAR_notification_channel_ids: ${{ vars.TF_VAR_NOTIFICATION_CHANNEL_IDS || '[]' }}
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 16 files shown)


.github/workflows/terraform-plans.yml
```diff
@@ -72,6 +81,37 @@ jobs:
...
+                  exit 1
+                fi
+                ;;
+              projects/*/notificationChannels/*)
+                if [[ ! "$channel_ids" =~ ^projects/[a-z][a-z0-9-]*/notificationChannels/[0-9]+$ ]]; then
+                  echo "::error title=Invalid notification channels::TF_VAR_NOTIFICATION_CHANNEL_IDS must name a valid Cloud Monitoring channel."
+                  echo "Plan stopped: notification channels must use projects/PROJECT_ID/notificationChannels/NUMERIC_ID." >> "$summary_file"
+                  exit 1
+                fi
+                export TF_VAR_notification_channel_ids="[\"$channel_ids\"]"
```
pavitrabhalla, line 105: This normalizes a bare channel string into a list before `terraform plan` runs. `terraform-cd.yml` and `terraform-rollout-production.yml` inject `TF_VAR_notification_channel_ids` from the same variable with no equivalent normalization. If the variable is ever set to a bare string, the PR plan check would pass here while the apply workflows pass the raw string straight to Terraform, which can't parse it as `list(string)`. Should the normalization live in one shared place, or does something on the apply side already guard against this?

.github/workflows/terraform-plans.yml
```diff
@@ -92,9 +132,9 @@ jobs:
             echo "Changes detected."
             echo "Changes detected." >> "$summary_file"
             echo >> "$summary_file"
-            echo "Plan output:" >> "$summary_file"
+            echo "Sanitized plan summary:" >> "$summary_file"
             echo '```' >> "$summary_file"
-            terraform show -no-color tfplan | head -n 200 >> "$summary_file" || true
+            "${GITHUB_WORKSPACE}/scripts/sanitize-terraform-plan.sh" tfplan >> "$summary_file"
```
pavitrabhalla, line 137: The previous line had `|| true` after rendering the plan summary; this call doesn't, and the step runs under `set -euo pipefail`. If `sanitize-terraform-plan.sh` fails for an unrelated reason (missing `jq`, a plan-read error), that now hard-fails the whole check even though `terraform plan` itself succeeded. Intentional?

scripts/sanitize-terraform-plan.sh
```diff
@@ -0,0 +1,25 @@
...
+if ! command -v jq >/dev/null 2>&1; then
+  echo "jq is required to produce a sanitized Terraform plan summary" >&2
+  exit 1
+fi
+
+echo "Sanitized Terraform plan summary"
+echo "Planned values, prior state, configuration, and variable values are omitted."
+echo
+terraform show -json "$plan_file" | jq -r '
+  .resource_changes[]?
```
pavitrabhalla, line 20: `resource_changes` includes an entry for every resource, including unchanged ones tagged `no-op` (checked this against a real plan). There's no filter on `.change.actions` here, so on an env with more resources the sanitized summary would be dominated by no-op lines, burying the actual create/update/delete changes reviewers need before approving a production apply. Worth adding `select(.change.actions != ["no-op"])`?

infra/README.md
```diff
@@ -5,4 +5,53 @@ Terraform skeleton layout for Superserve GCP adoption.
...
+instance name. If an existing policy covers a host, import or correct that
+policy instead of creating a duplicate. The owner must also verify the
+configured channel with `gcloud monitoring channels describe CHANNEL_ID`; this
+checks channel ownership and delivery configuration without exposing a Slack
+webhook credential. The reusable module contains one policy per instance only
+because the instance-ID filter keeps each host's Vanta evidence and response
+path unambiguous.
+
+For an existing matching policy, use its full Monitoring policy name with
+`terraform import module.observability.google_monitoring_alert_policy.compute_instance_cpu[\"sandbox_host\"] POLICY_NAME`
```
pavitrabhalla, line 36: This command fails immediately in zsh — the unescaped `[...]` gets parsed as a glob character class (`no matches found`) before terraform ever runs. Can the resource address be single-quoted here?

pavitrabhalla:
Confirmed this resolves the #254 firewall failure — the `for_each` split into ipv4/ipv6 rules plus the `moved` block fixes the `Error 400: Mixture of IPv4 and IPv6 in the same rule` that stranded the apply.

One remaining gap for a clean apply, though: the failing #254 run also hit `Error 403: Required 'compute.subnetworks.update'` on `rayai-dev` (staging) when enabling flow logs. This PR enables flow logs but doesn't grant that permission, and since the CD chain applies staging first, it'll 403 there and skip prod again — same shape as #254. Prod's SA already has networkAdmin (granted during the us-east4 host migration); staging (rayai-dev) still needs `compute.subnetworks.update` / `roles/compute.networkAdmin` before this applies cleanly.

pavitrabhalla (approved):
Please see the comments I left earlier, but they are all non-blocking.

PR #258, opened 2026-07-22
deploy-api: deploy the use4 primary cell on every merge

pavitrabhalla:
## Summary

`deploy-api.yml` never learned about the us-east4 (use4) cell. The `deploy-production` job deploys `CLOUD_RUN_SERVICE` (central) and a conditional `CLOUD_RUN_SERVICE_USW` (west) block — but nothing for use4. Since the host migration, **api.superserve.ai routes to use4**, so every API merge has been deploying central + west while the *primary* cell silently kept running a hand-deployed image from the migration window. Same "host-local config didn't migrate" class as the earlier deploy-proxy/vmd gaps.

## What this does

- Adds a `use4` cell block to `deploy-production` mirroring the usw pattern (deploy → route-to-latest → verify), gated on `vars.CLOUD_RUN_SERVICE_USE4 != ''`, reusing the SHA-tagged image already pushed to Artifact Registry.
- Unlike usw, use4 **has** a public invoker binding (it serves api.superserve.ai), so its verify step includes a `/health` probe.
- Repo vars set: `CLOUD_RUN_SERVICE_USE4=superserve-api-use4`, `GCP_REGION_USE4=us-east4`.

## Already done out-of-band (not in this PR)

- Un-staled use4 to current main: deployed `controlplane:<main-sha>` (SHA-pinned, matching central/west — it had been floating on `:latest`), verified `/health=200` at 100% traffic.

## Follow-up (not here)

- use4 is added as a *cell* step; central is still the workflow's "primary" block. When central is decommissioned (Phase E), the primary block should become use4 and the central steps removed. Until then a central hiccup would block the use4 step — acceptable short-term since central is idle.

## Testing

- YAML parses; workflow steps mirror the verified usw block. Vars set on the repo.

### DIFF
--- a/.github/workflows/deploy-api.yml
+++ b/.github/workflows/deploy-api.yml
@@ -306,13 +306,53 @@ jobs:
           fi
           curl -sf "$(echo "$DESC" | jq -r '.status.url')/health"
 
+      # The use4 cell is the primary production cell — api.superserve.ai routes
+      # here — so it deploys BEFORE the optional secondary cells below: a
+      # secondary-cell (usw) failure must not stop the job and leave the public
+      # primary on the old image. Same step-based pattern; reuses the SHA-tagged
+      # image pushed above. Steps skip until CLOUD_RUN_SERVICE_USE4 is set. use4
+      # has a public invoker binding, so /health is a meaningful probe.
+      - name: Deploy use4 cell to Cloud Run
+        if: vars.CLOUD_RUN_SERVICE_USE4 != ''
+        run: |
+          gcloud run services update ${{ vars.CLOUD_RUN_SERVICE_USE4 }} \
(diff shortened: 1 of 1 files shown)


meAmitPatil (approved):

awille-ai (approved):
Reviewed the current head. The earlier ordering concerns are fixed: use4 now deploys immediately after the image push, before the idle us-central1 cell and the usw secondary cell, so failures in those cells cannot leave the public primary stale.

This PR only changes the deployment workflow; there is no Terraform plan to inspect for this head. CI and Terraform Checks are green. The deploy/route/verify sequence for use4 matches the existing cell pattern, includes an explicit 100%-traffic check, and probes /health because the service is publicly invokable.

Approved.

PR #255, opened 2026-07-22
Codify us-east4 host migration resources (import-only)

pavitrabhalla:
## Summary

Reconciles Terraform with the production resources that were created imperatively during the us-central1 → us-east4 primary-host migration. **Import-only**: this models the live resources so state matches reality — no create/replace is intended.

## What's included

- CP `allUsers → roles/run.invoker` as a module resource (`allow_public_invoker`, default true) — covers all three cells
- New `cloud-run-cert-lb` module for the `api.superserve.ai` external HTTPS LB: serverless NEG → backend → url-map, Certificate-Manager dns-auth/cert/map, HTTPS proxy, global IP, forwarding rule
- OTEL env for the us-east4 control plane
- `us-east4` entry in `deploy/environments.yaml`
- `unbound` host bootstrap (startup-script) + `deploy-vmd.py` env plumbing (`CONTROL_PLANE_URL` / `INTERNAL_API_TOKEN` / `DAEMON_AUTH_TOKEN`, sourced from CI — no secret values hardcoded)

## Technical decisions / tradeoffs

- Written to be adopted via `terraform import`; **each import's plan must be a no-op before any apply.**
- The unbound bootstrap does **not** manage the guest DNS redirect — vmd owns `SANDBOX_DNS_REDIRECT` (`internal/network/hostfw.go`), and managing it from two places conflicts on a rebuild. unbound config + DoT upstreams match the live host.
- The public-invoker binding is default-on; the existing bindings on all three cells must be imported before the next apply in those envs, or apply will (idempotently) re-assert them.

## Testing

- `terraform validate` passes in `infra/envs/production/us-east4`; `terraform fmt` clean; `deploy-vmd.py` parses. **Not applied.**

## Import commands

From `infra/envs/production/us-east4`:
```
terraform import 'module.api.google_cloud_run_v2_service_iam_member.public_invoker[0]' 'projects/rayai-prod/locations/us-east4/services/superserve-api-use4 roles/run.invoker allUsers'
terraform import module.api_cert_lb.google_compute_region_network_endpoint_group.cloud_run projects/rayai-prod/regions/us-east4/networkEndpointGroups/superserve-api-use4-neg
terraform import module.api_cert_lb.google_compute_backend_service.cloud_run projects/rayai-prod/global/backendServices/superserve-api-backend-use4
terraform import module.api_cert_lb.google_compute_url_map.this projects/rayai-prod/global/urlMaps/superserve-api-url-map
[...]

PR #259, opened 2026-07-22
CD: cover the use4 primary cell (deploy-vmd, otel-collector, smoke-test)

pavitrabhalla:
## Summary

Closes the remaining "host-local config didn't migrate" CD gaps for the us-east4 (use4) primary cell, found in a post-migration audit. `deploy-vmd`, `deploy-otel-collector`, and `smoke-test-region` never learned about use4 (only central + a conditional usw).

## What this does

- **`deploy-vmd.yml` (critical):** add a use4 cell step mirroring usw. This was the root cause — the east host got **no host-agent deploys** (vmd/boxd/secretsproxy/template-builder/fc-cleanup), which is why it had to be hand-staged during the migration and why secretsproxy was missing.
- **`deploy-otel-collector.yml`:** add a use4 deploy step (scoped to us-east4) so the primary host's collector is CD-managed.
- **`smoke-test-region.sh`:** fold `production/us-east4` into the shared use-cell case (api/sandbox.superserve.ai).

## Verified NOT gaps (no change)

- `cd.yml` migrations & `seed-templates.yml` per-cell steps use `DATABASE_URL_USWEST`; **use4 shares central's DB**, so the primary invocation already covers it.
- `terraform-rollout-production.yml` is a manual west-only helper; use4 infra is covered by the automated `terraform-cd.yml`.

## Prereqs verified

- East host **is discoverable**: labels `component=vmd, environment=production, region=us-east4`, status RUNNING — matches `deploy-vmd.py`'s filter (`labels.component=vmd AND status=RUNNING`, zone-scoped) and the otel `VMD_FILTER`.
- Repo vars `CLOUD_RUN_SERVICE_USE4` / `GCP_REGION_USE4` are set.

## Follow-ups (noted, not in this PR)

- usw otel collector is still uncovered by `deploy-otel-collector.yml` (pre-existing gap).
- East host label `sandbox_status=provisioning` is stale (it's live-serving) — cosmetic.
- East `otelopscol` SA lacks `logging.logEntries.create` — separate IAM fix (ties to host observability).
- At decommission, each workflow's **primary** block should swap central→use4.

## Testing

- Both workflow YAMLs parse; `smoke-test-region.sh` passes `bash -n`.

### DIFF
--- a/.github/workflows/deploy-otel-collector.yml
+++ b/.github/workflows/deploy-otel-collector.yml
@@ -79,3 +79,17 @@ jobs:
           OTEL_COLLECTOR_PROJECT: ${{ vars.GCP_PROJECT }}
         run: |
           python3 .github/workflows/scripts/deploy-otel-collector.py
+
+      # The use4 cell host is the primary production host (api.superserve.ai).
+      # Same collector deploy scoped to us-east4; skipped until
+      # CLOUD_RUN_SERVICE_USE4 is set. (usw remains uncovered here — pre-existing
+      # gap, tracked separately.)
+      - name: Deploy OTEL Collector to use4 cell VMD instances
+        if: vars.CLOUD_RUN_SERVICE_USE4 != ''
+        env:
+          GCP_PROJECT: ${{ vars.GCP_PROJECT }}
+          GCP_REGION: us-east4
+          VMD_FILTER: labels.component=vmd AND labels.environment=production AND labels.region=us-east4
+          OTEL_COLLECTOR_PROJECT: ${{ vars.GCP_PROJECT }}
+        run: |
+          python3 .github/workflows/scripts/deploy-otel-collector.py
--- a/.github/workflows/deploy-vmd.yml
+++ b/.github/workflows/deploy-vmd.yml
@@ -124,9 +124,29 @@ jobs:
       - name: Set up Cloud SDK
         uses: google-github-actions/setup-gcloud@v2
 
-      # GCP_REGION keeps this fan-out on the primary region's hosts only;
-      # the usw cell host shares the project and label but deploys in the
-      # step below, after the primary succeeds.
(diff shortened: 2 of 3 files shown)


pavitrabhalla:
Re: the `component=vmd` discovery-label P1 — this is resolved by #255, not a change needed here.

#255 declares `component=vmd` + `sandbox_role=vmd` in the us-east4 host Terraform (`infra/envs/production/us-east4/main.tf`), so a `terraform apply` keeps the label instead of stripping it. The OTEL step in this PR filters on `component=vmd AND environment=production AND region=us-east4` and is gated on `CLOUD_RUN_SERVICE_USE4`, so the filter stays satisfiable once #255 lands.

Only constraint is merge order: land #255 before the next infra `terraform apply`. That's naturally satisfied — infra applies are currently paused behind the #254 firewall fix (#257), and #255 is queued right after it. No code change here.

awille-ai (changes requested):
Blocking on rollout ordering for the primary use4 host.

The new use4 VMD deploy step is currently last in the production job, after both the central and usw deploys. Because these steps run sequentially and fail-fast, any failure in the idle central host or secondary west host prevents the public primary use4 host from receiving the new vmd/boxd/secretsproxy/template-builder/fc-cleanup bundle. That recreates the exact primary-cell drift this PR is intended to close.

Please move the use4 VMD deploy immediately after authentication/setup and before central/usw, or otherwise decouple it from failures in those cells. The same ordering principle should apply to the OTEL workflow: use4 is the primary host, so its collector deploy should run before the legacy central step rather than being blocked by it.

The earlier discovery-label concern appears covered by PR #255's Terraform changes, assuming #255 lands first as documented. This PR does not change Terraform, so there is no Terraform plan artifact to inspect. CI and Terraform Checks are green.

awille-ai (approved):
All comments addressed

PR #261, opened 2026-07-23
vmd: batch the launcher prune unmounts and log pre-build legacy launches

meAmitPatil:
The launcher-namespace build pruned /run/netns with one `umount` fork per netns mount, making the build take minutes on hosts with a large live fleet. Until the build finishes, every launch silently falls back to the legacy `ip netns exec` path, which copies the full host mount table — so a burst of concurrent launches landing in that window serializes on the kernel's mount lock and can exceed the firecracker socket-wait budget.

Two changes:
- Batch the prune's unmounts through `xargs`, turning per-mount forks into a handful of execs. The build window shrinks from minutes to seconds regardless of fleet size.
- Log a warning when a launch takes the legacy path because the pin is not built yet. This window was previously invisible: the existing warning only covered a pin that was built and then went missing.

No flag semantics change; with launcher mode disabled the behavior is byte-identical.

### DIFF
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -574,9 +574,10 @@ func main() {
 
 	// Launcher launch path, enabled per host via VMD_LAUNCH_VIA_LAUNCHER_NS.
 	if launchViaLauncherNS {
-		// Build the pruned launcher mount namespace in the background: the prune is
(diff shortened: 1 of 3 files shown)


pavitrabhalla (approved):
Thanks for this fix!

PR #262, opened 2026-07-23
cd: use4 collector first + deploy-vmd self-trigger (follow-up to #259)

pavitrabhalla:
Addresses the two follow-up review comments on #259.

- **deploy-otel-collector.yml**: reorder the use4 (primary host) collector deploy ahead of the idle us-central1 step — a legacy-only failure was able to stop the job before the public host's collector updated. Mirrors the deploy-vmd/deploy-api use4-first ordering.
- **deploy-vmd.yml**: add `.github/workflows/deploy-vmd.yml` and `.github/workflows/scripts/deploy-vmd.py` to `push.paths`. Without them the workflow doesn't run when only the deploy tooling changes, so #259's new use4 host step never executed on merge — the use4 host is still on migration-era binaries. Adding these means **merging this PR self-triggers a vmd deploy that finally covers the use4 host** (deploy-otel already self-triggers).

Workflow-only; no infra apply.

### DIFF
--- a/.github/workflows/deploy-otel-collector.yml
+++ b/.github/workflows/deploy-otel-collector.yml
@@ -71,25 +71,28 @@ jobs:
       - name: Set up Cloud SDK
         uses: google-github-actions/setup-gcloud@v2
 
-      - name: Deploy OTEL Collector to production VMD instances
-        env:
-          GCP_PROJECT: ${{ vars.GCP_PROJECT }}
-          GCP_REGION: us-central1
-          VMD_FILTER: labels.component=vmd AND labels.environment=production AND labels.region=us-central1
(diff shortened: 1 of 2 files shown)


awille-ai (approved):
Reviewed the current head. This cleanly addresses both follow-ups from #259:

- The use4 OTEL collector deploy now runs before the idle us-central1 step, so a legacy-host failure cannot block the public primary host's collector update.
- deploy-vmd now self-triggers when either the workflow or deploy script changes, preventing deployment logic from landing without actually running.

This PR is workflow-only and does not modify Terraform resources, so there is no Terraform plan artifact to inspect. CI and Terraform Checks are green.

Approved.

PR #263, opened 2026-07-23
observability: drop invalid notification_rate_limit from metric alert policy

pavitrabhalla:
**Unblocks the production terraform-cd chain.** After #257 merged, `Apply production/us-west2` fails:

```
Error 400: alertStrategy.notificationRateLimit ... only log-based alert policies may specify a notification rate limit
```

`module.observability.google_monitoring_alert_policy.compute_instance_cpu` is a **metric-threshold** policy, and `notification_rate_limit` is only valid on **log-based** policies. Remove it; keep `auto_close` (valid for metric policies).

Staging already applied clean (the #254 firewall + rayai-dev grant are resolved), so this is the only thing left blocking the prod apply.

### DIFF
--- a/infra/modules/observability/main.tf
+++ b/infra/modules/observability/main.tf
@@ -55,9 +55,10 @@ resource "google_monitoring_alert_policy" "compute_instance_cpu" {
   }
 
   alert_strategy {
-    notification_rate_limit {
-      period = "900s"
-    }
+    # notification_rate_limit is only valid on log-based alert policies; this is
+    # a metric-threshold policy, so GCP rejects it (Error 400: "only log-based
+    # alert policies may specify a notification rate limit"). auto_close is valid
+    # for metric policies and is kept.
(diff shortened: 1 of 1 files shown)


awille-ai (approved):
Reviewed the code change and the Terraform Plans workflow artifacts.

The change is correct: this is a metric-threshold alert policy, so removing `notification_rate_limit` while retaining `auto_close` fixes the GCP 400 without weakening the policy's incident auto-close behavior.

The production-us-west2 plan now validates successfully and shows only the expected creation of `module.observability.google_monitoring_alert_policy.compute_instance_cpu["sandbox_host"]`; the policy was not created by the previously failed apply. There are no west-region updates, replacements, or destroys.

The other environment summaries still contain pre-existing pending changes from the broader Terraform rollout, but this one-line fix does not introduce them. Terraform Checks and Terraform Plans are green; CI was still running at review time.

Approved.

PR #265, opened 2026-07-23
Decommission the us-central1 primary control plane + vmd host, and its deploy targets

pavitrabhalla:
Draft — the destructive prerequisites below are executed manually alongside the merge. api.superserve.ai now serves from us-east4; the us-central1 control plane is idle (min=0, ~0 traffic) and its vmd host is drained (the DB host row points at us-east4).

## Terraform — removes (region-local, from `infra/envs/production/us-central1`)
- `module.api` — the `superserve-api` Cloud Run control plane
- `module.sandbox_host` — the `superserve-vmd-prod` host
- the `sandbox_host` CPU-saturation alert (the host it watched is gone)
- the now-dangling `sandbox_host` output + `imports.tf`

## Terraform — keeps
- `module.network` (shared VPC), `module.iam` (api-runner + github-actions SAs used by use4/usw), and the shared runtime-secret IAM.
- `module.observability` **dashboards** — the Production Operations/Fleet/Hosts/Database/Collector boards are **fleet-wide** (they span every region) and only rooted in this state historically. Deleting them here would remove the whole fleet's monitoring, so the module stays with just the dashboards; only the per-host alert is dropped.

## CD — stop targeting us-central1 (addresses Codex P1: retired deploy targets)
- `terraform-cd.yml`: drop the `Deploy API production/us-central1` job (Cloud Run deploy + traffic + /health + region smoke against the deleted service); rewire us-east4 to depend on the us-central1 **infra** apply. The us-central1 infra apply stays (it manages the shared VPC/IAM/dashboards).
- `deploy-api.yml`: drop the us-central1 Cloud Run service deploy/route/verify steps. The us-central1 image push stays — it is the shared Artifact Registry repo use4/usw pull from.
- `deploy-vmd.yml` / `deploy-otel-collector.yml`: drop the us-central1 host fan-out / collector steps.
- `environments.yaml`: drop the us-central1 production region entry.

Staging us-central1 is untouched — only the production primary is retired.

## Deletion protection (addresses Codex P1)
- Host: `module.sandbox_host` has `prevent_destroy = true`.
- CP: the api module sets `deletion_protection = true` (though the live service currently reports it **off**).

Rather than rely on a plain destroy, both are removed from state and deleted out-of-band, so the CD apply never attempts a guarded destroy:
1. `terraform state rm module.sandbox_host module.api`
2. `gcloud compute instances delete superserve-vmd-prod --zone us-central1-c` and `gcloud run services delete superserve-api --region us-central1`
3. Merge → terraform-cd applies the trimmed config: destroys only the host CPU alert, updates the dashboards in-place (`0 add, 5 change, 1 destroy` once the modules are out of state).

The boot disk is archived as the rollback path; reversible until the host is deleted.

## Follow-up (separate PR)
Release the host's static IP + its now-orphaned firewall rules.

### DIFF
--- a/.github/workflows/deploy-api.yml
+++ b/.github/workflows/deploy-api.yml
@@ -271,13 +271,13 @@ jobs:
           docker push ${IMAGE}:${{ github.sha }}
 
       # use4 is the primary production cell — api.superserve.ai routes here — so
-      # it deploys FIRST, right after the image push, before the idle us-central1
-      # cell and the usw secondary cell. That way a transient failure updating or
-      # probing those cells can't stop the job and leave the public primary on
-      # the old image. Steps skip until CLOUD_RUN_SERVICE_USE4 is set; use4 has a
-      # public invoker binding, so /health is a meaningful probe.
+      # it deploys FIRST, right after the image push, before the usw secondary
+      # cell. That way a transient failure updating or probing usw can't stop the
+      # job and leave the public primary on the old image. These steps are
+      # unguarded: use4 is the permanent primary, so a missing CLOUD_RUN_SERVICE_USE4
+      # must fail the deploy rather than silently skip the only primary update.
+      # use4 has a public invoker binding, so /health is a meaningful probe.
       - name: Deploy use4 cell to Cloud Run
-        if: vars.CLOUD_RUN_SERVICE_USE4 != ''
         run: |
           gcloud run services update ${{ vars.CLOUD_RUN_SERVICE_USE4 }} \
             --image ${{ vars.GCP_REGION }}-docker.pkg.dev/${{ vars.GCP_PROJECT }}/superserve/controlplane:${{ github.sha }} \
@@ -285,7 +285,6 @@ jobs:
             --project ${{ vars.GCP_PROJECT }}
 
       - name: Route use4 cell traffic to the new revision
-        if: vars.CLOUD_RUN_SERVICE_USE4 != ''
         run: |
           SVC=${{ vars.CLOUD_RUN_SERVICE_USE4 }}
           ARGS="--region ${{ vars.GCP_REGION_USE4 }} --project ${{ vars.GCP_PROJECT }}"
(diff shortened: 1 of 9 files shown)


pavitrabhalla:
@codex review

pavitrabhalla:
Addressed the latest review:

**P2 (proxy fan-out)** — fixed. `deploy-proxy.yml`'s production job had a us-central1 primary step I'd missed; removed it and moved the use4 step ahead of usw so the primary host deploys first.

**P2 (use4 deploy on infra pushes)** — fixed. Added a `Deploy API production/us-east4` job to `terraform-cd.yml` (deploy + traffic + /health + `smoke-test-region.sh production us-east4`) in place of the removed us-central1 api job, so infra-inclusive pushes — which skip Deploy API via its `infra_changed` guard — still roll api.superserve.ai to the new image.

**P1 (Cloud Run deletion protection)** — handled the way your comment's second option suggests (remove/import state after an out-of-band decommission), and the same way the host's `prevent_destroy` is handled. Before merge, both are removed from state and deleted out-of-band:
`terraform state rm module.api module.sandbox_host` → `gcloud run services delete superserve-api` + `gcloud compute instances delete superserve-vmd-prod`. The post-merge `production-us-central1` apply then never plans a guarded destroy — it only drops the host CPU alert (`0 add, 5 change, 1 destroy`). The host's `prevent_destroy` (a literal in the shared module) can't be toggled per-instance, so state surgery is required regardless; doing the CP the same way keeps it deterministic rather than relying on refresh reconciling the stale `deletion_protection=true` in state.

@codex review

pavitrabhalla:
**P2 (primary deploy gated by secondaries)** — fixed. Branched `production-us-east4-infra`/`-api` directly off `staging-us-central1-api` instead of the `us-west2 → us-central1` chain, so the primary (api.superserve.ai) rolls independently and a usw/west2 deploy-or-smoke failure can't block it. us-east4 only reads the shared api-runner SA via a data source, so it needs no apply ordering against us-central1.

@codex review

pavitrabhalla:
**P2 (job-level gating on production env vars)** — fixed. Removed the job-level `if: vars.CLOUD_RUN_SERVICE_USE4 != ''` on `production-us-east4-api`. You're right that a job-level `if` is evaluated before the `production` environment is resolved, so an environment-scoped var reads as empty there and would skip the primary deploy on every infra push. use4 is the permanent primary now, so it matches the us-west2 api job, which carries no such guard.

**P2 (disable deletion protection before dropping module.api)** — this is the documented pre-merge sequence, and it's unavoidable regardless: the host module's `prevent_destroy = true` is a literal in the shared `sandbox-host` module and can't be toggled per-instance, so the merge already requires a `state rm` + out-of-band delete of the host. The CP is decommissioned the same way in the same step — `terraform state rm module.api module.sandbox_host` then `gcloud run services delete superserve-api` / `gcloud compute instances delete superserve-vmd-prod` — so the post-merge `production/us-central1` apply never plans a guarded destroy (it only drops the host CPU alert).

@codex review

awille-ai (changes requested):
Blocking on the decommission procedure being manual and non-atomic relative to the merge/apply.

I reviewed the latest head, CI, Terraform Checks, and the Terraform Plans workflow. The code correctly removes the retired us-central1 deploy targets and adds a use4 API deploy/smoke path to Terraform CD, addressing the earlier rollout blockers. However, the PR still requires operators to run `terraform state rm module.sandbox_host module.api` and then delete both resources out-of-band before merge.

That creates two unsafe failure windows:

1. After `state rm` but before deletion/merge, the production host and Cloud Run service are unmanaged. A failed or interrupted decommission leaves live resources outside Terraform with no automated reconciliation.
2. If the PR merges or Terraform CD runs before both manual steps complete, the central plan attempts guarded destroys (`prevent_destroy` / Cloud Run deletion protection) and the rollout fails. Conversely, deleting first while deploy workflows from main still target central can cause release failures until this PR lands.

The current plan artifact still reflects the pre-state-rm world, so its successful validation does not prove the documented post-removal `0 add, 5 change, 1 destroy` result.

Please make the transition operationally atomic and verifiable. A safe pattern would be a staged change/runbook with explicit gates: first remove central from all deploy targets and merge that independently; then disable/adopt deletion protections and apply; then delete through Terraform where possible, or run a reviewed one-shot workflow that performs state removal + deletion + immediate post-action plan and aborts on any mismatch. At minimum, attach a freshly generated plan from the exact post-`state rm` state and document rollback/verification commands and ownership for each step.

CI, Terraform Checks, and Terraform Plans are green, but I don't think this should merge while correctness depends on an untracked manual race around production state.

pavitrabhalla:
**P2 (require use4 before the production deploy can pass)** — fixed. Removed the `if: vars.CLOUD_RUN_SERVICE_USE4 != ''` bring-up guards from all three use4 steps in `deploy-api.yml` (deploy / route / verify), so a missing or renamed primary var now fails the deploy instead of letting the job go green after only pushing the image. The usw steps keep their guards — usw is a genuinely optional cell; use4 is the permanent primary.

@codex review

awille-ai (approved):

pavitrabhalla:
**P2 x2 (use4 must be mandatory in vmd + proxy deploys)** — fixed, same as the deploy-api change: removed the use4 bring-up guards from `deploy-vmd.yml` and `deploy-proxy.yml` so a missing/renamed primary var fails the deploy instead of letting a green run leave the primary host on stale binaries/proxy. usw stays guarded (optional cell). This completes the guard-consistency sweep across all four deploy workflows.

@codex review

PR #270, opened 2026-07-23
cd: codify CONTROL_PLANE_URL/DATABASE_URL/INTERNAL_API_TOKEN for vmd deploys

pavitrabhalla:
## Summary

`CONTROL_PLANE_URL`, `DATABASE_URL`, and `INTERNAL_API_TOKEN` in a vmd host's `/etc/sandbox/vmd.env` were never set by any code path — purely hand-staged once per host. On `superserve-vmd-use4` all three ended up blank after the us-east4 cutover: heartbeat silently never started, and the reconciler silently fell back to BoltDB-only mode (no DB audit trail, but it kept tripping the auto-fail rate limiter every pass since that check is in-memory and DB-independent).

This wires all three explicitly per region in `deploy-vmd.yml`, reusing secrets that already exist (`DATABASE_URL_PROD`, `DATABASE_URL_USWEST`, `PROD_INTERNAL_API_TOKEN`, `PROD_USW_INTERNAL_API_TOKEN`), and adds `DATABASE_URL` upsert support to `deploy-vmd.py` (CD already had upsert logic for the other two, just never called with a value). Two new repo variables hold the actual URLs: `CONTROL_PLANE_URL_PROD` (`https://api.superserve.ai`, shared by use4 + the legacy us-central1 host) and `CONTROL_PLANE_URL_USW` (pulled from the working usw2 host's existing config, not guessed).

- Reused existing GH secrets rather than introducing new ones — `DATABASE_URL_PROD`/`DATABASE_URL_USWEST` already back the Supabase migration step in `cd.yml`, and the `INTERNAL_API_TOKEN` secrets already existed under their current names.
- `DATABASE_URL` upsert is treated as sensitive (created with `chmod 0600`, matches the `INTERNAL_API_TOKEN` handling) and only upserts on a non-empty value, so a CD run without the secret set can never wipe a working value — same "empty = skip" convention as the existing vars.
- Left the underlying `superserve-api-usw2` Cloud Run default URL as-is for `CONTROL_PLANE_URL_USW` rather than standing up a custom domain for that cell — that's a separate, larger change and out of scope here.

## Testing

- Applied this exact fix by hand on `superserve-vmd-use4` (same env file, same values) and confirmed via logs (`"reconciler DB connection ready"`, `"heartbeat started"`) and the `host`/`reconciler_log` tables that heartbeat and DB-backed reconciliation both recovered, with the running vmd restart reattaching all VMs cleanly (0 stale).
- `deploy-vmd.py` syntax-checked (`python3 -m py_compile`); the upsert logic itself is unchanged in shape from the existing `CONTROL_PLANE_URL`/`INTERNAL_API_TOKEN` blocks it mirrors.
- Have not run this through an actual CD `deploy-vmd` workflow execution end-to-end — next merge to `main` touching a vmd-tracked path (or manual `workflow_dispatch`) will be the first real exercise of this code path.

### DIFF
--- a/.github/workflows/deploy-vmd.yml
+++ b/.github/workflows/deploy-vmd.yml
@@ -189,6 +189,12 @@ jobs:
           DATABASE_URL: ${{ secrets.DATABASE_URL_USWEST }}
           INTERNAL_API_TOKEN: ${{ secrets.PROD_USW_INTERNAL_API_TOKEN }}
         run: |
+          # Fail closed: an empty GCP_REGION makes deploy-vmd.py drop the region
+          # filter and fan out to every VMD_LABEL match — which would push
+          # usw's DATABASE_URL/CONTROL_PLANE_URL/INTERNAL_API_TOKEN onto the
+          # use4 primary host below instead of just skipping usw. This must
+          # abort before the fail-closed checks below, not run unscoped.
+          : "${GCP_REGION:?GCP_REGION_USW is unset — refusing an unscoped usw fan-out}"
           # Same fail-closed reasoning as the use4 step above — see its comment.
           : "${CONTROL_PLANE_URL:?CONTROL_PLANE_URL_USW is unset — refusing to deploy without it}"
           : "${DATABASE_URL:?DATABASE_URL_USWEST is unset — refusing to deploy without it}"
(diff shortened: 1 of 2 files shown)


pavitrabhalla:
@codex all good?

meAmitPatil (approved):

PR #271, opened 2026-07-24
vmd: skip the systemd linger query for provably-fresh unit launches

meAmitPatil:
Before starting a VM's unit, the launch path asks systemd whether a unit with the same name is still winding down (the answer widens the socket budget for the stop phase). For a brand-new VM the unit name has never existed, so the query always answers not-lingering — but it still costs a systemd round trip per launch, and concurrent launches serialize those round trips on the shared D-Bus connection.

This skips the query when the unit name is provably fresh: no known instance for the vmID (BoltDB already folded into the in-memory map at entry) and no pre-existing rundir. The rundir check covers the daemon-crash window — a prior attempt that wrote start.sh and possibly started the unit before dying leaves its rundir behind, so such a vmID never classifies as fresh even with no persisted state.

Paths that can genuinely be replacing a live or winding-down unit keep the query: resume, snapshot verify, retried creates (a failed attempt leaves its instance in the map), and in-loop retries (which already require the unit to be confirmed dead before proceeding).

No behavior change for any skipped case: a never-existed unit reads NotLoaded, which the query already mapped to not-lingering.

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -1103,7 +1103,7 @@ func (m *Manager) ResumeVM(ctx context.Context, vmID, snapshotPath, memPath stri
 	}
 
 	tFcStart := time.Now()
-	pid, err := m.startFirecrackerViaSystemd(ctx, vmID, socketPath, rootfsPath, inst.Config.BasePath, inst.Namespace)
+	pid, err := m.startFirecrackerViaSystemd(ctx, vmID, socketPath, rootfsPath, inst.Config.BasePath, inst.Namespace, false)
 	if err != nil {
 		return nil, fmt.Errorf("start firecracker for restore: %w", err)
 	}
@@ -1298,7 +1298,7 @@ func (m *Manager) VerifySnapshot(ctx context.Context, vmID string) (string, erro
 	// way out. Safe because the VM is paused (no live FC to disrupt), but it must
 	// not run concurrently with a resume of the same sandbox — fine for the
 	// debug/staging use this endpoint is gated to.
-	if _, err := m.startFirecrackerViaSystemd(ctx, vmID, socketPath, rootfsPath, inst.Config.BasePath, inst.Namespace); err != nil {
+	if _, err := m.startFirecrackerViaSystemd(ctx, vmID, socketPath, rootfsPath, inst.Config.BasePath, inst.Namespace, false); err != nil {
 		return "", fmt.Errorf("start firecracker for verify: %w", err)
 	}
 	defer func() { _ = stopUnitWithBudget(context.Background(), systemdUnitName(vmID)) }()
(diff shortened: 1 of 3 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@claude review

pavitrabhalla (commented):
**Review — `vmd: skip the systemd linger query for provably-fresh unit launches`**

I traced the change against the actual code. No correctness bugs found — the optimization is sound and, importantly, fails safe: the only dangerous direction (classifying a still-live unit as "fresh" and shrinking its replacement stop budget) is guarded on every path I could find.

**What holds up:**
- `unitLingering` is a pure `ActiveState` read, so skipping it has no side effect beyond the returned bool — a never-existed unit reads `NotLoaded` → not-lingering, identical to the query's own answer.
- Every `confirmUnitStopped` clear is on a systemd-confirmed-dead outcome (not-loaded, job `done`/`skipped`, or a successful synchronous stop). The `settleExpiredStopWait` "unit looks dead" branch deliberately retains the record. So a record is never cleared while the unit could still be alive — the failure mode is over-conservative (an extra linger query), never under.
- The `orphanScanDone` gate + `recordUnitStop` on startup orphans + `attempt == 1` guard + `!isBuildVM` all line up with the paths that can genuinely replace a live/winding-down unit.

**Non-blocking suggestions:**
1. **Fragile timing coupling (the one I'd most like addressed).** `processSettleWindow` (180s) is the sole thing covering a predecessor-era unit still `deactivating` at startup — `listActiveFirecrackerUnits` filters `active` only, so the orphan scan does not catch it. Its safety depends on 180s > the reconciler's first orphan sweep: `Interval` (30s) + `GracePeriod` (60s) + `stopJobWaitCap` (35s) = 125s, ~55s margin. That's a hardcoded coupling to three constants in `reconciler.go`/`systemd.go`, and if any of them grows the freshness guard silently regresses into exactly the bug this PR prevents. Consider deriving `processSettleWindow` from those constants, or adding a test asserting `processSettleWindow > Interval + GracePeriod + stopJobWaitCap`.
2. **Document the `deactivating` gap at its source.** The reasoning that the active-only listing is backstopped by the clock window currently lives split across `manager.go` and `systemd.go`. A one-liner near `listActiveFirecrackerUnits` (or the `unitMaybeWindingDown` clock branch) noting the two mechanisms are complementary would save the next reader the cross-file trace.
3. **Minor:** `recentUnitStops` has no eviction for stops that never confirm (permanently-failed stop of a vmID never seen again). It's the safe direction and entries are tiny, but it's an unbounded-in-principle map over a very long-lived process.

Tests cover `unitMaybeWindingDown` and `stopJobResult` well; the `freshUnit`/`orphanScanDone`/`attempt` composition in `manager.go` is only exercised indirectly, which is understandable given it's integration-level.

Overall: looks good to merge. The suggestions are hardening, not blockers.

---
_Generated by [Claude Code](https://claude.ai/code)_

pavitrabhalla (approved):

PR #272, opened 2026-07-24
api: retry a deadline-exceeded vmd boot once on create and resume

meAmitPatil:
A create or resume whose vmd boot call dies on a transient — DeadlineExceeded, or Unavailable that outlived the vmd client's dial-site retry window — previously failed the request outright even though the daemon may be seconds from healthy. The daemon already serializes same-ID boots and recognizes a completed prior attempt (a retried boot either adopts the VM the first attempt brought up or boots cleanly), so the control plane can safely retry.

This wraps the boot calls (create RestoreSnapshot, ResumeInstance and its stateless fallback, and the reaper's pause-rollback resume) in a helper that retries exactly once on those transients, under a fresh budget. The resume sequence runs under one shared clock so the worst-case hold stays bounded. A request whose own context is already dead never retries — a boot landing after the client gave up would activate a sandbox the caller believes failed. Where a late-landing boot has no self-heal path (create's failed row, the reaper's terminal fail) a best-effort destroy follows; the resume error branches deliberately skip it, since a follow-up resume adopts the live VM and the reconciler reaps the paused-row/active-VM state otherwise.

Every other error code is a real answer and still fails fast. The API load balancer's backend timeout is raised to cover the full boot-plus-postwork envelope so a slow-but-successful path can't be cut off mid-flight.

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -192,30 +192,31 @@ func (h *Handlers) vmdForHost(ctx context.Context, hostID string) (VMDClient, er
 // vmdTimeout is the default deadline for VMD gRPC calls.
 const vmdTimeout = 30 * time.Second
 
-// retryBootOnDeadline runs one vmd boot call under vmdTimeout and, when the
-// attempt dies on DeadlineExceeded with the caller's context still alive,
-// runs exactly one more under a fresh vmdTimeout — worst case 2×vmdTimeout
-// for a caller that waits. This is the layer above the dial-site
-// Unavailable retry (retryUnavailableUnaryInterceptor absorbs a restarting
-// daemon; this absorbs the boot that then runs into its deadline). The
-// daemon serializes same-ID boots and recognizes a completed prior attempt,
-// so the retry either adopts the VM the first attempt brought up or boots
-// cleanly — never a double boot. A dead caller skips the retry: a boot
-// landing after the client gave up would activate a sandbox the caller
-// believes failed. (pauseWithRetry deliberately differs on both points —
-// any-error retry, detached context — because a late pause reconciles
-// state instead of surprising the caller.)
-func retryBootOnDeadline(parent context.Context, sandboxID string, boot func(context.Context) (string, uint32, uint32, error)) (ip string, vcpu, memMiB uint32, retried bool, err error) {
+// retryTransientBoot runs one vmd boot call under vmdTimeout and, when the
+// attempt dies on a transient — DeadlineExceeded, or Unavailable that
+// outlived the dial-site interceptor's window — with the caller's context
+// still alive, runs exactly one more under a fresh vmdTimeout; worst case
+// 2×vmdTimeout for a caller that waits (the API LB's backend timeout is
+// sized above that envelope). The daemon serializes same-ID boots and
+// recognizes a completed prior attempt, so the retry either adopts the VM
+// the first attempt brought up or boots cleanly — never a double boot; at
+// most one retry per boot and the daemon's restore semaphore bound the
+// added load. A dead caller skips the retry: a boot landing after the
+// client gave up would activate a sandbox the caller believes failed.
+// (pauseWithRetry deliberately differs on both points — any-error retry,
+// detached context — because a late pause reconciles state instead of
+// surprising the caller.)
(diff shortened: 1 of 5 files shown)


internal/api/reaper.go
```diff
@@ -347,11 +347,14 @@ func (h *Handlers) rollbackPausedVM(ctx context.Context, sbx db.ClaimExpiredSand
...
+	// failure, before the customer ever sees the transition. Same deadline
+	// retry as the user-facing boots — the background ctx never reads dead,
+	// which is right: a rollback landing late still beats marking the
+	// sandbox failed.
+	_, _, _, _, err := retryTransientBoot(ctx, sbx.ID.String(), func(rctx context.Context) (string, uint32, uint32, error) {
+		return vmd.ResumeInstance(rctx, sbx.ID.String(), snapshotPath, memPath)
+	})
 	if err != nil {
 		rl.Error().Err(err).Msg("reaper: rollback resume failed")
 		h.markSandboxFailed(ctx, sbx, "rollback resume failed after pause DB error", rl)
```
pavitrabhalla, line 368: This is the same race the CreateSandbox destroy call (handlers.go:1901) was added for: retryTransientBoot exhausting both attempts on DeadlineExceeded doesn't mean the boot didn't land late on the host. But here the sandbox goes to 'failed' with no automatic retry — unlike resume, which reverts to 'paused' and can re-adopt a late boot via a follow-up resume on the same ID. So there's no self-heal path here, just the reconciler eventually catching it. Should this get the same best-effort DestroyInstance before markSandboxFailed?

pavitrabhalla (commented):
PR description says "Only DeadlineExceeded is retried; Unavailable is already retried inside the vmd client" but the shipped code (isVMDUnavailable) retries Unavailable too, per the 4th commit. Might want to update the description before merging so it matches the actual behavior.

internal/api/handlers.go
```diff
@@ -568,20 +597,31 @@ func (h *Handlers) resumePausedSandbox(c *gin.Context, sandbox *db.Sandbox, team
...
 			// already has baked into the snapshot — not inject new ones.
-			ipAddress, actualVcpu, actualMemMiB, err = vmd.RestoreSnapshot(vmdCtx, sandboxID.String(), snapshotPath, memPath, resumeBasePath, "", sandbox.TeamID.String(), ownerIDFromContext(c), nil)
+			// Single attempt under whatever remains of the sequence budget:
+			// the fallback is already the second recovery layer.
+			fctx, fcancel := context.WithTimeout(bootCtx, vmdTimeout)
+			ipAddress, actualVcpu, actualMemMiB, err = vmd.RestoreSnapshot(fctx, sandboxID.String(), snapshotPath, memPath, resumeBasePath, "", sandbox.TeamID.String(), ownerIDFromContext(c), nil)
+			fcancel()
 			if err != nil {
 				log.Error().Err(err).Str("sandbox_id", sandboxID.String()).Msg("VMD RestoreSnapshot fallback failed")
 				revertToPaused()
```
pavitrabhalla, line 631: Same gap as the reaper rollback: if retryTransientBoot or this fallback RestoreSnapshot times out after the boot actually landed late on the host, there's no destroy call here, just revertToPaused(). Same applies to the ResumeInstance failure branch below (line 633). Softer than the reaper case since a follow-up resume on the same ID lets vmd adopt the late boot, but there's a window until that retry where the row says 'paused' while a VM is actually up. Worth a best-effort destroy here too, or is self-heal-on-retry considered sufficient?

internal/api/reaper.go
meAmitPatil: Right — failed is terminal here so there's no adopt-on-retry like resume has. Added the same best-effort destroy before markSandboxFailed.

internal/api/handlers.go
meAmitPatil: Went with self-heal as sufficient here, now documented in the code: a follow-up resume on the same ID adopts the late boot via the daemon's retried-launch check (which makes the retry an instant resume — a destroy would throw that away), and if no retry ever comes the reconciler's paused-in-DB-but-running drift rule reaps it on the next pass. So the window is bounded either way, and the adopt path is worth keeping.

meAmitPatil:
Description updated to match the shipped behavior (both transients retried + the destroy-vs-self-heal split spelled out).

pavitrabhalla (approved):

PR #273, opened 2026-07-24
vmd: add supervision-mode record field and direct-spawn cgroup primitives

meAmitPatil:
First of a series moving VM supervision from per-VM systemd units to direct-spawned processes in per-VM cgroups under the daemon's delegated subtree. systemd's PID 1 starts units strictly one at a time, so concurrent launches wait in a single-file queue; direct spawn removes that serialization at the source. Nothing is wired yet — this lands the self-contained foundation:

- A supervision field on the VM record, round-tripped through the store. The empty value is canonical for unit-supervised VMs, so records written by this binary remain correct under an older binary and legacy records need no migration.
- cgroup v2 primitives: delegated-subtree setup (daemon moved to a leaf group first, per the no-internal-process rule), per-VM group creation returning a directory fd for CLONE_INTO_CGROUP (the child is born inside its group — no move window), an atomic kill that waits for the group to empty before returning (SIGKILL is asynchronous for a userfaultfd-blocked process, and the network slot must not be recycled while the tap is still held), bounded removal, and a fail-closed scan.
- Direct-spawn helpers: a minimal environment allowlist (the daemon's own environment must never reach the sandbox boundary's process memory — test-enforced), a launch-script wrapper pinning NOFILE to the service defaults (the fd table preallocates to the soft limit), per-VM console files, a session-isolated spawn, and a reaper that records exit codes.

Follow-ups wire the launch/stop/liveness/reattach/reconciler paths behind a flag with a dual-mode fleet during migration.


internal/vm/cgroup_linux.go
```diff
@@ -0,0 +1,308 @@
...
+		if pid, err := strconv.Atoi(s); err == nil {
+			pids = append(pids, pid)
+		}
+	}
+	return pids, nil
+}
+
+// vmCgroupDir returns the per-VM cgroup path for vmID.
+func (t *cgroupTree) vmCgroupDir(vmID string) string {
+	return filepath.Join(t.vms, vmID)
```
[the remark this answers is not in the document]
meAmitPatil: primitives are dead code on this branch — nothing calls killVMCgroup/createVMCgroup until the launch wiring in #274, and thats where the guard lives (safeVMCgroupDir rejects non-leaf ids + isLeafName at the restore entry). not reachable here, covered in the stacked child.

internal/vm/directspawn_linux.go
```diff
@@ -0,0 +1,109 @@
...
+
+// openVMConsole opens the per-VM console file the child's stdout/stderr
+// attach to. Firecracker's own log line plus panic/abort output land here —
+// including the UFFD handler-death message that precedes an exit(70).
+// A file, not a journald stream: journald rate-limits by unit cgroup, so
+// shared streams would let one spamming VM suppress vmd's own logs. Growth
+// is bounded by the reconciler's disk sweep, not here.
+func openVMConsole(runDir, vmID string) (*os.File, error) {
+	return os.OpenFile(filepath.Join(runDir, vmID, "console.log"),
+		os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0o644)
```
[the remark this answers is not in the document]
meAmitPatil: fixed in #274 — openVMConsole opens O_TRUNC per launch there, bounded by the paired firecracker serial cap (1 MiB/boot). uncalled on this branch until the launch wiring, so it's a no-op here; the fix belongs in the child.

internal/vm/cgroup_linux.go
```diff
@@ -0,0 +1,308 @@
...
+// vms/<vmID>/ group. PID 1 is never involved per VM — that per-unit
+// serialization is the cost this path exists to delete.
+
+const (
+	cgroupMount = "/sys/fs/cgroup"
+	// vmMemoryMaxFraction of MemTotal caps the vms/ subtree, replacing
+	// sandboxes.slice's MemoryMax=95%. vmd itself lives in daemon/, outside
+	// the cap — the daemon must never compete with sandboxes for the
+	// ceiling that protects it.
+	vmMemoryMaxFraction = 0.95
```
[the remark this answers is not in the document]
meAmitPatil: fixed in #274 — VMs moved under a delegated superserve-vms.service in sandboxes.slice, so both populations share the one 95% cap, and arming refuses without the host reserve. this per-subtree cap is dead code on #273; the real fix lives where it's called.

internal/vm/directspawn_linux.go
```diff
@@ -0,0 +1,109 @@
...
+// Pdeathsig: children must outlive vmd; that is the survival property.
+func spawnDirect(scriptPath string, cgroupDir *os.File, console *os.File) (*exec.Cmd, error) {
+	cmd := exec.Command(scriptPath)
+	cmd.Env = directSpawnEnv()
+	cmd.Stdout = console
+	cmd.Stderr = console
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Setsid:      true,
+		UseCgroupFD: true,
+		CgroupFD:    int(cgroupDir.Fd()),
```
pavitrabhalla, line 95: cgroupDir's last use is `.Fd()` here — after that nothing keeps the `*os.File` reachable. `os.File` has a finalizer that closes the underlying fd on GC; if that runs before or during the clone3 call inside `cmd.Start()`, the fd gets closed out from under the copied int in `SysProcAttr.CgroupFD`. Should this have a `runtime.KeepAlive(cgroupDir)` after `Start()` returns?

internal/vm/cgroup_linux.go
```diff
@@ -0,0 +1,308 @@
...
+// thread of it.
+func (t *cgroupTree) createVMCgroup(vmID string) (*os.File, error) {
+	dir := t.vmCgroupDir(vmID)
+	if err := os.Mkdir(dir, 0o755); err != nil && !os.IsExist(err) {
+		return nil, fmt.Errorf("mkdir vm cgroup: %w", err)
+	}
+	if err := os.WriteFile(filepath.Join(dir, "memory.oom.group"), []byte("1"), 0o644); err != nil {
+		return nil, fmt.Errorf("set oom.group: %w", err)
+	}
+	f, err := os.Open(dir)
```
pavitrabhalla, line 180: This `*os.File` (the per-VM cgroup dir fd) is returned but never closed anywhere in this diff. It's only needed synchronously for the `CLONE_INTO_CGROUP` call in `spawnDirect` — once wired up, will the caller close it after `cmd.Start()` returns, or does that still need to be added? Left as-is this leaks one fd per VM launch.

internal/vm/directspawn_linux.go
```diff
@@ -0,0 +1,109 @@
...
+	return prefix + "prlimit --nofile=1024:524288 " + rest, nil
+}
+
+// openVMConsole opens the per-VM console file the child's stdout/stderr
+// attach to. Firecracker's own log line plus panic/abort output land here —
+// including the UFFD handler-death message that precedes an exit(70).
+// A file, not a journald stream: journald rate-limits by unit cgroup, so
+// shared streams would let one spamming VM suppress vmd's own logs. Growth
+// is bounded by the reconciler's disk sweep, not here.
+func openVMConsole(runDir, vmID string) (*os.File, error) {
```
pavitrabhalla, line 56: This doesn't `MkdirAll(runDir/vmID)` before opening `console.log`, unlike `startFirecrackerColdBoot`/`startFirecrackerViaSystemd` which defensively `MkdirAll` the target dir even though an earlier step is expected to have created it. Will the follow-up wiring guarantee this directory always exists before `openVMConsole` runs, or should this ensure it itself?

internal/vm/cgroup_linux.go
```diff
@@ -0,0 +1,308 @@
...
+	return f, nil
+}
+
+// vmCgroupPopulated reports whether the VM's cgroup holds any live process.
+// (false, nil) on a missing group — a removed group is definitively empty.
+// Read errors OTHER than not-exist return the error: callers on kill/reap
+// paths must treat an unreadable answer as "maybe alive", mirroring the
+// unit oracle's inconclusive-reads-alive rule, or an overloaded host reaps
+// live VMs.
+func (t *cgroupTree) vmCgroupPopulated(vmID string) (bool, error) {
```
pavitrabhalla, line 193: Comment says this mirrors 'the unit oracle's inconclusive-reads-alive rule' in systemd.go, but there's no shared helper/constant between the two — they're independently written. Worth a shared sentinel so the two supervision modes can't drift on what counts as inconclusive, or is duplicating the rule here intentional for this stage?

internal/vm/state.go
```diff
@@ -121,13 +121,33 @@ type VMRecord struct {
 	// Persisted so usage attribution survives a vmd restart.
 	TeamID  string `json:"team_id,omitempty"`
 	OwnerID string `json:"owner_id,omitempty"`
+	// Supervision dispatches liveness/stop/reattach for this VM's current
+	// run. Empty (SupervisionUnit) is canonical for systemd-unit VMs so
+	// records written by this binary stay readable-and-correct under a
+	// rollback binary that predates the field; never write a non-empty
+	// value for unit mode.
+	Supervision string `json:"supervision,omitempty"`
```
pavitrabhalla, line 177: `Status` on this same struct is a typed `VMStatus` with a `String()` method; `Supervision` is a bare string with untyped constants. Given the follow-up PRs will be switching on this value across liveness/stop/reattach/reconciler, was a named type (`type Supervision string`) considered here?

internal/vm/cgroup_linux.go
```diff
@@ -0,0 +1,308 @@
...
+			return nil, fmt.Errorf("mkdir %s: %w", d, err)
+		}
+	}
+	// Move every root-resident process (vmd + any helpers) into daemon/.
+	// Writing a PID to cgroup.procs moves all its threads.
+	pids, err := readCgroupProcs(t.root)
+	if err != nil {
+		return nil, fmt.Errorf("read root procs: %w", err)
+	}
+	for _, pid := range pids {
```
pavitrabhalla, line 84: If a pid in `readCgroupProcs`'s snapshot exits before its individual write to `daemon/cgroup.procs`, the kernel returns ESRCH, which isn't `os.IsNotExist`, so this aborts `setupCgroupTree` entirely. The comment above anticipates 'vmd + any helpers' — should an exited-pid write be treated as non-fatal here?

internal/vm/cgroup_linux.go
```diff
@@ -0,0 +1,308 @@
...
+	return false
+}
+
+// killVMCgroup SIGKILLs every process in the VM's group (one atomic write)
+// and waits until the group reads empty. SIGKILL is asynchronous for a
+// UFFD-blocked Firecracker — it holds its netns and tap until it actually
+// exits — so returning before populated==0 would resurrect the tap-busy
+// failure class the network pool's verify discipline exists to prevent.
+// The wait is event-driven (inotify on cgroup.events) with a poll floor.
+func (t *cgroupTree) killVMCgroup(ctx context.Context, vmID string, budget time.Duration) error {
```
pavitrabhalla, line 229: This wait-with-deadline-and-fsnotify-fallback shape is similar to `waitForSocketFile`/`waitForSocketFilePolling` in manager.go, and the kill-then-wait shape echoes `sigkillPID`/`waitForPIDExit`. Was factoring a shared helper considered, or do the differences (dir-watch-for-create vs content-repoll, ctx support) make that not worth it?

internal/vm/cgroup_linux.go
```diff
@@ -0,0 +1,308 @@
...
+				return ctx.Err()
+			}
+		}
+	}
+}
+
+// removeVMCgroup rmdirs the VM's group. The kernel can lag rmdir briefly
+// after the last exit; bounded retries absorb that, and the reconciler
+// sweeps any straggler.
+func (t *cgroupTree) removeVMCgroup(vmID string) error {
```
pavitrabhalla, line 288: This retry loop has no `context.Context`, so it can't be interrupted during shutdown (up to 500ms worst case). `waitForSocketFilePolling`/`waitForSocketConnectable` elsewhere in the package have the same gap, so this may just be consistent with existing style — flagging since `killVMCgroup` right above does honor ctx.

awille-ai (changes requested):
The overall foundation is thoughtfully structured, and the previously raised controller issue is fixed. CI and Terraform checks are green. I also agree that several rollout-only hazards can reasonably live in the stacked wiring PR if these helpers remain unreachable here.

One correctness issue is intrinsic to this helper and blocks approval: `spawnDirect` copies `int(cgroupDir.Fd())` into `SysProcAttr.CgroupFD`, but it neither keeps the `*os.File` alive through `cmd.Start()` nor closes it afterward. Because `os.File` has a finalizer, the object can become unreachable immediately after the `.Fd()` call and be closed by GC before or during the `clone3(CLONE_INTO_CGROUP)` path, producing an intermittent invalid/incorrect FD at launch. Conversely, if callers retain it defensively, the current API makes it easy to leak one directory FD per VM launch.

Please make `spawnDirect` own the synchronous FD lifecycle explicitly: keep the file alive until `cmd.Start()` returns (for example with `runtime.KeepAlive(cgroupDir)`), then close it on both success and failure, or redesign the API so ownership is unambiguous and enforced in one place. Add a regression test around repeated launches/GC or otherwise test the ownership contract.

The unresolved `openVMConsole` directory-creation question is secondary, but this FD lifetime bug can make valid launches fail nondeterministically and should be fixed in this foundational PR rather than deferred to the child wiring.

internal/vm/directspawn_linux.go
meAmitPatil: caller defers cgDir.Close() so the pending defer keeps it reachable through cmd.Start()'s clone3 — no finalizer window today. but you're right it's implicit; i'll add an explicit runtime.KeepAlive(cgDir) in #274, where spawnDirect is actually called.

internal/vm/cgroup_linux.go
meAmitPatil: already handled in #274 — the launch path defers cgDir.Close() right after createVMCgroup. dead code on #273 so nothing leaks here; the close lives with the caller in the wiring PR.

internal/vm/directspawn_linux.go
meAmitPatil: the dir is RunDir/vmID, which the caller already MkdirAll's (via filepath.Dir(socketPath)) before this. i'll add a defensive MkdirAll in #274 so it's not order-dependent.

internal/vm/cgroup_linux.go
meAmitPatil: duplicated on purpose for now — the two oracles read different files (cgroup.events vs ActiveState). i'll fold them into a shared "inconclusive = alive" sentinel in #274, where both call paths land.

internal/vm/state.go
meAmitPatil: fair. the switches that make a type worthwhile are all in #274, so i'll promote it to `type Supervision string` (+ String()) there and keep #273 as just the field add.

internal/vm/cgroup_linux.go
meAmitPatil: already resolved in #274 — setupCgroupTree is replaced there by adoptVmsScope, which only moves the keeper (sleep infinity, or nothing on restart) and fails-closed to the unit path on error. this broader loop is gone in the wiring PR.

internal/vm/cgroup_linux.go
meAmitPatil: left separate for now — ctx support + content-repoll (vs dir-watch-for-create) make a shared helper awkward. i'll extract in #274 if a third caller shows up.

internal/vm/cgroup_linux.go
meAmitPatil: consistent with waitForSocketFilePolling as you noted; ~500ms worst case and the reconciler sweeps stragglers. i'll thread ctx through in #274 to match killVMCgroup.

meAmitPatil:
good call — moved the whole fd lifecycle into spawnDirect (b03cd30). it now runtime.KeepAlive's cgroupDir across cmd.Start()'s clone3, then closes it on both success and failure via a defer, so ownership is unambiguous: callers hand it off and never touch it — no finalizer race, no per-caller leak. added TestSpawnDirectOwnsCgroupFD that drives the failure path and asserts the fd is closed on return. left the openVMConsole MkdirAll for #274 as discussed since the caller already MkdirAll's that path first.

internal/vm/directspawn_linux.go
meAmitPatil: update: ended up doing this in #273 itself (b03cd30) rather than #274 — spawnDirect now KeepAlive's cgroupDir across clone3 and closes it, per Alejandro's review. fixed here.

internal/vm/cgroup_linux.go
meAmitPatil: update: moved into #273 (b03cd30) — spawnDirect owns the close on every path now, so it no longer leans on the caller's defer. supersedes my #274 note above.

awille-ai (approved):
Re-reviewed at the latest head. The blocker from my previous review is fixed correctly in this PR: `spawnDirect` now explicitly owns the cgroup directory FD, keeps the `*os.File` alive through `cmd.Start()`/`clone3`, and closes it on both success and failure. That removes both the GC/finalizer race and the per-launch FD leak risk.

I also rechecked the broader foundation: cgroup controller propagation, environment isolation, bounded kill semantics, supervision-record compatibility, and the stacked-PR boundaries remain coherent. CI and Terraform checks are green. The remaining open discussion items are either non-blocking design questions or explicitly addressed in the child wiring PR where they become reachable.

Approved.

awille-ai:
One fail-closed edge case still looks intrinsic to this foundation: `parseCgroupPopulated` returns `false` both for a valid `populated 0` and for a missing/malformed `populated` field. Consequently `vmCgroupPopulated` returns `(false, nil)`, and callers can treat an inconclusive kernel-file response as definitively empty. In #274 this becomes reachable from liveness, kill completion, reconciliation, and drain checks, despite the stated “inconclusive = alive” contract.

Could the parser return `(bool, error)` (accepting only exactly `populated 0` or `populated 1`) and propagate malformed/missing data as inconclusive? The existing tests currently codify the unsafe case by expecting `"frozen 0\n"` and `""` to be false rather than errors.

meAmitPatil:
@codex review

pavitrabhalla (approved):
LGTM

PR #278, opened 2026-07-27
ci: use dedicated smoke-test API key secrets

meAmitPatil:
The smoke tests authenticate with team API keys, but the workflows sourced them from the `*_INTERNAL_API_TOKEN` secrets — which deploy-vmd also consumes as the host heartbeat bearer token. One secret serving two incompatible purposes means updating it for either consumer breaks the other.

Point the smoke-test `API_KEY` env at the new dedicated secrets:
- `SS_TEST_API_KEY_USW` for production/us-west2 (terraform-cd and terraform-rollout-production)
- `SS_TEST_API_KEY_USEAST` for the use cell (terraform-cd us-east4, terraform-rollout-production usc)

Also fixes the rollout-production usw job, which pointed at the use cell's secret.

The `*_INTERNAL_API_TOKEN` secrets are now used only by deploy-vmd, matching their name. Staging still uses `STAGING_INTERNAL_API_TOKEN` for its smoke test; it has no dual consumer, so it can move to a dedicated secret separately.


awille-ai (approved):
Reviewed the latest head. The temporary verification workflow flagged by Codex has been removed; the PR now changes only the two production rollout/CD workflows.

The secret mapping is now consistent with the target cells:

- production us-west2 uses `SS_TEST_API_KEY_USW` in both terraform-cd and rollout-production;
- the use cell uses `SS_TEST_API_KEY_USEAST` for both the us-east4 CD smoke test and the usc rollout smoke test;
- the internal API-token secrets are no longer shared with production smoke tests, avoiding coupling between team API-key auth and VMD heartbeat auth.

This also corrects the previous us-west rollout mismatch, which referenced the use-cell secret. Staging remains unchanged as described.

CI, Terraform Checks, and Terraform Plans are green. No Terraform resources are modified.

Approved.

PR #281, opened 2026-07-28
Roll up preview URL authentication stack

mohamedsobhi777:
## What changed

Rolls the complete preview URL authentication stack into `main`, including:

- per-port public/private access policy
- header-token authentication and token rotation
- browser-compatible signed preview links
- VMD/proxy capability advertisement and enforcement
- database migrations and generated query updates

This branch was created from current `main` and merges the final stacked branch, `agent/preview-port-header-auth`, so all previously reviewed phases land together without regressing newer mainline changes.

## Why

PRs #240, #242, and #243 were merged into successive stacked feature branches rather than `main`. Only the publication phase from #239 reached production. This roll-up makes the complete stack deployable.

## Rollout

Before merging, deploy Proxy and then VMD from this branch through staging and production. Wait for host heartbeats to advertise the complete preview capability chain. After that, merge this PR; the merge-to-main workflows will apply migrations and deploy API, Proxy, and VMD.

## Validation

- Clean merge with current `main`
- Relevant API, auth, preview, proxy, and scheduler tests pass locally
- Full Linux CI is required for VMD/network packages that do not compile on macOS


mohamedsobhi777:
@awille-ai This is the clean roll-up of the already-reviewed preview-auth stack onto current main. Could you review/approve it for the ordered Proxy → VMD → API production rollout? Full CI is green.

awille-ai (approved):
Re-reviewed the latest roll-up head at the same security/reliability bar.

The previously reported CORS-preflight authentication bypass is fixed. Private preview preflights are now:

- recognized separately from ordinary `OPTIONS` requests;
- gated by the proxy's configured origin allowlist;
- answered locally with a 204 response;
- never forwarded unauthenticated to sandbox code.

The new tests cover allowed and disallowed origins and explicitly assert that preflights and unauthenticated `OPTIONS` requests are not forwarded.

I also rechecked the major security and rollout surfaces:

- preview tokens use a domain-separated HMAC, canonical authenticated claims, bounded hostile-input parsing, exact sandbox/port/generation scope, and expiry checks;
- token carriers are stripped before forwarding to user-controlled code on public, legacy, and private ports;
- browser bootstrap uses a secure host-only `__Host-` cookie, `HttpOnly`, `SameSite=None`, `Partitioned`, `Cache-Control: no-store`, and `Referrer-Policy: no-referrer`, then redirects to an absolute same-origin URL with the credential removed;
- unknown policy values, missing generations, invalid seeds, and incomplete capabilities fail closed;
- per-port token generations are durable tombstones, advance on re-publication and real access changes, and do not wrap;
- migrations preserve existing public behavior while maintaining a restrictive rollback-compatible fallback for mixed/private policies;
- API mutations serialize on the sandbox row, validate host capabilities inside the same transaction, and distinguish policy delivery from credential issuance;
- VMD/proxy capability negotiation and revisioned snapshots support the ordered rollout described in the PR.

The main CI, integration, build, vet, sqlc, migration-order, and Terraform workflows are green. The `loop-pr-loop` check failed only because the external review model hit a 429 session limit after starting its review; that is not a repository-code failure.

Approved, with an important operational condition: follow the documented rollout order and verify host heartbeats advertise the complete capability chain before the API starts allowing private publication or issuing credentials. In practice: deploy Proxy, then VMD, wait for capability convergence, and only then merge/deploy the control plane and migrations.

PR #282, opened 2026-07-28
deploy: enforce host patching policy on every vmd deploy

meAmitPatil:
VM hosts must not install OS updates automatically, and library-upgrade tooling must never restart the firecracker or platform units — a restarted firecracker unit is a destroyed customer VM, and firecracker-netns units hold live VM network namespaces.

Enforced at both host lifecycle layers:

**At instance creation** (`sandbox-host` module): the policy script is prepended to the environment's startup script, or installed alone where none is defined. The module's existing `ignore_changes = [metadata]` means this applies only to newly created instances — running hosts are not perturbed by an apply.

**On every vmd deploy** (deploy assets): the same two files are installed alongside the systemd units the deploy already manages, and `apt-daily-upgrade.timer` is disabled — so a drifted or hand-edited host converges back to the policy on its next deploy.

- `deploy/needrestart-superserve.conf` → `/etc/needrestart/conf.d/50-superserve.conf`: needrestart never auto-restarts `firecracker*`, `superserve-*`, `proxy`, or `unbound` units.
- `deploy/apt-no-auto-upgrades.conf` → `/etc/apt/apt.conf.d/99superserve-no-auto-upgrades`: unattended package installation off; package-list refresh stays on so maintenance windows see accurate pending updates.

Patching happens in deliberate maintenance windows (pause fleet → patch → reboot if required → resume).


awille-ai (approved):
Re-reviewed at the latest head after the rebase. I am approving this revision.

I previously treated failure to disable `apt-daily-upgrade.timer` as a blocker. On closer review, that was too strict: the authoritative control is the installed apt configuration setting `APT::Periodic::Unattended-Upgrade "0"`; `apt.systemd.daily` consults that setting, so leaving the timer enabled does not by itself permit unattended package installation. Disabling the timer is additional defense in depth, and the best-effort `|| true` is reasonable for image/version portability.

The rest of the implementation is coherent: the policy is applied on first boot and reasserted on each VMD deploy, package-list refresh remains enabled for maintenance windows, needrestart exclusions protect Firecracker/network/platform units, the deploy workflow includes and self-triggers on the policy assets, and existing instances are not unexpectedly rewritten by Terraform metadata changes.

CI, Terraform checks, and Terraform plans are green. Approved.

pavitrabhalla (approved):

PR #283, opened 2026-07-28
vmd: reconciler detects empty firecracker shells

meAmitPatil:
Unit-level liveness cannot distinguish a running VM from an empty Firecracker shell: a unit restarted outside vmd (host tooling, manual systemctl) destroys the guest, and the replacement process only ever becomes a VM again through vmd's own restore path — yet it looks fully active to systemd, so the reconciler's dead-VM rule never fires and errors against it read as transient. Sandboxes can sit dead behind an active-looking process indefinitely.

Probe the layer that cannot be fooled: ask Firecracker itself. A new reconciler drift rule (`fc_empty_shell`) queries each active unit's FC API; a healthy API explicitly answering "Not started" for a DB-active sandbox is an empty shell. The rule stops the shell, marks the sandbox failed, and cleans vmd state — with the full existing safety stack: only an affirmative "Not started" counts (probe errors are never evidence — they're the socket/unit rules' failure class), the grace period filters the brief legitimate "Not started" window during launch, the per-VM op lock plus a re-probe under the lock protect in-flight restores, and the auto-fail budget caps blast radius. Stop failure leaves the row untouched for the next pass rather than stranding a live unit behind a failed row.

`VMState` is a small helper over the vendored DescribeInstance operation; tests cover the probe against a fake FC on a unix socket, including the probe-error-is-not-evidence property.


awille-ai (changes requested):
The underlying detection strategy is sound, and the prior review findings have been addressed well:

- probes now share a bounded stage budget and run with capped concurrency;
- the Unix dial honors context cancellation;
- keep-alives are disabled so successful probes do not leak idle socket FDs;
- the empty-shell verdict requires an affirmative Firecracker `Not started` response;
- probe errors remain non-evidence;
- the VM op lock plus re-probe protects concurrent restore/launch work;
- post-stop persistence uses a fresh bounded context;
- CI and Terraform checks are green, and all existing review threads are resolved.

One sequencing issue still blocks approval: the auto-fail budget is consumed before `stopUnit` succeeds. If stopping the empty shell fails, the DB row is intentionally left active for the next pass, but that attempt has already spent one of the limited failure-budget slots. Repeated transient systemd failures can therefore exhaust the budget without ever performing the protected state transition, after which the reconciler will suppress future attempts and leave the sandbox indefinitely active behind an empty shell—the exact condition this PR is intended to repair.

Please consume the auto-fail budget only after the shell has been successfully stopped and immediately before the irreversible DB/local-state transition, or add an explicit reservation/rollback mechanism so a failed stop does not permanently spend budget. The audit semantics should also distinguish a stop failure from an actual auto-fail action.

Once failed stop attempts cannot exhaust the remediation budget, I do not see another blocker.

awille-ai (approved):
Re-reviewed at the latest head. The budget-accounting blocker is fixed correctly.

The reconciler now distinguishes three outcomes after reserving a slot:
- stop not enqueued: refund the slot and retry later;
- stop enqueued but completion unconfirmed: retain the slot because the destructive action may still complete;
- stop confirmed: retain the slot and persist the failed state with a fresh bounded context.

That preserves the blast-radius cap without permanently exhausting it on provable pre-action failures. The earlier probe-latency, context-aware Unix dial, idle-connection, and detached-persistence issues are also addressed. The per-VM lock and second probe still protect restores, probe errors remain non-evidence, and CI plus Terraform checks are green.

Approved.

meAmitPatil:
@codex review

PR #284, opened 2026-07-29
DB: add saved snapshot expand and host-admission guard

mohamedsobhi777:
## Stack

**1 of 4** for saved snapshots and independent forks. This is the additive expand and rolling-deploy safety layer; later PRs add VMD/storage, the control plane/API, then indexes/contracts/canary rollout. Part of #221.

## What changes

- Adds saved-snapshot identity, lifecycle, lineage, quota, and storage-ledger schema.
- Keeps `sandbox_snapshots_v1` globally disabled and clears any enabled team override.
- Preserves the legacy non-partial `snapshot_sandbox_unique` index so old pause writers remain compatible.
- Preserves secret scrub history for rolling old writers with a database trigger plus a concurrency-safe, merge-only backfill.
- Enables RLS on internal storage-ledger tables with no client-visible policies.
- Splits hot-table changes into bounded migrations, adds constraints `NOT VALID`, and validates them separately.
- Defers nonessential hot-table indexes until the pre-contract maintenance gate while the feature is dark.
- Adds a database-owned active-host admission guard for new launches, resumes, host moves, and template builds. It serializes with host drain/status updates and fails closed with `SS006`.
- Makes ordinary/template sandbox creation commit the trigger-backed DB admission before any VMD restore, preventing a rejected insert from leaving an untracked VM during mixed-version rollout.

## Rollback posture

Rollback to this PR (or a later stack revision) is safe: VMD launch cannot run before the database admission succeeds. Database rollback is intentionally **leave-forward**—the expand changes are additive, and removing columns/types after a mixed-version deploy would be riskier than leaving the dark schema in place. The legacy runtime identity index remains in place.

Before routing traffic to any **pre-PR1** API revision, first return every `draining` host to `active` and verify convergence. Those older binaries launch VMD and perform the INSERT in parallel; if a host cannot be reactivated, do not roll back that far. This is recorded in `deploy/README.md`.

## Validation

- `go run github.com/sqlc-dev/sqlc/cmd/sqlc@v1.30.0 generate`
- Full current integration and migrate-team suites with `-race`
- Full integration suite from a binary compiled at the exact `main` base, run against the expanded schema
- Real-Postgres host-admission coverage for active/draining/unhealthy/missing hosts, resume/template/host-move paths, lock serialization, and non-owner trigger execution
- Regression proving `SS006`, quota, and template-race INSERT failures make zero VMD restore/destroy calls
- Lock-orchestrated regression proving a concurrent legacy secret attach cannot be overwritten by the history backfill
- Focused `-race` tests for db/API/control-plane and `go vet`
- Linux builds for controlplane, vmd, boxd, template-builder, and secretsproxy
- Post-fix migration rehearsal on 100k sandboxes, 100k runtime snapshots, and 50k secret bindings: ~3.4s total; concurrency-safe secret-history backfill 2.76s; required composite index 0.13s; each catalog-only lock step <0.06s

## Pre-production gates

- Run the data preflight queries in staging (secret-history cardinality and team/lineage consistency).
- Confirm the manually repaired `20260716000001` migration-history state is absent/clean in staging.
- Repeat the timing rehearsal with staging row counts and monitor lock waits.
- Keep the feature flag dark through PRs 1–3.

### DIFF
--- a/db/queries/sandboxes.sql
+++ b/db/queries/sandboxes.sql
@@ -90,7 +90,7 @@ WITH tpl AS (
   SELECT ins.id, (@secret_ids::uuid[])[i], (@env_keys::text[])[i], (@proxy_tokens::text[])[i]
   FROM ins, generate_subscripts(@secret_ids::uuid[], 1) AS g(i)
 )
-SELECT ins.id, ins.team_id, ins.name, ins.status, ins.vcpu_count, ins.memory_mib, ins.host_id, ins.ip_address, ins.pid, ins.snapshot_id, ins.created_at, ins.updated_at, ins.destroyed_at, ins.network_config, ins.timeout_seconds, ins.metadata, ins.template_id, ins.snapshot_path, ins.mem_path, ins.base_path, ins.delta_path, ins.disk_mib, ins.auto_delete_seconds, ins.auto_delete_at
+SELECT ins.*
 FROM ins
 JOIN preview_policy ON preview_policy.sandbox_id = ins.id;
 
--- a/db/queries/sandboxes.sql
+++ b/db/queries/sandboxes.sql
@@ -1,8 +1,8 @@
 -- name: CreateSandbox :one
 -- ID is supplied by the caller (generated in Go via uuid.New()) rather
--- than defaulted in SQL, so the caller can parallelize this INSERT with
--- the VMD CreateVM call — both need the same sandbox_id and generating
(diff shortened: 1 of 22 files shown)


pavitrabhalla:
Closing per team decision — snapshot/restore is being redone from scratch with a filesystem-only, GCS-backed contract; see #287 for context.

PR #286, opened 2026-07-29
API: add saved snapshot control plane

mohamedsobhi777:
## Stack

**3 of 4** for saved snapshots and independent forks. Depends on #285 (and #284 beneath it). This PR adds the dormant control-plane/API implementation and reconciliation workers. The team-scoped feature remains disabled by default; PR4 supplies the hot indexes, public OpenAPI contract, full canary suite, and rollout gate.

## Product behavior

- A sandbox can capture an immutable saved snapshot with an optional 1–64 character name.
- A ready snapshot can be listed, deleted when it is a leaf, and forked into any number of independently writable sandboxes through `POST /sandboxes` with `from_snapshot`.
- Forks can inherit or replace network policy and secret identities; every child receives fresh writable storage, proxy tokens, JWTs, and access token material.
- Public/private preview defaults propagate through persistence, restore, live VMD re-attestation, and the response. Private forks require the complete browser-auth capability chain.

## Safety and lifecycle

- Adds atomic capture/fork/quota/storage-ledger queries with host-first lock order, source-operation fencing, idempotency, lineage pins, and exact-host artifact ownership.
- Re-attests the live VMD before injecting env/credentials or returning 201; rolling hosts that ignore new restore fields fail closed.
- Scrubs captured source-specific environment/token history and resolves inherited secret IDs against current non-revoked team secrets.
- Persists revocation before teardown and releases snapshot pins only after exact-host writable-layer cleanup.
- Adds bounded saved-snapshot reconciliation, drain/cleanup workers, writable-layer accounting, and allowlisted operational metrics.
- Makes team migration fail closed under global→team advisory fences, including cross-team system-template dependencies.
- Prevents reconciliation queue starvation with durable compare-and-swap deferral and immutable source lifecycle fencing.
- Preserves PR1's DB-first host admission: quota/`SS006`/template rejection performs zero VMD launch or destroy calls.

## Rollout posture

- `sandbox_snapshots_v1` remains globally off; an explicit team override is required.
- The API routes are present but intentionally exempted from the OpenAPI sync assertion until PR4 publishes the complete contract in the same stack.
- Staging Cloud Run keeps the reconciliation worker resident (`min=1`, CPU not throttled); production activation is still gated.
- Do not enable a team until PR4's indexes replace the legacy global runtime-snapshot uniqueness index and the staging canary passes.

## Validation

- `go run github.com/sqlc-dev/sqlc/cmd/sqlc@v1.30.0 generate` with a clean generated diff
[...]

PR #285, opened 2026-07-29
VMD: add saved snapshot data plane

mohamedsobhi777:
## Stack

**2 of 4** for saved snapshots and independent forks. Depends on #284. This PR adds the data-plane primitives and dormant staging storage machinery; it does not expose the public API or enable the feature.

## What changes

- Adds VMD capability negotiation plus immutable saved-snapshot capture, restore, delete, and writable-layer accounting.
- Validates contained manifest/artifact paths and SHA-256 digests before allocating a child.
- Restores every fork onto a fresh writable layer and scrubs captured source-specific environment defaults before exposure.
- Carries current preview access, explicit published ports, preview revision, and egress policy through restore and restart.
- Adds pidfd-backed process containment, reflink/extent accounting, sparse-copy handling, and Firecracker API connection cleanup.
- Persists disk size and egress in the rollback-stable VMD sidecar without regressing current preview-policy persistence.
- Adds a fail-closed XFS setup unit/script for the existing staging disk and stops boot-time deletion of runtime/snapshot trees.

## Storage activation posture

First activation is **manual-only**, staging-only, defaults off, and accepts exactly `/dev/disk/by-id/google-superserve-sandbox-data`. Push deploys cannot format or migrate a disk. The unit is enabled only after successful authority commit; pre-authority failures remove automatic retry state. Once a host is authoritative, ordinary deploys update and verify the boot-critical storage gate.

Do not run first activation until PR3's declarative host-drain worker is deployed and the host is proven drained. No storage workflow or staging resource was changed while preparing this PR.

## Compatibility and rollback

- Existing CreateSnapshot/RestoreSnapshot callers retain legacy semantics.
- New control planes must pass the capability handshake before using optional RPCs.
- The current preview RPCs, launcher hardening, and fresh-unit linger optimization are preserved.
- Keep the feature dark until PR3 and the PR4 index/canary gate.
- After restrictive forks exist, VMD/storage rollback is drain-first: an older VMD preserves but cannot re-enforce restrictive domain egress after restart.
- Rollback to PR1 retains its DB-first host-admission guarantee; rollback to a pre-PR1 API follows the reactivation gate in `deploy/README.md`.

## Validation

- Pinned VMD protobuf regeneration (`protoc` 34.1, `protoc-gen-go` 1.36.11, `protoc-gen-go-grpc` 1.6.2)
- 14 storage/deploy regression tests, Python compile check, `bash -n`, and actionlint
- Terraform fmt plus staging init/validate in an isolated temporary copy
[...]

PR #287, opened 2026-07-29
Release: gate saved snapshots for staging canary

mohamedsobhi777:
## Stack

Stack 4/4. Depends on #286.

- #284 — database expand and host-admission safety
- #285 — immutable VMD capture/restore/delete
- #286 — saved-snapshot control plane
- This PR — contract indexes, public API contract, regression coverage, and staging ship gate

## What this ships

- Prebuilds and verifies the exact 10-index saved-snapshot contract with sequential `CREATE INDEX CONCURRENTLY`; manual recovery is invalid-index-only and fail-closed.
- Adds bounded migrations `20260729000010` and `20260729000011`. Migration 11 requires the feature globally dark, zero enabled team overrides, zero saved/fork/operation residue, exact index definitions, and drops the legacy pause index atomically and last.
- Publishes the named snapshot/fork OpenAPI contract, including direct lineage, nullable inheritance rules, host/path opacity, dependency conflicts, retryable host errors, and private preview isolation.
- Adds bounded saved-snapshot telemetry without snapshot IDs, host paths, or raw errors.
- Restores the full saved-snapshot database/API/security/race/observability regression suite.
- Adds a strict staging-only canary: one temporary team override, exact-main API/VMD restart binding, idempotent capture/accounting, active and paused capture, three independent forks plus nested fork, memory/files/private-preview/secret/network isolation, dependency-safe deletion, and fail-safe dark cleanup.

## Rollout gates

- Keep `sandbox_snapshots_v1` globally disabled and all team overrides disabled while merging/deploying.
- PR3 must already be serving every traffic cell before migration 11 runs. After migration 11, PR3 is the application rollback floor because the legacy `snapshot_sandbox_unique` conflict target is gone.
- Drain the staging VMD host and confirm no Firecracker processes before the one-time allowlisted XFS activation.
- Run `snapshot-canary.yml` only from the current `main` SHA after all four PRs are merged and deployed. The workflow temporarily enables only the staging system team and must finish with global dark plus zero enabled overrides.
- This PR does not enable saved snapshots in production.

## Rollback

Database changes are leave-forward. Turn the team gate off first, drain/clean up snapshot workloads, and roll application/VMD code back only within the documented compatibility floor. Never recreate the legacy index once named snapshots or forks exist.

## Validation

- PostgreSQL 16: fresh history, populated prebuild/adoption, dark/residue contract refusal, exact invalid-index recovery, valid/wrong-table drift refusal, and advisory-lock overlap timeout.
- `go test -p 1 -tags integration -count=1 -race -timeout 10m ./internal/integration ./cmd/migrate-team`
- `go test -count=1 -race ./cmd/controlplane ./internal/api ./internal/db ./internal/supervisor`
- `go vet ./cmd/controlplane ./internal/api ./internal/db ./internal/supervisor`
- Bash syntax, YAML parse, actionlint, OpenAPI sync tests, SQLC generation check, and `git diff --check`.
- Canary harness fault tests: exact dispatch success/missing-ID fail-closed and disposable-PostgreSQL state restoration for absent, null, false, failure, and pre-enabled override cases.
- Exact-head CI at `5c653e221d9850a5b5b60e13c5d50207bea09b35`: https://github.com/superserve-ai/sandbox/actions/runs/30471382451 ✅

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -25,6 +25,10 @@ on:
     paths:
       - 'supabase/migrations/**'
 
+concurrency:
+  group: cd-migrate
+  cancel-in-progress: false
+
 jobs:
   migrate-staging:
     name: Migrate Staging
@@ -36,6 +40,25 @@ jobs:
       - uses: supabase/setup-cli@v1
         with:
           version: latest
+
+      - name: Ensure PostgreSQL client is available
+        run: |
+          if ! command -v psql >/dev/null 2>&1; then
+            sudo apt-get update
+            sudo apt-get install -y postgresql-client
+          fi
+
+      # These hot-table indexes must be built one at a time outside the
+      # transactional Supabase migration. The script also verifies their exact
+      # table, btree, uniqueness, key, predicate, and catalog state contract.
+      - name: Prebuild and verify saved-snapshot indexes in staging
+        run: >-
+          psql
+          --dbname="$DATABASE_URL"
+          --no-psqlrc
+          --file=scripts/prebuild-saved-snapshot-indexes.sql
+        env:
+          DATABASE_URL: ${{ secrets.DATABASE_URL_STAGING }}
 
       - name: Push migrations to staging
         run: supabase db push --db-url "$DATABASE_URL"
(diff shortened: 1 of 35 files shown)


pavitrabhalla:
Closing per team decision: we're rebuilding snapshot/restore from scratch around a filesystem-only, GCS-backed contract (resizable restores, artifacts durable and not host-pinned), so the memory-inclusive host-local design here won't ship. The schema/lineage modeling, quota/storage-ledger, and race-fencing work in this stack is solid prior art and will inform the redo.

PR #294, opened 2026-07-30
feat(api): add admin billing UI

awille-ai:
feat(api): add admin billing UI

Summary
- Add the platform billing listing path used by the admin UI.
- Reconstruct current-period available credits from the billing ledger so `credits_applied_usd` and `credits_remaining_usd` do not double-apply consumption that is already recorded for the open period.
- Count platform billing search length by Unicode characters so the documented 200-character limit matches multibyte input.
- Keep the existing row-level partial-failure behavior and deterministic pagination.

Testing
- `docker run --rm -v /Users/lando/superserve/wip/sandbox-addBillingUI:/src -v /private/tmp/gomodcache:/go/pkg/mod -v /private/tmp/gocache:/root/.cache/go-build -w /src golang:1.26 go test ./... -count=1`


awille-ai (commented):
I reviewed the current draft head. The authorization layering, row-level partial-failure behavior, deterministic pagination, and test coverage are directionally strong. I left this as a comment-only review because this is your PR.

One accounting concern looks worth addressing before this becomes the source for the admin UI: the response derives `credits_applied_usd` and `credits_remaining_usd` by applying the current open-period charges against `team_credit_grant.remaining_usd`. That column is already mutated when finalized periods consume credits. For the active period that is fine if no current-period credits have been persisted yet, but the endpoint has no explicit guard tying that assumption to an unfinalized period or excluding ledger entries for the same period. If a period is partially/retried-finalized, imported, or otherwise has credit consumption recorded while remaining open, the UI can subtract the same charges from an already-reduced grant balance and understate both credits applied and credits remaining.

Could we make the invariant explicit in the query or data model—for example, calculate available credits at the period start by adding back ledger consumption for this period, or source the displayed credit application from the period/ledger when it exists? I would also add an integration case where a grant has prior-period consumption and current-period ledger activity, so the admin totals cannot silently double-apply credits.

CI and Terraform checks are still in progress on this head, so I did not treat their status as evidence either way.

awille-ai:
@codex review

awille-ai:
@codex review

awille-ai (commented):
I compared the current head against SS-244. The implementation now covers the core acceptance criteria: a dedicated internal endpoint, `platform:billing:read`, no tenant impersonation/API-key loop, server-side pagination/search/sorting, page-independent aggregate totals, and row-level pricing failures. The tenant `/billing/summary` path is unchanged.

One remaining implementation concern: both platform billing query variants duplicate the tenant billing calculation in a 600+ line handler constant. That makes the new platform endpoint likely to drift from `/billing/summary` as billing-period selection, pricing, credit application, and usage calculation evolve. We already had to repair the credit-ledger behavior inside this PR, which demonstrates the risk. Please move the common billing-summary calculation into a shared SQL query/view/function or shared repository service that both tenant and platform paths consume, leaving this endpoint responsible for cross-team filtering, ordering, pagination, totals, and per-team error shaping. At minimum, add a parity integration test that seeds one team and asserts its successful platform row matches the tenant `/billing/summary` response field-for-field.

meAmitPatil (approved):

PR #295, opened 2026-07-30
fix(terraform): add cloud ids rollout

awille-ai:
fix(terraform): add cloud ids rollout

- Add Cloud IDS endpoints to `us-east4` and `us-west2`.
- Bootstrap the shared Cloud IDS project services from `us-central1` before the regional applies.
- Document the Cloud IDS rollout, validation steps, and Vanta network disposition.

Validation:
- `terraform fmt -check -recursive infra`
- `git diff --check`

Notes:
- The branch was trimmed to Cloud IDS-only changes.
- No unrelated rollout, IAM, or account changes remain in the diff.

### DIFF
--- a/.github/workflows/terraform-cd.yml
+++ b/.github/workflows/terraform-cd.yml
@@ -201,11 +201,56 @@ jobs:
           scripts/smoke-test-region.sh staging us-central1
           echo "staging/us-central1 API deploy and smoke validation passed." >> "$GITHUB_STEP_SUMMARY"
 
+  production-us-central1-bootstrap:
+    name: Bootstrap production/us-central1 project services
+    runs-on: ubuntu-latest
+    environment: production
+    needs: [staging-us-central1-infra]
+    permissions:
+      contents: read
+      id-token: write
+    env:
+      TF_IN_AUTOMATION: 'true'
+      TF_VAR_notification_channel_ids: ${{ vars.TF_VAR_NOTIFICATION_CHANNEL_IDS || '[]' }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_wrapper: false
+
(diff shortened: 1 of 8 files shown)


awille-ai (commented):
I reviewed the current draft head with the same architecture/correctness/security/reliability bar. The Cloud IDS module is generally well structured: regional endpoint placement, explicit packet-mirroring scope, alert-policy preconditions, shared notification channels, and production-state serialization are all sensible.

I found one workflow issue worth fixing before rollout. In `.github/workflows/terraform-rollout-production.yml`, both `west2` and `east4` declare `needs: [bootstrap]`, but their `if` expressions are ordinary conditions rather than `always()`-guarded conditions. GitHub evaluates the default dependency-success requirement before the job-level condition, so a skipped `bootstrap` causes dependent jobs to be skipped as well. Today `bootstrap` is selected for all three valid targets, so this is not observable in the current matrix, but it makes the workflow brittle: any future target or condition that intentionally skips bootstrap will silently suppress the regional rollout. More importantly, the previous workflow used `always()` explicitly for this reason.

I recommend making the dependency semantics explicit, e.g. `if: ${{ always() && github.ref == ... && needs.bootstrap.result == 'success' }}` for both region jobs. That prevents accidental skip propagation and documents that a failed prerequisite blocks rollout while an unrelated skip does not.

A second, smaller concern: the PR mixes Cloud IDS work with a fairly broad runtime-IAM/service-account migration in `us-west2` (`superserve-api-runtime`, `superserve-sandbox-runtime`, secret IAM moves, KMS verification, and `allow_stopping_for_update`). Those changes may be valid, but they materially expand blast radius and make rollback harder to reason about. I would strongly consider splitting them into a separate PR unless they are strictly required for the IDS rollout.

Terraform Plans and Terraform Checks are green. CI was still in progress at the time of review.

awille-ai:
@codex review

meAmitPatil (approved):

PR #296, opened 2026-07-30
api: keep host capability and preview policy reads off the request hot paths

meAmitPatil:
The capability pre-flight on create/resume ran a FOR SHARE read against the host row on every request, so concurrent creates serialized behind the host's heartbeat writer; activate and get-sandbox paid the preview policy read as an extra serial round-trip.

- Add an unlocked variant of the capability query for standalone pre-flight reads; the transactional mutation path keeps the locking variant to pin the host across its commit.
- Serve pre-flights from a positive-only, TTL-bounded attestation cache (singleflight on miss, `HOST_CAPABILITY_CACHE_TTL=0` disables). Negatives and errors are never cached, so a capability 409 is always based on a fresh read; VMD's post-boot policy attestation is unchanged and remains the fail-closed gate.
- Overlap the preview policy read with the sandbox load on activate and get-sandbox instead of running it serially.

Tests: cache behavior (positive hit, negative never cached, per-capability-set keys, expiry/eviction, kill switch) and a query-structure test pinning that the unlocked variant runs the same relational division without the row lock.

### DIFF
--- a/db/queries/hosts.sql
+++ b/db/queries/hosts.sql
@@ -80,6 +80,34 @@ SELECT EXISTS (
   )
 );
 
+-- name: HostHasCapabilitiesUnlocked :one
+-- HostHasCapabilities without the row lock, for standalone pre-flight reads
+-- outside a mutation transaction: omitting the lock keeps concurrent checks
+-- from serializing behind the host's heartbeat writer. Transactional callers
+-- that must pin the host across a commit use HostHasCapabilities.
+WITH target_host AS MATERIALIZED (
+  SELECT id, last_heartbeat_at
+  FROM host
+  WHERE id = sqlc.arg('host_id')
+    AND status = 'active'
+    AND last_heartbeat_at IS NOT NULL
+)
+SELECT EXISTS (
+  SELECT 1
+  FROM target_host h
+  WHERE NOT EXISTS (
+    SELECT 1
+    FROM unnest(sqlc.arg('required_capabilities')::text[]) AS required(capability)
+    WHERE NOT EXISTS (
+      SELECT 1
+      FROM host_capability hc
+      WHERE hc.host_id = h.id
+        AND hc.capability = required.capability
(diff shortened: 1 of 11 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
Reviewed at the latest head. The hot-path changes preserve the important fail-closed boundaries while removing avoidable contention and serial reads.

The unlocked capability query is appropriately limited to standalone pre-flight checks; mutation paths still use the locked query, and VMD's post-boot attestation remains authoritative. The positive-only cache has bounded freshness, does not cache negatives or errors, supports cancellation for individual waiters via `DoChan`, and now sweeps retired-host entries. The stale grace is short and only affects a pre-flight gate, with downstream transactional/VMD enforcement still protecting correctness.

The preview-policy consolidation also looks sound: active reads return the policy from the same DB snapshot as the sandbox row, while resume returns the post-restore policy actually pushed to VMD. Tests cover cache keys, expiry, negative behavior, kill switch, query routing, and schema compatibility.

CI and Terraform Checks are green. Approved.

pavitrabhalla (approved):
Nits, non-blocking:

- handlers_test.go sandboxRow: branches on len(dest) == 25 to decide whether to scan the trailing access column for GetSandboxWithPreviewPolicy. Works and is pinned by TestActivateSandbox_AlreadyActive_200WithSandboxResponse, but a magic column count is fragile if Sandbox gains another field. Consider a named param or split helper.
- The `(strings.Contains(sql, "...HostHasCapabilities...") || strings.Contains(sql, "...Unlocked..."))` pattern is repeated across ~10 test files. A small isCapabilityCheck(sql string) bool helper would de-dupe it.

Locked/unlocked query parity, cache fail-open scope, and query routing all check out and are well-pinned by tests.

PR #299, opened 2026-07-30
vm: harden resume env injection and stop leaking the VM when it fails

meAmitPatil:
Resume made a blocking `/init` call even with zero env vars, just to re-stamp the guest hostname — and a transient in-guest stall on that path could fail an otherwise-healthy resume, leaving the VM running behind a reverted record.

- Hostname-only resumes no longer block: the stamp is a background best-effort with a warn log on failure.
- Env-var injection gets a 15s budget (was 5s) and one retry; `/init` is idempotent.
- If injection still fails, `AbortResume` stops the unit and reverts the record to Paused, so host and control plane agree and a retry does a fresh restore.
- boxd logs split timings when a hostname set is unexpectedly slow, to pinpoint any future stall.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
Reviewed the latest head with the same architecture, correctness, security, scalability, and reliability bar.

The previous failure modes are addressed: zero-env resumes still gate on boxd health before success; background hostname stamping is ownership-checked against both the instance record and network slot; env injection is serialized with lifecycle operations and retries; and abort-on-failure now runs while holding the VM lifecycle lock, stops the unit, returns the record to Paused, and avoids resurrecting a concurrently destroyed record.

The retry path also revalidates IP ownership before redelivery, which is important because the payload can contain credentials. The explicit no-keepalive transport and bounded timeouts are appropriate for the recycled-IP model. CI and Terraform Checks are green.

Approved.

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
Re-reviewed the latest head at the same architecture, correctness, security, concurrency, rollout, and reliability bar.

The new destroy tombstone and non-reattaching tracked-instance lookups close the lazy-reattach resurrection window. `vmOwnsIP` now checks both the tracked Running instance and the network manager's authoritative slot ownership, so `/init` retries and the detached readiness probe cannot trust a recycled IP. The resume lifecycle lock remains held across restore, readiness, injection, and abort; `abortResumeLocked` is non-resurrecting and persists only if the record still exists. The 30-second detached boxd readiness budget is consistent with the existing cold-fault tolerance and now revalidates ownership before reporting success.

The prior review threads are resolved, the relevant mid-destroy and ownership regressions are covered by tests, and CI plus Terraform Checks are green.

Approved.

PR #301, opened 2026-07-30
vmd: attest preview policy in the restore response instead of a separate round-trip

meAmitPatil:
Create issued a second CP-to-vmd RPC after every boot whose only purpose on that path was attestation: the restore request already carries the preview policy fields, but they are additive, so an older vmd would ignore them silently and the extra call forced a loud failure.

- RestoreSnapshotResponse gains preview_protocol: a policy-aware vmd echoes the enforced capability name after validating and applying the request's policy fields.
- The create path checks the echo instead of issuing UpdateSandboxPreviewPolicy — one RPC per create instead of two, same fail-closed behavior (a missing attestation tears the VM down before any 201-visible side effect).
- Resume is unchanged: its post-restore reapply stays, since it also reconciles a policy mutation that raced the restore. The RPC itself remains for the mutation and resume paths.

Version skew is safe in both directions: a new control plane on an old vmd sees an empty echo and fails closed exactly as the RPC's Unimplemented did; an old control plane ignores the new response field and keeps issuing the RPC, which vmd still serves. Both sides reference the same capability constant, so the attestation value cannot drift.

Tests: create must not issue the separate policy RPC; a boot without the attestation fails closed (teardown, no env injection, 500); retry-helper coverage widened to the new return.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

pavitrabhalla:
Re-reviewed after the skew-handling commit. The P1 above (destroying every booted sandbox during a same-SHA rollout when the API version ships ahead of vmd) is resolved: an empty echo now falls back to `UpdateSandboxPreviewPolicy` before failing closed, and the fallback path is covered by `TestCreateSandbox_PreviewAttestationSkewAndFailClosed`.

Checked the rest of the diff: `preview.HostCapabilityPorts` is a single shared constant referenced by both control plane and vmd, so the echo can't drift from the capability gate at scheduling time; `deploy-vmd.yml` picking up `proto/**` closes a real gap (api already had it). `retryTransientBoot` calls attempts sequentially so the `previewProtocol` closure capture is race-free.

Verified `go build`/`vet` clean under `GOOS=linux` (including `-tags integration`), and `internal/api`/`internal/telemetry` tests pass on 87bb982. No other issues found.

pavitrabhalla (approved):
Verified: the rollout-skew fail-closed issue is resolved with a tested fallback, build/vet clean, tests pass.

PR #309, opened 2026-07-31
fix(terraform): add required GCP compute labels

awille-ai:
## Summary
- Add the required Compute Engine labels on each Terraform-managed sandbox host instance
- Keep role-specific labels on the host modules without using provider-wide defaults
- Update the infra inventory to reflect the live host set and the decommissioned us-central1 production host

## Notes
- This is a direct implementation of the required GCP Compute Engine labels for Vanta compliance
- Required labels are applied at the instance label map level so the values match each host's actual role

## Validation
- `terraform fmt` on the edited `.tf` files
- `terraform init -backend=false -input=false` in `infra/envs/staging/us-central1`
- `terraform validate` is still blocked here by the Google provider plugin handshake in this sandbox

### DIFF
--- a/infra/README.md
+++ b/infra/README.md
@@ -5,32 +5,49 @@ Terraform skeleton layout for Superserve GCP adoption.
 - `modules/` defines resource-area contracts
 - `envs/` wires those contracts into staging, production, and a greenfield new-region target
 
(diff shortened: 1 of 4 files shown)


awille-ai:
@codex review

meAmitPatil (approved):

PR #312, opened 2026-07-31
sandbox: index base_path for destroy-time artifact GC lookups

meAmitPatil:
CountActiveSandboxesAtBasePath runs on every destroy (build-artifact GC refcount) and ListPinnedBuildPaths feeds the reconciler; neither predicate is indexed, so both sequential-scan the sandbox table, and the GC check runs inside the delete request path.

Adds a partial index on `sandbox(base_path)` restricted to live rows with a base_path — the only rows either query can match. `lock_timeout` bounds the wait behind long-running sandbox writers so a queued CREATE INDEX can't stall sandbox writes; the migration is idempotent and safe to re-apply.

### DIFF
--- a/supabase/migrations/20260731000005_sandbox_base_path_index.sql
+++ b/supabase/migrations/20260731000005_sandbox_base_path_index.sql
@@ -28,6 +28,21 @@ BEGIN;
 SET LOCAL lock_timeout = '5s';
 SET LOCAL statement_timeout = '10s';
 
+-- IF NOT EXISTS matches by name only: an interrupted concurrent pre-build
+-- leaves an INVALID index that would be silently kept, recording this
+-- migration as applied while both queries still sequential-scan. Fail the
+-- push instead; the fix is DROP INDEX + re-run the pre-build above.
+DO $$
+BEGIN
+  IF EXISTS (
+    SELECT 1 FROM pg_index
+    WHERE indexrelid = to_regclass('public.idx_sandbox_base_path_active')
+      AND NOT indisvalid
+  ) THEN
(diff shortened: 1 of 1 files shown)


meAmitPatil:
@codex review

meAmitPatil:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

meAmitPatil:
Addressed in 7e3bc2c6. `CREATE INDEX CONCURRENTLY` cannot run in this pipeline: migrations apply via `supabase db push`, which wraps each file in a transaction (no migration in this repo uses CONCURRENTLY for that reason). The restructure uses the standard two-step for hot tables instead: the index is pre-built `CONCURRENTLY` by hand before merge (exact SQL + INVALID-index verify/retry documented in the migration header), making the migration itself a no-op schema record on populated databases and an instant build on fresh ones. If the pre-build is ever skipped, `statement_timeout = '10s'` + `lock_timeout = '5s'` fail the push loudly with a bounded stall instead of blocking sandbox writes for an unbounded build.

meAmitPatil:
@codex review

meAmitPatil:
Addressed in 46a95080: the migration now queries `pg_index.indisvalid` and raises if the prebuilt index exists but is invalid, so the push fails loudly instead of recording the migration over an unusable index. No index → builds; valid → no-op; invalid → deploy fails with the drop/rebuild instruction in the error.

meAmitPatil:
@codex review

pavitrabhalla (approved):
Assuming you have created the index concurrently in prod DB manually, this is good to go.

PR #313, opened 2026-07-31
quota: widen the sandbox counter to 64 shards

meAmitPatil:
Concurrent creates for one team queue on the quota counter's shard row locks — each transaction holds its shard row until commit, so burst admissions serialize in chains of burst_size/16. Widening the modulus to 64 cuts the expected chain depth 4x.

Enforcement, the drift check, and the console all read the shard SUM, which is shard-count agnostic — verified across both repos. Rows for shards 16-63 appear lazily on first use. Rollback is re-applying the previous function bodies; leftover wide-shard rows keep being summed, so totals stay exact in both directions.

The function bodies are verbatim from the sharded-counter migration except the modulus (the per-transaction shard choice and its batch-deadlock rationale are preserved). Atomic function swap — no manual pre-step, no config changes, no control-plane deploy.


meAmitPatil:
@codex review

awille-ai (approved):
Reviewed the latest head at the same architecture, correctness, security, scalability, and reliability bar.

The migration is narrowly scoped to widening the per-transaction quota-counter modulus from 16 to 64. Enforcement remains shard-count agnostic because all authoritative reads sum every row for the team; new shard rows are created lazily, and rollback to the prior function bodies remains count-correct because rows 16-63 continue to be included in that sum.

The near-limit advisory-lock path is unchanged, so exact enforcement and rollback-on-quota-error remain transactional. Admission and release use the same widened modulus, avoiding systematic counter skew, and the function replacement is atomic within the migration transaction. There are no unresolved review threads.

CI and Terraform Checks are green. Approved.

meAmitPatil:
@codex review

PR #316, opened 2026-08-01
db: expose home_region on team spend/sandbox analytics views

pavitrabhalla:
## Summary
- Filters `analytics.team_hourly_spend` and `analytics.team_sandbox_events` to teams still attached to the cell they're queried from, so a cell's Grafana panels stop ranking usage from teams that have since migrated away.

## Why
A team that migrates between cells keeps its historical usage rows in the source cell's database as a cold fallback until purge (see `cmd/migrate-team`). The "top 10 teams by spend/sandbox count" rank charts kept ranking that historical usage alongside genuinely-homed teams.

**Revised from the original approach** (see review thread below): the first version filtered on `team.home_region`, but `cmd/migrate-team`'s detach step never updates the source row's `home_region` — that column isn't part of `membershipTables` and detach deliberately leaves it untouched, along with the rest of the team's data, until purge. So during the normal soak period a detached team's source-side `home_region` still reads the old region, and filtering on it wouldn't actually exclude anything in the default case. Filters on `team_member` presence instead — the same "is this team still live here" signal `cmd/migrate-team`'s own `sourceDetached` helper checks.

## Testing
- Verified via direct query: a team confirmed detached (zero rows in `team_member`/`team_memberships`/`user_role_assignments`) is excluded from the filtered view while other teams' numbers are unchanged.
- No Grafana panel changes needed this time — filtering lives in the view itself, so each cell's queries are automatically scoped without an extra `WHERE home_region = ...` per panel.

### DIFF
--- a/internal/db/models.go
+++ b/internal/db/models.go
@@ -180,16 +180,14 @@ type AnalyticsDailySandboxStart struct {
 }
 
 type AnalyticsTeamHourlySpend struct {
-	TeamName   string         `json:"team_name"`
-	HomeRegion string         `json:"home_region"`
-	HourStart  time.Time      `json:"hour_start"`
-	SpendUsd   pgtype.Numeric `json:"spend_usd"`
+	TeamName  string         `json:"team_name"`
+	HourStart time.Time      `json:"hour_start"`
+	SpendUsd  pgtype.Numeric `json:"spend_usd"`
 }
 
 type AnalyticsTeamSandboxEvent struct {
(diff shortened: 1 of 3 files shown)


pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

PR #321, opened 2026-08-03
fix(api): abort the activate settle-poll immediately on client disconnect

chill-czar:
## Summary

- Replace `time.Sleep(activateSettlePoll)` in `loadActiveOrResumeSandbox` with a `select` racing `time.After` against `c.Request.Context().Done()`.
- Add unit test `TestActivateSandbox_ClientDisconnectDuringSettle` in `internal/api/handlers_test.go` to verify immediate abort when client disconnects during the settle-poll window.

## Why

The polling loop in `loadActiveOrResumeSandbox` waits for a starting or resuming sandbox to flip to `active`. Previously, it used `time.Sleep(activateSettlePoll)` unconditionally. If a client disconnected during this settle window, the server goroutine was blocked for the full sleep duration before noticing the context was cancelled. Utilizing a `select` statement on `ctx.Done()` ensures the handler exits immediately upon client disconnection, freeing up server resources.

## Test Plan

- [x] Unit tests pass: `go test ./internal/api/ -short -run 'TestActivateSandbox' -v` passed


chill-czar:

Could you please take a look at this PR when you have a moment?

**Context:** This change replaces `time.Sleep` in `loadActiveOrResumeSandbox` with a `select` statement that listens on `c.Request.Context().Done()`. This allows the handler to abort immediately if a client disconnects while waiting out the settle window, avoiding blocked server goroutines.

Thanks!

meAmitPatil:
@codex review

meAmitPatil (changes requested):
Thanks for the fix — the settle-poll absolutely should be cancel-aware, and the `select` on the request context is the right shape. Two changes needed before this can merge:

**1. Build break after syncing with main**

`loadActiveOrResumeSandbox` now returns `(*db.Sandbox, string)` on main, so the new branch's `return nil` no longer compiles (CI: `internal/api/handlers.go:554: not enough return values`). It needs to be `return nil, ""`, matching the other error exits in the function.

**2. Don't respond with `ErrInternal` on disconnect**

A cancelled request context here means the caller hung up — that's expected client behavior, not a server fault, so a 500 is the wrong classification. It also inflates 5xx metrics and trips the error-level request logging for what is a routine event. There's an existing precedent for this exact situation in `handlers_template.go` (the poll loop there bare-returns on `ctx.Done()` without writing a response, since there's no client left to read it). So this case can simply be:

```go
case <-c.Request.Context().Done():
    return nil, ""
```

Minor, optional: the new test only asserts elapsed time. Once the response-writing is dropped, asserting that no body/status was written (or just keeping the timing assertion) is fine — but if you keep any response, please assert `w.Code` like the neighboring `TestActivateSandbox_*` tests do.

Happy to merge once those two are in.

chill-czar:
Thanks @meAmitPatil! Good points.

I've updated the PR:
- Rebased on `upstream/main` and updated `loadActiveOrResumeSandbox` to bare-return `return nil, ""` on `ctx.Done()`, fixing the build break and avoiding the unnecessary 500 error response.
- Verified compilation and tests pass locally.

chill-czar:
Also updated `TestActivateSandbox_ClientDisconnectDuringSettle` in `handlers_test.go` to explicitly assert `w.Body.Len() == 0`, confirming no error response body is written on client disconnect.

PR #322, opened 2026-08-03
fix(api): give each DB call in cleanupExpiredRevocations its own timeout

chill-czar:
## Summary

- Give `DeleteExpiredSandboxRevocations` and `DeleteExpiredRevokedProxyTokens` independent 5-second timeout contexts (`revokeCtx` and `tokenCtx`) in `cleanupExpiredRevocations`.
- Add unit test `TestCleanupExpiredRevocations_IndependentTimeouts` in `internal/api/reaper_test.go`.

## Why

Both DB cleanup operations previously shared a single 5-second timeout context. If `DeleteExpiredSandboxRevocations` consumed most of the 5-second budget (e.g., under DB contention), `DeleteExpiredRevokedProxyTokens` received a near-zero remaining deadline and failed with `DeadlineExceeded`. Providing independent contexts ensures neither cleanup step starves the other regardless of execution duration.

## Test Plan

- [x] Unit tests pass: `go test ./internal/api/ -short -run 'TestCleanupExpiredRevocations' -v` passed

### DIFF
--- a/internal/api/reaper.go
+++ b/internal/api/reaper.go
@@ -114,9 +114,9 @@ func (h *Handlers) sweepOrphanedIntervals(ctx context.Context) {
 // cleanupExpiredRevocations deletes sandbox_revocation and revoked_proxy_token
 // rows past their TTL — once no live JWT can carry them, the entries are moot.
 func (h *Handlers) cleanupExpiredRevocations(ctx context.Context) {
-	queryCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
-	defer cancel()
-	n, err := h.DB.DeleteExpiredSandboxRevocations(queryCtx)
+	revokeCtx, revokeCancel := context.WithTimeout(ctx, 5*time.Second)
+	defer revokeCancel()
+	n, err := h.DB.DeleteExpiredSandboxRevocations(revokeCtx)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


chill-czar:

Could you please review this PR when you get a chance?

**Context:** This fix updates `cleanupExpiredRevocations` so that `DeleteExpiredSandboxRevocations` and `DeleteExpiredRevokedProxyTokens` each get their own independent 5-second timeout context. Previously, sharing a single context meant if the first query took time, the second query would receive a near-zero deadline and fail with a timeout error.

Thanks!

meAmitPatil (changes requested):
Thanks — the fix is right and matches how the rest of this file scopes per-call timeouts.

One change requested: the new test passes even against the old shared-context code — the mock returns instantly, so both deadlines land ~5s out either way. Capture the two contexts' deadlines in the mock and assert `d2.After(d1)` instead; identical deadlines = shared context, no sleeps needed.

chill-czar:
Thanks @meAmitPatil! Good call on the test assertion.

Updated `TestCleanupExpiredRevocations_IndependentTimeouts` in `reaper_test.go` to capture `firstDeadline` and `secondDeadline` in the mock `execFn` and assert `secondDeadline.After(firstDeadline)`.

I verified that the updated test fails against the old shared-context implementation and passes against the new code.

PR #323, opened 2026-08-03
fix(api): update email subject line for quota alerts

chill-czar:
## Summary

- Change quota alert email subject line in `EmailQuotaNotifier.Notify` from `"Verify your team on Superserve"` to `"Sandbox limit reached for your team"`.
- Update subject assertion in `TestEmailQuotaNotifierSends` in `internal/api/quota_watcher_test.go`.

## Why

The quota alert email previously used the onboarding/verification subject line `"Verify your team on Superserve"` due to copy-paste during initial template creation. Updating it to `"Sandbox limit reached for your team"` correctly conveys the notification purpose to team owners when active sandbox thresholds are reached.

## Test Plan

- [x] Unit tests pass: `go test ./internal/api/ -short -run 'TestEmailQuotaNotifier' -v` passed

### DIFF
--- a/internal/api/quota_watcher_test.go
+++ b/internal/api/quota_watcher_test.go
@@ -134,8 +134,8 @@ func TestEmailQuotaNotifierSends(t *testing.T) {
 	if len(payload.To) != 1 || payload.To[0] != "owner@acme.test" {
 		t.Errorf("to = %v, want [owner@acme.test]", payload.To)
 	}
-	if !strings.Contains(payload.Subject, "Verify your team") {
-		t.Errorf("subject = %q, want it to mention verification", payload.Subject)
+	if !strings.Contains(payload.Subject, "Sandbox limit reached") {
+		t.Errorf("subject = %q, want it to mention limit reached", payload.Subject)
 	}
 	if !strings.Contains(payload.HTML, "Verify your team on Superserve") || !strings.Contains(payload.HTML, "Book a Meeting") {
 		t.Errorf("html missing verification heading/CTA: %q", payload.HTML)
--- a/internal/api/quota_watcher_test.go
+++ b/internal/api/quota_watcher_test.go
@@ -137,8 +137,8 @@ func TestEmailQuotaNotifierSends(t *testing.T) {
 	if !strings.Contains(payload.Subject, "Sandbox limit reached") {
 		t.Errorf("subject = %q, want it to mention limit reached", payload.Subject)
 	}
-	if !strings.Contains(payload.HTML, "Verify your team on Superserve") || !strings.Contains(payload.HTML, "Book a Meeting") {
-		t.Errorf("html missing verification heading/CTA: %q", payload.HTML)
+	if !strings.Contains(payload.HTML, "Sandbox limit reached for your team") || !strings.Contains(payload.HTML, "Book a Meeting") {
+		t.Errorf("html missing quota limit heading/CTA: %q", payload.HTML)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


chill-czar:

Could you please take a look at this PR when you have a moment?

**Context:** This updates the email subject line in `EmailQuotaNotifier.Notify` from `"Verify your team on Superserve"` to `"Sandbox limit reached for your team"`. It fixes a copy-paste mismatch so team owners receive an accurate subject line when their sandbox threshold is reached.

Thanks!

internal/api/quota_watcher_test.go
```diff
@@ -134,8 +134,8 @@ func TestEmailQuotaNotifierSends(t *testing.T) {
 	if len(payload.To) != 1 || payload.To[0] != "owner@acme.test" {
 		t.Errorf("to = %v, want [owner@acme.test]", payload.To)
 	}
-	if !strings.Contains(payload.Subject, "Verify your team") {
-		t.Errorf("subject = %q, want it to mention verification", payload.Subject)
+	if !strings.Contains(payload.Subject, "Sandbox limit reached") {
+		t.Errorf("subject = %q, want it to mention limit reached", payload.Subject)
 	}
 	if !strings.Contains(payload.HTML, "Verify your team on Superserve") || !strings.Contains(payload.HTML, "Book a Meeting") {
```
pavitrabhalla: you may want to update the string here as well.

pavitrabhalla (commented):
Thanks for the fix, catching the copy-paste subject line is the right call. This is directionally correct but incomplete: it only changes the `subject` field, `quotaEmailHTML` (quota_watcher.go:368-378) still renders the old copy for the actual email.

Specifically: the `<h1>` still reads "Verify your team on Superserve", and the `emailShell` preview/preheader text (quota_watcher.go:378) is the same string, that's the snippet Gmail/etc. show right next to the subject in the inbox. So recipients would see subject "Sandbox limit reached for your team" paired with a preheader saying "Verify your team on Superserve", a visible contradiction in the inbox row itself, not just something buried in the body. The body copy below it ("we limit the number of active sandboxes on new accounts...") is also from the verification flow, `QuotaAlert` is a general usage-threshold alert, not account verification, so worth checking whether that copy still applies at all.

Can you update the h1/preheader/body copy in `quotaEmailHTML` to match the new subject? That'd also make the html assertion you already flagged actually meaningful, right now it still passes because the body genuinely wasn't touched.

chill-czar:
Thanks for catching that @pavitrabhalla! Good spot on the inbox preheader and heading mismatch.

Just pushed a commit to fix it:
- Updated the `<h1>` and `emailShell` preview string in `quotaEmailHTML` to "Sandbox limit reached for your team".
- Updated `TestEmailQuotaNotifierSends` to check for the new string in the HTML body.

internal/api/quota_watcher_test.go
chill-czar: Updated this string as well in the latest commit, thanks!

meAmitPatil (commented):
Thanks — subject, heading, and preheader are consistent now. Two small copy points before merge:

1. This alert actually fires at 80% of the limit (`quotaAlertPct`), not at the limit — so "Sandbox limit reached" can be premature for a team at 16/20. Suggest "You're approaching your sandbox limit" (subject, h1, and preheader) so it's accurate across the whole range.
2. The closing sentence still says "this verification helps us allocate capacity" — nothing in the email mentions verification anymore. "this review helps us allocate capacity" reads better.

Good to merge after that.

chill-czar:
Thanks @meAmitPatil! Makes total sense since this fires at 80%.

Updated the PR:
- Updated subject, `<h1>`, and `emailShell` preheader to "You're approaching your sandbox limit".
- Updated closing sentence to "this review helps us allocate capacity".
- Updated test assertions in `quota_watcher_test.go`.

PR #326, opened 2026-08-04
api: cap concurrent activity-log inserts so bursts cannot crowd the DB pool

meAmitPatil:
Every sandbox lifecycle operation fires fire-and-forget activity-log inserts at the moment it runs, so operation bursts land dozens of bookkeeping statements on the connection pool exactly when the create path is queueing for the same connections.

All async activity writes now funnel through a shared writer holding a 2-permit gate: at most two activity inserts occupy connections at once, the rest wait in-process (free — they are already background goroutines) and shed with an attributable log if the pool stays saturated past their deadline. This also de-duplicates three near-identical goroutine bodies into one gated writer with uniform structured error logs.

The preview token-rotation audit write deliberately bypasses the gate: it is synchronous in the request by design (audit-before-response on a committed revocation) and must not queue behind bulk lifecycle logging.

No API, schema, or config changes. Unit test covers the gate's bound, deadline give-up, and release semantics.

### DIFF
--- a/internal/api/activity_gate_test.go
+++ b/internal/api/activity_gate_test.go
@@ -1,0 +1,32 @@
+package api
+
+import (
+	"context"
+	"testing"
+	"time"
+)
+
+func TestActivityGateBoundsConcurrency(t *testing.T) {
+	h := &Handlers{}
+	ctx := context.Background()
+
+	for i := 0; i < activityWriteConcurrency; i++ {
+		if !h.acquireActivityGate(ctx) {
+			t.Fatalf("acquire %d should succeed immediately", i+1)
+		}
+	}
+
+	// With all permits held, the next acquire must wait — and give up at the
+	// caller's deadline rather than blocking forever.
+	deadlineCtx, cancel := context.WithTimeout(ctx, 20*time.Millisecond)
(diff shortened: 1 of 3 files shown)


meAmitPatil:
@codex review

pavitrabhalla (approved):
Reviewed the diff and walked through the gate lifecycle in handlers.go and the preview bypass in handlers_preview.go.

Gate acquire/release is race-free (sync.Once init, release only deferred after a successful acquire, no permit leak on panic since writeActivity's defer unwinds before sentrylog.Recover catches it). Gate wait and DB call share the same ctx, so a slow wait correctly eats into the asyncTimeout budget rather than extending it. Confirmed logPreviewTokenRotationActivity is called directly (not in a goroutine), so the sync-in-request bypass claim holds.

Test covers cap enforcement, deadline give-up, and release-unblocks-waiter, good enough for the gate logic. CI is green across the board.

One non-blocking thought: activityWriteConcurrency is a hardcoded const rather than a var like activateSettleWindow above it. If this cap ever needs tuning after a burst test, might be worth making it overridable, but not a reason to hold this up.

Approving.

PR #327, opened 2026-08-05
vmd: release a VM's slot only once its unit is terminal, and finish releases that were abandoned

meAmitPatil:
The reconciler releases a VM's namespace, IP and tap device when it reaps one.
That release could go wrong two ways, and nothing caught either.

**Too early.** A nil `stopUnit` only means the stop job finished, and absence
from the active-unit snapshot is not terminal — either way Firecracker may still
hold the tap, so the next VM can be launched onto a device the old process owns.

**Not at all.** A failed BoltDB delete abandons the release with the record and
its slot still held, and nothing revisits it: every rule that reaps a live VM
matches on an ACTIVE unit, which its own stop has already retired.

Six of the eight `markStale` call sites had one or both. Only Drift 8 gated on
`unitFullyDown`.

- `markStale` remembers an incomplete release; `retryPendingReleases` finishes it
  on a later pass. Recorded inside `markStale`, so every caller inherits it.
- `markStale` refuses to release while the unit still has a process, returning a
  deferral rather than a failure. Fail-closed.
- Callers route through `finalizeRelease`, so a deferral is no longer audited as
  a completed reap.

The guard is only safe with the retry: `unitFullyDown` is fail-closed, so without
something that revisits, a probe error would turn a rare corruption into a
permanent leak. Separate commits, but they land together.

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -2587,7 +2587,14 @@ func (m *Manager) reattachRecord(ctx context.Context, rec VMRecord, cleanupStale
 				sigkillPID(rec.PID, 500*time.Millisecond)
 				log.Info().Int("pid", rec.PID).Msg("killed orphan Firecracker process")
 			}
-			m.state.Delete(rec.ID)
+			// Delete-first, like markStale: freeing the slot while the record
+			// survives lets a later reattach of that record rebind a slot the
+			// pool has already handed to someone else. On failure keep both —
+			// the reconciler owns the retry.
+			if derr := m.state.Delete(rec.ID); derr != nil {
+				log.Error().Err(derr).Msg("failed to delete stale record; keeping its slot reserved")
+				return nil, false
+			}
 			// Free this record's namespace/slot directly instead of a broad
 			// re-sweep (which would also delete the warm pool's netns).
 			m.netMgr.CleanupVMOrNamespace(rec.ID, rec.Namespace)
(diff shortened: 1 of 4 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
Reviewed the latest head at the usual architecture, correctness, concurrency, crash-consistency, and reliability bar.

The deferred-release design now closes the two original hazards without introducing a stale-reap race: release is gated on terminal unit state, abandoned deletes are retried, pending episodes carry their originating drift marker, lifecycle ABA is rejected using incarnation plus monotonic generation, and retries are bounded by the reconcile pass context. The newer guards also preserve DB-outage semantics and validate/void pending entries before any control-plane failure transition.

I checked the startup stale-delete path as well: failed deletion parks a tracked Error instance with its slot retained, handing ownership to Drift 8 rather than abandoning the record or freeing networking early.

All review threads are resolved. CI and Terraform Checks are green. Approved.

PR #329, opened 2026-08-05
vmd: hold a launch while an unconfirmed stop for its unit is outstanding

meAmitPatil:
A launch whose unit-start job is delayed inside PID1 can time out waiting for
the API socket and enqueue its cleanup stop while the start is still queued or
executing. Under a congested PID1 both jobs run later, in order: the start
produces a live VM, a retry finds the socket up and adopts it — and the stale
stop then executes against the adopted VM, killing it mid-restore. A stop job
executes against whatever holds the unit name when it lands, not the
incarnation it was issued for.

The stop registry already records exactly this state (`recentUnitStops`,
cleared only when systemd confirms the stop done), but until now it only
widened the linger check. This adds the missing consumer: before starting a
unit, a launch that finds an unconfirmed stop on record waits — bounded by its
ctx — for the stop to settle, and fails rather than launch over one that never
does. The gate sits before anything is enqueued, so its failure needs no
cleanup and seeds no further stops. It keys on recorded stops only, not the
young-process settle window, which would otherwise stall every launch after a
vmd restart.

Replayed against the motivating failure, the retry waits a few seconds for the
stale stop to land and then launches a clean unit: the resume succeeds slower
instead of dying.

Tests cover all three gate outcomes; the fail-closed case was verified to fail
against a gate that launches anyway.

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -3701,6 +3701,25 @@ func (m *Manager) startFirecrackerViaSystemd(ctx context.Context, vmID, socketPa
 	replacingLive := !freshUnit && unitLingering(ctx, systemdUnitName(vmID))
 	lingerCheckMs := time.Since(tLinger).Milliseconds()
 
+	// An unconfirmed stop for this unit may still be sitting in PID1's job
+	// queue — issued for an incarnation that no longer exists, executing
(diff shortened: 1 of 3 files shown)


meAmitPatil:
@codex review

PR #336, opened 2026-08-07
billing: fence the team flag check behind the deduped interval scan

meAmitPatil:
The rollup scheduler's team-discovery queries (enqueueHour and nextTeamBackfillBatches) apply feature_enabled() before the DISTINCT collapses the interval rows, and the planner pushes the function into both interval-table scans — evaluating it once per interval row instead of once per team. As the interval tables grow, each tick burns progressively more CPU on redundant flag lookups; EXPLAIN ANALYZE shows the flag calls dominating the query by more than an order of magnitude over the scan itself.

Both queries now compute the deduplicated team set in a MATERIALIZED CTE first and flag-check that set. MATERIALIZED is load-bearing: it fences predicate pushdown, which is exactly the behavior the previous shape suffered from. The result set is unchanged (the flag is a per-team predicate, so filtering after dedup is equivalent), and the downstream CTE names and shapes are untouched.

Verified on a production-shaped dataset: the backfill discovery drops ~19x; the hourly enqueue has the same pathology at smaller scale and gets the same fix. feature_enabled is already marked STABLE, so no function changes are needed.


meAmitPatil:
@codex review

awille-ai (approved):
Reviewed the latest head for correctness, query semantics, planner behavior, scalability, and regressions.

The rewrite is semantically sound: `feature_enabled('billing_hourly_rollups', team_id)` is a per-team predicate, so applying it after `DISTINCT team_id` yields the same team set while avoiding redundant evaluations across interval rows. Using a `MATERIALIZED` CTE is appropriate here because the optimization depends specifically on preventing predicate pushdown back into the interval scans.

Both hourly enqueue and backfill discovery use the same fenced shape, downstream CTE contracts remain unchanged, there are no unresolved review threads, and CI plus Terraform Checks are green. I did not find a correctness or scalability regression. Approved.

PR #339, opened 2026-08-09
vm: split wait_socket into chain and firecracker phases

meAmitPatil:
Adds chain_ms / fc_socket_ms to the fc startup phases log via a pre-exec timestamp stamped by the launch script, so wait_socket variance can be attributed without guessing.


meAmitPatil:
Folded into #395 (the latency metrics PR): the pre-exec stamp now feeds both the phases log (chain_ms / fc_socket_ms) and the phase histogram (chain / fc_socket under vmd/launch), so the wait_socket split lands in dashboards directly.

PR #340, opened 2026-08-09
vm: spare cgroup pool for burst launches

meAmitPatil:
Creating a memory cgroup allocates per-CPU and per-NUMA-node bookkeeping under globally serialized kernel locks, so concurrent direct-spawn launches queue on it — the wider the burst and the bigger the host, the longer the line. This adds an optional pool of pre-created spare cgroups: the expensive creation happens in a background refiller, and a launch claims a spare with a cheap same-parent rename instead of a mkdir.

- Off by default; enabled with `VMD_CGROUP_SPARE_POOL=true` (size via `VMD_CGROUP_SPARE_TARGET`, default 256), and only active while direct spawn is armed.
- Strictly an optimization: any claim failure (empty pool, rename race, unwritable attribute) falls back to the existing inline creation path, so correctness never depends on the pool.
- `spare-*` joins `keeper` as a reserved name via one shared predicate: scans, the drain report, and the host rollback guard never read a spare as a VM; disarmed hosts reap leftover spares at startup.
- Spares survive vmd restarts and are re-adopted; excess and unparsable ones are removed.
- The launch phase log gains `cgroup_ms` + `spare_cgroup` so the claim-vs-create split is measurable per launch.

### DIFF
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -279,6 +279,20 @@ func runDrainCheck() int {
 	return 3
 }
 
+// spareCgroupTargetFromEnv resolves the spare cgroup pool size:
+// VMD_CGROUP_SPARE_POOL=true enables it at VMD_CGROUP_SPARE_TARGET
+// (default 256); anything else disables (0).
+func spareCgroupTargetFromEnv() int {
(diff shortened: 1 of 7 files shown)


meAmitPatil:
@codex review

PR #341, opened 2026-08-10
backup: escalate staged-task loss to the error channel

pavitrabhalla:
## Summary

First slice of backup observability: durability-loss events for staged tasks now reach the error channel, which the Sentry writer forwards. A staged task uploads from immutable snapshots, so a vanished or digest-divergent source there is corruption or an invariant break, not the designed fast-resume abandonment that unstaged tasks legitimately hit; today both log a warning visible only in one host's journald.

## Technical decisions

Level is chosen per task: error when staged, warn when unstaged, so the expected mutable-path abandonment pattern cannot flood Sentry while the never-expected staged losses page. Owner fields ride on every loss line.

## Testing

Both packages green on linux. Behavior change is log level only.


pavitrabhalla:
@codex review

meAmitPatil (approved):

PR #344, opened 2026-08-11
api: seed new sandbox rows with their requested shape

meAmitPatil:
A sandbox row is inserted with 1 vCPU / 1 MiB placeholders while the create is in flight, so anything reading the sandbox in 'starting' shows a bogus shape — and a create that fails before activation keeps those values forever. The real shape is already known at insert time (the template's vcpu/memory, or the default), so seed the row with it; ActivateSandbox still overwrites with the actuals vmd reports once the VM is up.


meAmitPatil:
@codex review

pavitrabhalla (approved):

PR #345, opened 2026-08-11
vm: optional bare per-VM cgroups

meAmitPatil:
Attaching controllers is the dominant cost of cgroup creation — the kernel allocates per-CPU and per-NUMA-node state under globally serialized locks, so concurrent direct-spawn launches queue on it, and the cost grows with host size. Behind a new flag, per-VM cgroups are created with no controllers enabled: creation drops to the cheap core path while the lifecycle machinery (cgroup.kill, cgroup.events, populated checks) is controller-independent and unchanged.

- `VMD_BARE_VM_CGROUPS=true` to enable; default off (today's behavior). The revert direction is the existing enable path.
- Containment is unaffected: the aggregate memory ceiling lives at sandboxes.slice, and the guest RAM bound is Firecracker's own. The trade is per-VM CPU/IO scheduling fairness (VMs share the service's domain per-thread) and per-VM memory accounting.
- Enabling on a host with running VMs converges immediately: disabling controllers is legal with live children, whose controller state tears down in place.
- The disable verifies `cgroup.subtree_control` is actually empty afterward and logs loudly if not — never a silent partial state.
- `memory.oom.group` is written when the file exists, skipped only on definitive absence (any other stat failure fails the create); whole-VM OOM semantics rely on the launch chain remaining a single process, now documented at the write site.
- The launch phase log gains `cgroup_us`, and the periodic gauge logs the scope's `nr_dying_descendants` so cgroup reclaim health is visible.
- Hardening found en route: a non-directory squatting on a VM's cgroup path is now rejected instead of tolerated by the EEXIST path.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

internal/vm/cgroup_linux.go
```diff
@@ -425,10 +436,16 @@ func adoptVmsScope(ctx context.Context) (*cgroupTree, error) {
...
+		// No controllers on the children (see ManagerConfig.BareVMCgroups
+		// for the why and the trade). The caller runs
+		// disableChildControllers to converge a previously-enabled host.
+		return t, nil
+	}
 	// root's subtree_control gives its children (vms/<id>/) their controller
 	// files. memory+pids are REQUIRED — per-VM memory.oom.group must exist or
 	// createVMCgroup fails — so a failure here is a fail-closed arm refusal.
-	subtree := filepath.Join(t.vms, "cgroup.subtree_control")
 	if err := os.WriteFile(subtree, []byte("+memory +pids"), 0o644); err != nil {
```
pavitrabhalla, line 488: **Blocking:** bare→non-bare transition can silently leave live VMs with per-process (not whole-group) OOM semantics.

When `BareVMCgroups` flips from true to false on a host with bare-created VMs still running, this write re-enables `+memory +pids` on the parent. The kernel will then expose `memory.oom.group` on those already-running child cgroups, but nothing re-writes `1` into it — that only happens inside `createVMCgroup` at *create* time, which doesn't re-run for VMs that already exist. Those specific VMs keep running with `memory.oom.group` unset (defaults to 0) until they're relaunched, with no log/error surfacing the gap.

The PR body documents the *opposite* direction (enabling bare with live children) as an intentional, legal convergence, but this direction isn't mentioned. Worth either: (a) confirming this is an accepted transitional gap and documenting it here the way the reverse direction is documented at `ArmDirectSpawn`'s `disableChildControllers` call, or (b) walking existing children and writing `memory.oom.group=1` on convergence back to non-bare.

internal/vm/cgroup_linux.go
```diff
@@ -141,6 +142,16 @@ func (m *Manager) ArmDirectSpawn(ctx context.Context) (bool, error) {
 		}
 	}
 	m.cgroups = tree // existing cgroup VMs are now manageable
+	if m.cfg.BareVMCgroups {
```
pavitrabhalla: **Non-blocking:** this bare-converge block runs whenever `BareVMCgroups` is true, regardless of `wantArm` — including manage-only mode (rollback/drain, `DirectSpawn=false`). If an operator sets `BareVMCgroups=true` without `DirectSpawn=true` (an unvalidated combination), this would disable controllers under pre-existing legacy cgroup-supervised VMs that rely on `memory.oom.group`, same risk as the comment on line 449 but via a different trigger. If that combination isn't meant to be supported, consider gating this block on `wantArm` or validating the combination at config load; if it's intentional, a short comment noting why manage-only mode should also converge would help future readers.

internal/vm/cgroup_linux_test.go
```diff
@@ -406,3 +406,55 @@ func TestPidsCeilingAdequate(t *testing.T) {
...
+		t.Fatalf("empty subtree must verify clean: %v", err)
+	}
+
+	// Plain files keep the written "-..." spec, standing in for a kernel that
+	// left controllers enabled: the verify must catch it and error.
+	if err := os.WriteFile(p, []byte("memory pids"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	if err := disableChildControllers(p); err == nil {
+		t.Fatal("controllers still visible after disable must be an error")
```
pavitrabhalla, line 458: **Non-blocking:** the new pure functions (`disableSpecFor`, `disableChildControllers` against a plain file) are well tested, but there's no test exercising the integration path — `adoptVmsScope(ctx, true)` or `ArmDirectSpawn`'s call into `disableChildControllers` against a real (or faked) cgroup tree. That matches this file's existing boundary (anything needing real cgroupfs/systemd isn't unit-tested here), so not a new gap introduced by this PR — just flagging it since this is the riskiest new code path and currently only exercised by manual/host testing.

pavitrabhalla (changes requested):
Solid change, consistent with the file's existing fail-closed/loud-never-silent conventions and well covered by unit tests for the new pure logic (`disableSpecFor`, `disableChildControllers`, the oom.group skip in `createVMCgroup`). One correctness gap worth resolving or explicitly documenting before merge (inline, marked blocking); the rest are non-blocking suggestions.

internal/vm/cgroup_linux.go
meAmitPatil: Fixed — went with option (b). Leaving bare mode now walks the existing children and writes `memory.oom.group=1` on any that have the file but don't claim it, logging the count, or a loud error if it can't. Verified end to end on a host: a VM created while bare read 0 after the revert before the fix, and converges to 1 with the fix while it keeps running. Unit test covers stamped / already-set / no-controller / keeper cases.

internal/vm/cgroup_linux.go
meAmitPatil: Good catch — that combination isn't supported. The converge block is now gated on `wantArm`, so a manage-only pass never changes controller mode on VMs that were created with controllers; the comment says why. Verified on a host: with the bare flag set but direct spawn off, the subtree keeps its controllers and existing VMs keep `oom.group=1`.

internal/vm/cgroup_linux_test.go
meAmitPatil: Agreed on the boundary. Exercised on a real host instead: both flip directions over live VMs, daemon restart, mid-create SIGKILL, keeper kill, pause-then-resume across modes, and repeated create/destroy cycles with `nr_dying_descendants` watched for growth.

pavitrabhalla (approved):

meAmitPatil:
@codex review

internal/vm/cgroup_linux.go
```diff
@@ -141,6 +142,34 @@ func (m *Manager) ArmDirectSpawn(ctx context.Context) (bool, error) {
...
+			// place; charges reparent to the service). Loud, never silent, but
+			// not fatal: with controllers still enabled the host is merely on
+			// the slower pre-bare behavior, which is correct.
+			if derr := disableChildControllers(filepath.Join(tree.vms, "cgroup.subtree_control")); derr != nil {
+				m.log.Error().Err(derr).Msg("bare vm cgroups requested but child controllers remain enabled — launches pay full cgroup-creation cost")
+			}
+		} else if !wasBare {
+			// No bare→controllers transition happened; the restore would be
+			// a per-live-VM scan for nothing on every ordinary startup.
+		} else if n, cerr := restoreChildOOMGroup(tree); cerr != nil {
```
meAmitPatil: Fixed: `adoptVmsScope` now captures whether the memory controller was absent *before* re-enabling it, and the restore runs only on that actual bare→controllers transition — never as unconditional startup work. An unreadable pre-state fails toward restoring (cheap, idempotent) rather than toward skipping a repair.

One correction to the cost model for the record: paused VMs hold no cgroup directories, so the scan was O(live VMs) — tens of file reads on a typical host, not fleet-sized. Gated anyway: startup work should track transitions, not defaults.

meAmitPatil:
@codex review

meAmitPatil:
Both findings fixed in 603575f9:

**P1 (manage-only mutation):** adoption and controller configuration are now separate functions. `adoptVmsScope` only resolves the scope and places the keeper — it never touches `cgroup.subtree_control`. `configureChildControllers` runs exclusively under `wantArm`, so a manage-only pass adopts the tree exactly as the previous life left it: no mode mutation during drain, no per-child kernel allocations on that startup path, and an enable failure now refuses arming (the fleet stays manageable — kill/liveness are controller-independent) instead of feeding `fatalIfManaging` into a crash loop. The `wasBare` capture moved inside the configure step, so the oom.group restore fires exactly on the real bare→controllers transition, which also closes the fourth consequence (re-enable without restore).

**P2 (swallowed read failures):** `restoreChildOOMGroup` now skips only `os.IsNotExist`; any other read failure is collected and returned like write failures, and the arming log reports it loudly.

Tests: `configureChildControllers` covered directly (bare-host enable → wasBare, enabled-host → no transition, stuck-disable errors, unwritable-subtree refuses arming) plus a read-failure propagation test; both fixes mutation-checked, full package suite green under the race detector.

internal/vm/cgroup_linux.go
meAmitPatil: Resolved differently per the review outcome: the restore stays synchronous inside the transition — enabling controllers across existing children is already synchronous and dominates the scan, so deferring only the sweep wouldn't change the startup profile. The real startup-path issue in this area was manage-only mode re-enabling controllers, which is now fixed: adoption no longer touches `subtree_control` at all, so non-arming startups do zero controller work of any kind. Resume-gating isn't added — the restored setting is redundant while each VM is a single process, so a brief unrestored window has no effect.

internal/vm/cgroup_linux.go
```diff
@@ -439,7 +486,112 @@ func adoptVmsScope(ctx context.Context) (*cgroupTree, error) {
...
+	restored := 0
+	var firstErr error
+	for _, e := range entries {
+		if !e.IsDir() || e.Name() == keeperSubdir {
+			continue
+		}
+		path := filepath.Join(t.vms, e.Name(), "memory.oom.group")
+		data, rerr := os.ReadFile(path)
+		if rerr != nil {
+			continue // no memory controller on this child: nothing to restore
```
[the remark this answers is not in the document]
meAmitPatil: Fixed: only `os.IsNotExist` skips; every other read failure is collected and returned the same way write failures are, so a transition over an unreadable child reports the error instead of a clean pass. Regression-tested (a non-ENOENT read failure must surface) and mutation-checked.

internal/vm/cgroup_linux.go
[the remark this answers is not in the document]
meAmitPatil: Superseded by the manage-only fix: adoption no longer touches `subtree_control`, so non-arming startups perform no controller work at all, and the restore still runs only inside a real bare→controllers arming transition — where the controller enable itself is already synchronous and dominates the scan. Resume-gating isn't added; the restored setting is redundant while each VM is a single process.

meAmitPatil:
@codex review

internal/vm/cgroup_linux.go
```diff
@@ -141,6 +142,42 @@ func (m *Manager) ArmDirectSpawn(ctx context.Context) (bool, error) {
...
+	// allocations during an incident-time startup, turn an enable failure
+	// into a crash loop of an otherwise manageable fleet, and leave
+	// bare-era children with oom.group at the kernel default.
+	if wantArm {
+		if m.cfg.BareVMCgroups {
+			// Disable failure is loud, never silent, but not fatal: with
+			// controllers still enabled the host is merely on the slower
+			// pre-bare behavior, which is correct. (Disabling is legal with
+			// live children — their controller state tears down in place.)
+			if _, derr := configureChildControllers(filepath.Join(tree.vms, "cgroup.subtree_control"), true); derr != nil {
```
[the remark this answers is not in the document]
meAmitPatil: Fixed: the controller-mode convergence now runs only after every arm precondition (launch-path validation and the rollback-guard check) has passed, immediately before arming — a failed arm no longer mutates live children in either direction. Manage-only passes still never reach it.

meAmitPatil:
@codex review

PR #346, opened 2026-08-11
infra: bound control-plane scaling to the connection pooler in both prod cells

meAmitPatil:
Each control-plane instance opens its own client connections, so max_instances * DB_MAX_CONNS is the fleet's demand on the connection pooler. At max 100 that product exceeded the pooler's client limit in both cells — a latent overcommit that only stayed harmless because the services never scaled that far.

Both cells now run min 10 / max 30 with 15 connections per instance, which keeps the product comfortably under the limit while leaving roughly double the headroom over the highest instance count either cell has been observed at. West's per-instance limit also moves 12 to 15 to match east.

### DIFF
--- a/infra/envs/production/us-east4/main.tf
+++ b/infra/envs/production/us-east4/main.tf
@@ -151,11 +151,11 @@ module "api" {
 
   cpu_limit    = "2"
   memory_limit = "1Gi"
-  # 6 matches the live service (raised from 2 during burst tuning). scaling is
-  # in the module's ignore_changes, but that path is ineffective for the v2
-  # resource, so declaring the live value keeps the plan a no-op.
-  min_instances     = 6
-  max_instances     = 100
+  # max_instances * DB_MAX_CONNS must stay under the pooler's client-connection
+  # limit; raising either means re-checking that product. Declared rather than
+  # left to drift: the module's ignore_changes is ineffective for the v2 resource.
+  min_instances     = 10
+  max_instances     = 30
   startup_cpu_boost = true
   cpu_idle          = true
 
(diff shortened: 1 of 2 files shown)


awille-ai (commented):
Generic review-only pass: the change is small and directionally sound, but I couldn't verify the actual pooler client limits from this repo. Since the safety argument depends on `max_instances * DB_MAX_CONNS` staying comfortably below those limits, I left one non-blocking comment asking to make that capacity assumption explicit/traceable so future changes can re-check it reliably.

infra/envs/production/us-east4/main.tf
```diff
@@ -151,11 +151,11 @@ module "api" {
...
   cpu_limit    = "2"
   memory_limit = "1Gi"
-  # 6 matches the live service (raised from 2 during burst tuning). scaling is
-  # in the module's ignore_changes, but that path is ineffective for the v2
-  # resource, so declaring the live value keeps the plan a no-op.
-  min_instances     = 6
-  max_instances     = 100
+  # max_instances * DB_MAX_CONNS must stay under the pooler's client-connection
+  # limit; raising either means re-checking that product. Declared rather than
+  # left to drift: the module's ignore_changes is ineffective for the v2 resource.
```
awille-ai, line 156: Non-blocking: the safety of `30 * 15 = 450` depends entirely on the pooler's actual client-connection limit, but that limit isn't captured here (and I couldn't find it elsewhere in the repo). Could we make the capacity assumption traceable — e.g. include the per-cell limit or a pointer to where it's defined/owned? That would make the comment's “must stay under” invariant actionable for the next person changing either knob rather than relying on tribal knowledge.

awille-ai (approved):
Reviewed the latest head for correctness, capacity math, rollout risk, and Terraform behavior. The change consistently caps both production cells at 30 instances, with 15 DB connections per instance, for a declared 450-connection fleet ceiling per cell; west also aligns its per-instance pool size with east. CI, Terraform Checks, and Terraform Plans are green. The remaining open comment about making the external pooler client limit traceable in-repo is useful for maintainability, but I don't see a correctness or rollout blocker in this change itself. Approved.

PR #350, opened 2026-08-11
vmd: widen the veth range so slots are not capped at 32k

meAmitPatil:
A host stops creating sandboxes once it has handed out 32,000 slot indexes, and every sandbox holds its index for its whole life — paused included — so a busy cell reaches that ceiling as a matter of course rather than as an anomaly.

The cap comes from addressing, not from hardware: host IPs take one address each out of a /16 and support 65,536 slots, but veth pairs take two each out of a /16 and run out at 32,768. Widening the veth range to a /15 lifts both sides to the same 65,536.

- Address each veth end by its flat offset into the wider range. Offsets below 65,536 land on exactly the addresses the old scheme produced, so no existing namespace moves and no running or paused VM needs rebuilding. Pairs stay /31-aligned across the boundary.
- Raise the slot ceiling to 65,000, under the 65,536 both ranges now allow.
- Route teardown carried its own hardcoded copy of the veth derivation, which wraps to a wrong address in the widened half and would leave a stale host route behind on every teardown up there; it now shares the one derivation.

Tests pin that all 32,768 existing slots keep their addresses, that the crossover and the top of the range land where they should, that host IPs stay inside the range the firewall matches on, and that each pair shares a /31.

Checked before choosing the range: nothing in either repo, no subnet in either GCP project, and no address or route on either production host uses it.

### DIFF
--- a/internal/network/manager.go
+++ b/internal/network/manager.go
@@ -71,9 +71,12 @@ type VMNetInfo struct {
 // ---------------------------------------------------------------------------
 
 // MaxSlots is the maximum number of concurrent VMs. Limited by the IP scheme:
-// hostIP uses 10.11.0.0/16 (one IP per VM), veth pairs use 10.12.0.0/16 (two IPs per VM).
-// This supports up to ~32K concurrent VMs per node — hardware (RAM/CPU) is the real limit.
-const MaxSlots = 32000
+// hostIP uses 10.11.0.0/16 (one IP per VM), veth pairs use 10.12.0.0/15 (two
+// IPs per VM), so both sides top out at 65,536 and this sits just under it.
+// Every sandbox holds a slot for its whole life, paused included, so this is a
+// ceiling on sandboxes per host, not on running VMs — hardware and kernel
+// scale (namespace and mount table growth) bite well before the arithmetic.
+const MaxSlots = 65000
 
 // ErrNoSlots is returned when no network slots are available.
 var ErrNoSlots = fmt.Errorf("no available network slots (max %d concurrent VMs)", MaxSlots)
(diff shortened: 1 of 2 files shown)


meAmitPatil:
@codex review

pavitrabhalla (approved):

awille-ai (approved):
Reviewed the latest head for correctness, backward compatibility, network isolation, teardown symmetry, scalability, and rollout safety.

The widened addressing scheme is sound: slots below 32,768 retain their exact existing veth addresses, the new half maps into 10.13.0.0/16 without pair crossover, host IPs remain inside the existing 10.11.0.0/16 firewall/NAT range, and each veth pair remains /31-aligned. I also checked the host firewall/routing assumptions: they key on the veth interface prefix and host-IP range rather than constraining veth addresses to 10.12.0.0/16, so the /15 expansion does not require firewall changes.

Teardown now shares the same vpeer derivation as setup, avoiding stale host routes for slots in the widened half. The tests pin backward compatibility, the crossover/top-of-range arithmetic, declared-range membership, /31 pairing, and cleanup derivation.

No unresolved review threads. CI and Terraform Checks are green. Approved.

awille-ai (approved):
Reviewed the latest head for correctness, backward compatibility, routing/firewall assumptions, scalability, and teardown behavior.

The /15 widening is implemented compatibly: slots below 32,768 retain their exact prior veth addresses, slot pairs remain /31-aligned across the 10.12/10.13 boundary, and the raised 65,000 ceiling remains below both the host-IP and veth address capacities. Cleanup now uses the same vpeer derivation as setup, avoiding stale-route teardown failures in the widened half.

I also searched for other hardcoded 10.12/16 assumptions in this repo and did not find another active routing/firewall dependency outside this network manager path. The tests cover old-range stability, the crossover, top-of-range arithmetic, declared subnet containment, /31 pairing, and teardown derivation.

No unresolved review threads. CI and Terraform Checks are green. Approved.

PR #353, opened 2026-08-11
vmd: scan slot reclaim without holding the allocator lock

pavitrabhalla:
## Summary
Intermittent sandbox-creation failures on us-west (usw2) persisted after #348/#350 landed. Root cause: `claimSlotIndex` held the single global `Manager.mu` across the reclaim scan added in #348 — two full directory reads (`/run/netns`, `/sys/class/net`, tens of thousands of entries at the ceiling) plus a loop over up to `MaxSlots` (65,000 as of #350) indexes. Every other slot claim on the host, including an on-demand `RestoreSnapshot` build racing the API's gRPC deadline, blocked on that lock for the scan's full duration — exactly when the host is under the most pressure (fresh index range spent).

## Technical decisions
- `reclaimUnusedSlots` (renamed from `reclaimUnusedSlotsLocked`) now manages its own locking: the directory reads run unlocked; only the cooldown check and the final in-memory merge into `freeSlots` take the lock, with a re-check of the cooldown after reacquiring it in case a concurrent caller scanned first.
- `claimSlotIndex` unlocks before calling the reclaim, relocks after, and re-reads state from the top of its loop — standard recheck-after-reacquire, no new invariants.
- Scoped to this lock-contention fix only. Two other threads from the investigation (tap-reset-on-recycle churn forcing full rebuilds, and the reconciler's chronic auto-fail budget exhaustion on usw2) are separate and need more evidence before touching — not addressed here.

## Testing
- `GOOS=linux go build ./... ` and `go vet` clean.
- Full `internal/network` suite green, including `-race -count=3` (Docker `golang:1.25`, since nftables deps are Linux-only).
- Updated the 4 test call sites that previously held `m.mu` themselves before calling the (now self-locking) reclaim function.

### DIFF
--- a/internal/network/manager.go
+++ b/internal/network/manager.go
@@ -939,7 +939,6 @@ func (m *Manager) claimTeardown(idx int, owner string) bool {
 // or the vmID for an on-demand SetupVM.
 func (m *Manager) claimSlotIndex(owner string) (int, error) {
 	m.mu.Lock()
-	defer m.mu.Unlock()
 
 	for {
 		var idx int
@@ -958,9 +957,21 @@ func (m *Manager) claimSlotIndex(owner string) (int, error) {
 				// Out of fresh range, which is not the same as out of slots:
 				// indexes are discarded (never returned) whenever a namespace
 				// outlives them, so the gaps below are the only inventory left.
-				if n := m.reclaimUnusedSlotsLocked(); n > 0 {
+				//
+				// reclaimUnusedSlots manages its own locking and must be
+				// called without m.mu held: it reads the whole netns and
+				// host-veth directories, which at the ceiling means tens of
+				// thousands of entries. Holding the single allocator lock
+				// across that scan would stall every other claim on the
+				// host — including an on-demand RestoreSnapshot setup
+				// racing its gRPC deadline — for the scan's full duration.
+				m.mu.Unlock()
+				n := m.reclaimUnusedSlots()
+				m.mu.Lock()
+				if n > 0 {
 					continue
 				}
+				m.mu.Unlock()
 				return 0, ErrNoSlots
 			}
 			idx = m.nextSlot
@@ -975,6 +986,7 @@ func (m *Manager) claimSlotIndex(owner string) (int, error) {
 		}
 
 		m.assignSlotLocked(idx, owner)
+		m.mu.Unlock()
 		return idx, nil
 	}
 }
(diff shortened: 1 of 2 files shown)


pavitrabhalla:
@codex review

pavitrabhalla:
## Self-review (parallel to @codex)

### Overview
This restructures `claimSlotIndex`/`reclaimUnusedSlots` so the expensive directory scans (`/run/netns`, `/sys/class/net`) run without holding `Manager.mu`, addressing real lock-contention on the slot-claim hot path. Root-cause diagnosis and the unlocked-scan approach are the right shape of fix. One correctness issue needs to be resolved before merge.

### Correctness — blocking

**`claimSlotIndex` can return a spurious `ErrNoSlots` even though slots were just reclaimed, under concurrent pressure at the ceiling.**

```go
if m.nextSlot > ceiling {
    m.mu.Unlock()
    n := m.reclaimUnusedSlots()
    m.mu.Lock()
    if n > 0 {
        continue
    }
    m.mu.Unlock()
    return 0, ErrNoSlots
}
```

`reclaimUnusedSlots()`'s own merge step re-checks the cooldown after reacquiring the lock and **defers to a concurrent scan that armed the cooldown first**, returning `0` without merging — that part is correct and intentional (per its own comment). But the caller here treats *its own* `n == 0` as "no slots available," when it actually only means "my scan lost the race to a fresher one." If two goroutines hit the ceiling concurrently, the loser's `n` is `0` even though the winner's scan just populated `m.freeSlots`.

Trace:
1. Goroutines A and B both see `nextSlot > ceiling`, both unlock, both scan concurrently (neither is on cooldown yet).
2. B finishes first, arms the cooldown, merges 500 slots into `m.freeSlots`, returns `500`. B's caller continues, pops a slot, succeeds.
3. A finishes its scan, relocks, sees the cooldown now armed (by B), skips its own merge, returns `0`.
4. A's caller sees `n == 0`, unlocks, and returns `ErrNoSlots` — **despite `m.freeSlots` holding ~499 reclaimed slots at that exact moment.**

This is a regression path this same PR could hit in the exact "burst of concurrent creates at the ceiling" scenario the PR is written to fix, turning a latency problem into spurious hard failures for a subset of concurrent callers.

**Suggested fix:** after reacquiring the lock, check actual current state instead of trusting this call's own `n`:
```go
m.mu.Unlock()
m.reclaimUnusedSlots()
m.mu.Lock()
if len(m.freeSlots) > 0 {
    continue
}
m.mu.Unlock()
return 0, ErrNoSlots
```
(the top of the loop already re-reads `len(m.freeSlots)` correctly on `continue`, so this is a minimal change — just stop keying the retry decision off `n`.)

### Test coverage
No test exercises concurrent callers racing into the ceiling/reclaim branch — `TestClaimSlotIndex_ReclaimsUnusedIndexesAtCeiling` is single-goroutine, and `TestClaimSlotIndex_ConcurrentDistinct` never reaches `MaxSlots`. Worth adding a concurrent-at-ceiling test alongside the fix so this can't silently regress.

### Minor / non-blocking
- `nsExists()` inside `claimSlotIndex`'s discard loop is still a per-index filesystem stat done under `m.mu` (pre-existing, not introduced here) — under a long run of discarded/phantom indexes this reintroduces a smaller version of the same class of problem this PR fixes. Worth a follow-up if it shows up in practice; not blocking this change.
- Concurrent callers that simultaneously hit the ceiling now each pay the full cost of the two directory reads redundantly (only one gets to merge). Correctness is preserved via the cooldown re-check, but under a real "creation storm" this could mean several full-host directory scans in flight at once. A `singleflight`-style dedup (only one goroutine actually scans; others wait on its result) would avoid the redundant I/O — worth considering as a follow-up, not blocking.

### Everything else
Rename, locking discipline in the merge section, and the `ReclaimUnusedSlots()`/test updates all look correct and consistent with the codebase's existing "Locked"-suffix convention. `-race` clean per the PR description makes sense — the bug above is a logic error, not a data race.

meAmitPatil (approved):

pavitrabhalla:
CI / Unit Tests work fine locally. They seem stuck due to GitHub actions. I'll override and merge this to put out the fire in production.

PR #354, opened 2026-08-11
vmd: fix pinned-slot reclaim hang; gate VMD deploys on CI

pavitrabhalla:
## Summary
Two follow-ups after #353 merged:

1. **Fixes a hang #353 introduced.** #353 changed the ceiling-branch retry in `claimSlotIndex` to recheck `freeSlots` after a reclaim scan instead of trusting the scan's own return value — correct for the general pool (a claim that finds `freeSlots` non-empty always pops one, draining it each iteration). But a `WithExactSlot` (pinned) claim never draws from `freeSlots` at all, so a reclaim triggered by its own ceiling failure fills `freeSlots` with entries nothing will ever consume, and the retry check kept finding it non-empty forever. `TestClaimSlotIndex_ExactSlotDoesNotAdvance` hung the whole `internal/network` package when run as part of the full suite — this is what was stuck in both the PR's CI and the post-merge run on main. A `-run`-scoped run of just the new test never reproduced it, which is how it slipped through. Fix: reclaim can never help a pinned claim (it only ever refills `freeSlots`), so skip straight to `ErrNoSlots` for the pinned case instead of scanning.
2. **Gates VMD deploys on CI.** `Deploy VMD` triggers independently off `push: branches: [main]` with no dependency on `CI` at all — a red or hung unit-tests run doesn't block staging/production rollout. Adds a `wait-for-ci` job that polls the same-SHA `CI` run, same pattern as Deploy API's `wait-for-migrations` (`needs:` can't reach a job in a different workflow). Manual `workflow_dispatch` skips the gate — operator owns that call.

## Technical decisions
- Also hardened the new concurrency regression test from #353 with explicit `select`/timeout bounds on its channel waits, so a real hang fails fast with a clear message instead of taking down CI if this class of bug recurs.
- `wait-for-ci` fails closed (deploy refused) on CI failure or a ~20 min timeout waiting for the run to appear/complete — mirrors `wait-for-migrations`'s budget and reasoning exactly.

## Testing
- `go build`/`go vet` clean (`internal/network`).
- Full `internal/network` suite green under `go test -v -short -race ./...` in a Linux container — including `TestClaimSlotIndex_ExactSlotDoesNotAdvance`, which previously hung indefinitely and now completes instantly.

### DIFF
--- a/.github/workflows/deploy-vmd.yml
+++ b/.github/workflows/deploy-vmd.yml
@@ -51,8 +51,48 @@ on:
       - '.github/workflows/scripts/deploy-vmd.py'
 
 jobs:
+  # Failing CI must not reach production control-plane hosts. CI is a
+  # separate workflow, so `needs:` can't reach it directly — poll the
+  # same-SHA run instead, same pattern as Deploy API's wait-for-migrations.
+  wait-for-ci:
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+    steps:
+      - name: Wait for same-SHA CI to succeed
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          if [ "${{ github.event_name }}" != "push" ]; then
+            echo "Manual dispatch — operator owns this decision; skipping gate."
+            exit 0
+          fi
+
+          # CI runs on every push to main (no path filter), so a run for this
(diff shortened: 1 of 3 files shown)


pavitrabhalla:
@codex review

meAmitPatil (approved):

PR #356, opened 2026-08-11
deploy: fix false host maintenance alerts on the JSON NONE sentinel

meAmitPatil:
The watcher queries the metadata server with `alt=json`, so the value comes back JSON-encoded: "nothing scheduled" arrives as the quoted string `"NONE"`, not the bare `NONE` the comparison expects. Every host therefore reported a pending maintenance notice of `"NONE"` on its first poll.

The same mismatch broke the clear path: when a real notice ends, the value becomes `"NONE"` rather than empty, so the watcher would have sent another notice message instead of the "cleared" one.

Strips one surrounding pair of quotes before the sentinel comparison, so both the JSON and bare forms read as "no notice". Adds tests that run the real script against a fake metadata endpoint and webhook, covering the quoted and bare sentinels, the notice → repeat → clear lifecycle, and the non-200 skip; they fail against the shipped script and pass with the fix.


meAmitPatil:
@codex review

pavitrabhalla (approved):

PR #359, opened 2026-08-12
deploy: move backup uploader journal off the root disk

pavitrabhalla:
## Summary

The backup uploader's journal (backup.db, a BoltDB file tracking queued pause backups) defaults to a path on the host's root disk instead of the local-SSD array that RUN_DIR and SNAPSHOT_DIR already use. When the root disk fills, journal writes fail and the uploader stalls, backing up the drain queue until queued generations get abandoned outright.

Adds `BACKUP_JOURNAL_PATH` as a deploy-time override, mirroring the existing `BACKUP_BUCKET` pattern. vmd already reads this via `envOrDefault`, so no Go changes are needed, just the deploy plumbing. Set to `/mnt/localssd/backup.db` for staging, use4, and usw2 (all three confirmed to have the same local-SSD XFS array).

## Technical decisions

- Kept the fix as a per-host env override rather than changing vmd's compiled default, since the local-SSD path isn't referenced anywhere else in the codebase and every deploy target needed verifying first (use4 and usw2 both confirmed).
- Did not place the journal inside RUN_DIR itself, since RUN_DIR is subject to disk-reclaim logic and mixing persistent app state into a reclaimed directory risks a different failure mode.

## Testing

- `python3 -m py_compile` on the deploy script
- YAML parse check on the workflow file
- Manually verified via SSH that both `superserve-vmd-usw2` and `superserve-vmd-use4` have `/mnt/localssd` as a separate XFS mount with ample headroom


pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

awille-ai (changes requested):
Re-reviewed the latest head. CI and Terraform Checks are green, and several earlier migration issues are fixed (RUN_DIR-derived source path, publish-after-migrate ordering, cross-filesystem temp+rename, mount precondition, workflow serialization). I’m still holding on the socket-activation race during journal migration.

The deploy stops only the vmd service while `superserve-vmd.socket` remains active. A connection during the migration window can therefore reactivate vmd against the old persisted env/journal path while the script is copying/renaming the database. The later `restart` does ensure the process eventually moves to the new path, but any writes made by the reactivated process after the destination copy was taken land in the old/renamed journal and are not merged into the destination. Those queued records are then lost on restart. Renaming the source to `.migrated` makes the writes recoverable for an operator, but it does not make the migration correct automatically.

Please quiesce socket activation (or otherwise prevent service activation) for the actual copy/source-retirement/env-switch critical section, then restart/reenable once the destination is authoritative. With that serialization in place, the `.migrated` retention can be simplified/cleaned up after validation as a follow-up if desired.

The root-disk cleanup concern by itself is not blocking for me; the correctness issue is the possibility of journal divergence while a socket-activated vmd is still able to write the source during migration.

pavitrabhalla:
@codex review

pavitrabhalla:
Fixed: stop superserve-vmd.socket alongside superserve-vmd.service right before the journal copy/rename/env-switch, so nothing can hold or reopen the old journal while it's mid-migration. Re-enabled by the existing restart/start block once vmd.env names the new path. bdf802f8

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

.github/workflows/scripts/deploy-vmd.py
```diff
@@ -670,6 +727,182 @@ def deploy(inst):
...
+                # the migration would silently skip a journal that does
+                # exist. Strip one matching pair of leading/trailing quotes
+                # from each before using it as a path.
+                BJ_Q1="'"
+                BJ_Q2='"'
+                OLD_BACKUP_JOURNAL_PATH=$(sudo grep '^BACKUP_JOURNAL_PATH=' /etc/sandbox/vmd.env 2>/dev/null | head -1 | cut -d= -f2-) || true
+                case "$OLD_BACKUP_JOURNAL_PATH" in
+                    "$BJ_Q2"*"$BJ_Q2") OLD_BACKUP_JOURNAL_PATH="${{OLD_BACKUP_JOURNAL_PATH#$BJ_Q2}}"; OLD_BACKUP_JOURNAL_PATH="${{OLD_BACKUP_JOURNAL_PATH%$BJ_Q2}}" ;;
+                    "$BJ_Q1"*"$BJ_Q1") OLD_BACKUP_JOURNAL_PATH="${{OLD_BACKUP_JOURNAL_PATH#$BJ_Q1}}"; OLD_BACKUP_JOURNAL_PATH="${{OLD_BACKUP_JOURNAL_PATH%$BJ_Q1}}" ;;
+                esac
```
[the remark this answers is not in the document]
pavitrabhalla: Scoping this to quote-handling, not a full EnvironmentFile parser: every write to BACKUP_JOURNAL_PATH/RUN_DIR in this script produces a plain `KEY=/path` line (no whitespace, no backslash escapes, no line continuation) — the script itself can never produce the syntax this comment describes. The quote-stripping already added covers the concrete case raised earlier (a hand-edited value using systemd-valid quoting). Backslash escapes and line continuation require an operator to hand-edit vmd.env with syntax this automation never writes; that's an operational risk a parser here can't meaningfully close, and implementing systemd's full env-file grammar (comments, line continuation, differing escape rules inside single vs double quotes) inline in bash is a substantially larger and riskier undertaking than the actual threat justifies. Not implementing further.

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

.github/workflows/scripts/deploy-vmd.py
```diff
@@ -670,6 +727,200 @@ def deploy(inst):
...
+                # mount vanished in between, vmd would silently open a
+                # root-backed journal at the same configured path, hiding
+                # the real one until the SSD is remounted.
+                BJ_ANCESTOR="$(dirname "$NEW_BACKUP_JOURNAL_PATH")"
+                while [ ! -d "$BJ_ANCESTOR" ] && [ "$BJ_ANCESTOR" != "/" ]; do
+                    BJ_ANCESTOR="$(dirname "$BJ_ANCESTOR")"
+                done
+                if [ "$(sudo stat -c %d "$BJ_ANCESTOR")" = "$(sudo stat -c %d /)" ]; then
+                    echo "ERROR: $BJ_ANCESTOR is on the root filesystem, not a separate mount; refusing to deploy" >&2
+                    exit 1
```
meAmitPatil, line 845: Please don't let the EXIT trap restart vmd after this guard fires. On an already-migrated host, a missing `/mnt/localssd` means the persisted path resolves to the underlying root directory, so recovery can create a second journal that is later hidden; keep both units stopped here, or condition recovery on a valid mount.

.github/workflows/scripts/deploy-vmd.py
```diff
@@ -293,6 +303,47 @@ def deploy(inst):
...
+                fi
+                # $BJ_ANCESTOR is confirmed on a real mount, so it's now safe
+                # to create the rest of the path down to the parent — vmd
+                # opens the journal directly with no mkdir of its own
+                # (cmd/vmd/main.go), so a nested path like
+                # /mnt/localssd/journals/backup.db needs this directory to
+                # exist before vmd (or the migration copy below) can open a
+                # file in it. A no-op when the parent already exists.
+                sudo install -d -m 0755 "$(dirname {q_backup_journal})"
```
meAmitPatil: `install -d -m 0755` changes the mode of an existing destination, including the `/mnt/localssd` mount root. Only create the directory when it is missing so deployments preserve the host's existing permissions.

meAmitPatil (changes requested):
Requesting changes for the journal recovery path: a failed late mount validation must not restart vmd onto the underlying root filesystem. The directory-mode issue should be corrected in the same pass.

pavitrabhalla:
Fixed both: the late mount recheck now disarms the recovery trap and stops the socket (not just skips restarting {service} myself) before exiting, so nothing -- neither my trap nor systemd's own socket activation -- can bring vmd back up against a path that just proved to be root-backed. And tightened the journal's parent directory from 0755 to 0700, matching the 0600 already used on the journal file itself; install -d re-chmods existing directories too, so this corrects any host that already created it at 0755. 425fb685

pavitrabhalla:
@codex review

meAmitPatil (approved):

awille-ai (changes requested):
Re-review of the latest head. The earlier socket-activation/journal migration race is addressed, but two current issues remain: (1) if the late mount revalidation fails, the EXIT recovery path can still restart vmd against the configured /mnt/localssd path while that mount is absent, creating a root-backed journal that will be hidden when the SSD returns; recovery on this guard must keep the service/socket down or otherwise require a valid mount. (2) `install -d -m 0700 "$(dirname BACKUP_JOURNAL_PATH)"` changes the mode of an existing parent; with the configured `/mnt/localssd/backup.db`, that chmods the mount root to 0700 on every deploy. Only apply the mode when creating a missing directory. CI and Terraform Checks are green, but these are still correctness/availability issues in the rollout path.

pavitrabhalla:
Both fixed: (1) the recovery trap now re-checks $BJ_ANCESTOR against / at fire time and stops (never starts) both units if it's still root-backed, instead of only special-casing the one exit site -- so a mount that vanishes after the recheck passes but before the later restart is still caught. (2) install -d -m 0700 now only runs when the directory doesn't already exist, so the common case (dirname of /mnt/localssd/backup.db is the mount root itself) is left untouched instead of getting re-chmod'd on every deploy. b122b2e8

pavitrabhalla:
@codex review

awille-ai (approved):

PR #360, opened 2026-08-12
vmd: release paused slots by namespace name and lower the reclaim band

meAmitPatil:
Releases a paused sandbox's network slot by namespace name so a VM missing from the net manager's device table can't strand its netns, veth, tap and mount entries after the record has already been cleared.

Also lowers the reclaim band to 8000/4000 namespaces with a 20-per-pass cap, so the controller engages below the point where deploys and ssh degrade and can actually reach its floor; applies min-warm-age under every pressure type, arms the reclaim cooldown on any evaluated pass, and drops an unused record field.


meAmitPatil:
@codex review

pavitrabhalla (approved):

meAmitPatil:
Checked the production state DB rather than reasoning from the comment: `paused_network_release_at` is present on 468 records (`omitempty` does not suppress a `time.Time`, so it serializes on every write) but every single occurrence is `0001-01-01T00:00:00Z` — no code path ever assigns it, so there is no timestamp to preserve.

meAmitPatil:
@codex review

PR #363, opened 2026-08-12
vmd: retry the launcher pin build, and alert when launches fall back

meAmitPatil:
Retries the pruned launcher-namespace build instead of leaving a host on the legacy launch path — where every Firecracker start walks the full host mount table — for the rest of the process's life when the boot-time build fails.

Also exports `vmd_launcher_ready` so that state can be alerted on rather than only logged, adds alert policies for it and for live-namespace growth on both production cells, and lowers the paused-network reclaim band to 6000/3000 with a 150 per-pass cap so the controller drains rather than tracking inflow.


meAmitPatil:
@codex review

pavitrabhalla (approved):
Reviewed the launcher retry/single-flight logic (traced the boot-build vs. sampler-tick race it guards against), the new launch-path-alerts Terraform module (structure matches backup-alerts/host-disk-alerts conventions), and the reclaim-tuning env changes across cells. No blockers.

Minor non-blocking notes: the new module doesn't expose both threshold and duration per condition the way the existing alert vars do (each condition here tunes one or the other, not both) — worth a look if these need per-cell tuning later. Also worth confirming the netns_accumulation alert default (8000) is meant to sit above the new reclaim threshold (6000) as a controller-failure backstop rather than a leftover from the old value.

meAmitPatil:
Good catch — confirmed and fixed, and it went deeper than the new policies.

The `OTelRecorder` really did attach only `service.name` and `environment`, while `BackupRecorder` stamps `host_id` itself and the collector deliberately skips OTLP senders (`# OTLP senders (control plane, vmd) already set their own host_id`). So `vmd_launcher_ready` and the pre-existing `vmd_network_*` gauges carried no `host_id` and the queries would have selected nothing.

Fixed by passing `HOST_ID` into the recorder and stamping it via a new `selfAttrs`, used by the launcher and paused-network gauges. It is deliberately separate from `attrs`: `RecordSandboxTransition`, `RecordVMDCall` and `RecordHostCapacity` already pass a `host_id` meaning "the host this operation addressed", and adding a second key there would collide with it.

Two further problems surfaced while checking this, both pre-existing:

- The alert matchers used the **instance name**, but this cell's vmd runs `HOST_ID=usw2` — its host-table row predates the name-matching convention and must not be changed. So the existing backup alerts for this cell have never been able to match either. Both now key on an explicit `metrics_host_id`.
- One host emits **two different `host_id` values**: the collector stamps the instance name on its hostmetrics pipeline, while vmd stamps its own `HOST_ID` on OTLP series. So the disk-utilization alerts are correct using the instance name and the vmd-metric alerts are not. Documented in both the local and the variable docs so the two are not "unified" later.

meAmitPatil:
Both fixed in `bafd10d2`.

**Tunable window.** Added `netns_total_duration` (default `1800s`), so that condition now exposes threshold and window like the existing alert vars. The launcher one still exposes only a duration — its gauge is binary 1/0, so a threshold there would have nothing to tune.

**The 8000 default.** You were right that it was a leftover: it was written when the reclaim band was 8000/4000 and I lowered the band to 6000/3000 in a later commit without revisiting it. It happens to be correct now, but by accident rather than intent — so I have made the intent explicit instead of leaving it to be rediscovered:

```
# COUPLED to vmd's reclaim band (VMD_PAUSED_NETWORK_NETNS_THRESHOLD /
# _HYSTERESIS): this must sit ABOVE the reclaim ceiling, so it fires only
# when the controller engaged and still could not hold the line ...
```

With the band at 6000/3000, an alert at 8000 means "the controller engaged and lost", and stays quiet through the normal sawtooth between floor and ceiling. Same note at both call sites, since that is where someone changing one of the two numbers would be looking.

meAmitPatil:
@codex review

meAmitPatil:
Real gap — fixed in the commit above.

`!launcherBuilt` only covered a pin that was never built. The second path is exactly as you describe: a pin that built at boot and later goes invalid clears `launcherReady` but leaves `launcherBuilt` set, so the rebuild was skipped and the process stayed on the legacy path until a restart — the same failure I was fixing, reached a different way. The alert runbook also claims rebuilds retry every five minutes, which would have been false in that case.

Now gated on `!m.launcherReady.Load()`, which is true only for a pin that is both valid and built this boot, so it covers both paths. Single-flight guard is unchanged, and it also makes the check a no-op while the boot build is still in flight.

Added a regression test that sets `launcherBuilt=true, launcherReady=false` — the exact state `revalidateLauncher` leaves behind on invalidation — and asserts a rebuild gets scheduled.

meAmitPatil:
@codex review

meAmitPatil:
Both removed in `878af46a` — thanks, these were my error.

Neither was intentional: they were untracked files already sitting in the working tree, and I swept them in with a `git add -A <dir>` while fixing the host_id labelling. They have no relationship to this change.

Removed with `git rm --cached`, so they are out of the branch but still present locally as the untracked artifacts they were before — nothing of anyone's work is destroyed. The branch diff is now clean of both.

meAmitPatil:
@codex review

awille-ai (commented):
Re-reviewed the latest head. The prior substantive findings are addressed: vmd's host-local metrics now carry HOST_ID, and launcher rebuilds retry whenever the current pin is not ready, including post-startup invalidation. Terraform Plans/Checks are green. I don't see a new correctness/security/rollout blocker in the current diff. CI is still in progress on this exact head, so I'm holding approval until it finishes.

meAmitPatil:
@codex review

PR #368, opened 2026-08-12
vmd: bound reconciler stale-cleanup loops to the pass deadline

pavitrabhalla:
## Summary
- The reconciler's stale-cleanup loops (markStale, stopVM, removeVMCgroup) run on contexts independent of the reconcile pass's own 25s runTimeout, so a large stale-VM-record backlog keeps grinding in real wall-clock time regardless of the budget.
- Meanwhile writeAudit, which does inherit the pass context, starts failing every call the instant it expires — that's why a big backlog produces a wall of "context deadline exceeded" errors instead of a clean, bounded pass.
- Adds a ctx.Err() check at the top of every stale-cleanup loop (11 sites), matching the pattern demotePausedCgroupRecords already used, so a pass stops taking on new destructive work once its budget is spent and defers the rest to the next tick.

## Technical decisions
- Checked ctx at the loop level (not by threading it into markStale itself) to keep the diff minimal and match the existing convention already in this file.
- Left MaxAutoFailPerHour untouched — this bounds pass duration, it doesn't change how much destructive work a pass is allowed to do.
- Left the disk-orphan reclaim path (detectDiskOrphans/reclaimDiskOrphans) alone — it already has its own DiskDeleteBudget bound and a ctx check before its destructive move.

## Testing
- Added TestReapDeadActiveVMs_ExpiredContextStopsBeforeAnyWork: five stale candidates against an already-expired context, asserts zero records are touched.
- Full internal/vm, internal/network, and whole-repo test suites pass (run under Linux, since this package needs NFPROTO_* constants unavailable on darwin — unrelated pre-existing platform gap).

### DIFF
--- a/internal/vm/reconciler.go
+++ b/internal/vm/reconciler.go
@@ -369,6 +369,9 @@ func (r *Reconciler) runOnce(ctx context.Context) {
 	// lock, re-checking that it's still recordless so an in-flight launch
 	// that just made the cgroup is never killed.
 	for id := range active {
+		if ctx.Err() != nil {
+			return
+		}
 		if supervisionOf[id] != SupervisionCgroup {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


pavitrabhalla:
@codex review

meAmitPatil (approved):

PR #369, opened 2026-08-13
observability: reference launch-path metrics by fully-qualified typed name

meAmitPatil:
The launch-path alert policies fail to apply: Managed Prometheus resolves a bare PromQL name ending in `_total` as a counter — which does not exist for our `vmd_network_netns_total` gauge — and rejects policies referencing metrics that have not emitted yet (`vmd_launcher_ready` ships in the same change as its policy).

Fully-qualified typed references (`prometheus.googleapis.com/<name>/gauge`) resolve both cases; re-running the failed applies after merge completes the policy creation on both cells.

### DIFF
--- a/infra/modules/observability/launch-path-alerts.tf
+++ b/infra/modules/observability/launch-path-alerts.tf
@@ -10,6 +10,12 @@
 # watch the same bounded host_id label as the backup metrics.
 
 locals {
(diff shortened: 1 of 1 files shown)


meAmitPatil:
@codex review

pavitrabhalla (approved):

PR #370, opened 2026-08-13
observability: use threshold conditions for launch-path alerts

meAmitPatil:
Second fix for the launch-path policy applies: the managed-Prometheus policy validator rejects the bare gauge name (a `_total` suffix resolves as a nonexistent counter), rejects the fully-qualified typed form, and refuses metrics that have not emitted yet — `vmd_launcher_ready` ships in the same change as its policy, so PromQL validation can never pass on first apply.

Classic `condition_threshold` with an explicit `metric.type` skips PromQL validation entirely, creates fine before first emission, and matches the module's existing CPU policies. Same thresholds, durations, and per-host scoping.

### DIFF
--- a/infra/modules/observability/launch-path-alerts.tf
+++ b/infra/modules/observability/launch-path-alerts.tf
@@ -10,13 +10,14 @@
 # watch the same bounded host_id label as the backup metrics.
 
 locals {
-  # Queries below use fully-qualified typed metric names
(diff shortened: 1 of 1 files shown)


meAmitPatil:
@codex review

pavitrabhalla (approved):

PR #374, opened 2026-08-13
boxd: retry writes on stale file handle, log boxd errors at the proxy

pavitrabhalla:
## Summary
Customer-reported 500s on `POST /files` on usw2, all `write: ... stale file handle` (ESTALE). Two independent fixes to unblock while the underlying cause (correlates with netns reclaim churn on usw2, not yet fully root-caused) is investigated separately:

- `cmd/boxd/main.go`: `handleFileUpload` now buffers the request body and retries the mkdir+open+write sequence up to 4 times with backoff when it hits ESTALE, instead of failing the request on the first hit. A fresh open() shortly after typically succeeds.
- `internal/proxy/files.go`: the `/files` reverse proxy now logs sandbox id, path, status, duration, and a body snippet whenever boxd returns >=400. boxd runs in-guest, so today these errors never reach any central log — the only way to see one right now is reconstructing it from the LB access log. This closes that gap for whatever shows up next.

## Technical decisions
- Retry only on `ESTALE`, not other errors — ENOSPC still maps to the existing 507 response, everything else still surfaces as a 500 as before.
- Body must be buffered before retrying since `io.Copy` from `r.Body` can't be replayed once partially drained.
- Proxy logging is scoped to `/files` only (where the actual issue is), not all boxd-port traffic.

## Testing
- `go build`/`go vet` clean on both packages.
- `go test ./internal/proxy/... ./cmd/boxd/...` passes, including existing `files_test.go` coverage.

### DIFF
--- a/cmd/boxd/main.go
+++ b/cmd/boxd/main.go
@@ -1391,47 +1391,70 @@ func writeStorageFull(w http.ResponseWriter, partialPath string) {
 	_, _ = w.Write([]byte(storageFullResponse))
 }
 
-func handleFileUpload(w http.ResponseWriter, r *http.Request, path string) {
+// staleHandleMaxAttempts bounds how many times we'll retry a write that
+// fails with ESTALE. The backing mount handing us a stale handle is
+// transient by nature (the guest's rootfs briefly loses and regains its
+// backing block device) — a fresh open() a few hundred milliseconds later
+// almost always gets a valid handle again, so we retry a handful of times
+// before giving up and surfacing a 500.
+const staleHandleMaxAttempts = 4
+
+func isStaleHandle(err error) bool {
+	return errors.Is(err, syscall.ESTALE)
+}
+
+// writeFileAttempt performs one mkdir+open+write+close cycle. It's split
+// out so handleFileUpload can retry the whole sequence on ESTALE — the
+// mkdir and the open can each independently observe a stale handle on the
+// same mount.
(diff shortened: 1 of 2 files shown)


meAmitPatil:
@codex review

pavitrabhalla:
@codex review

meAmitPatil (approved):

pavitrabhalla:
@codex review

awille-ai (approved):

awille-ai (commented):
Re-reviewed the latest head (`23358c65ae1e447f8d95ccb5c2e1096698a26cde`). The three prior findings are addressed in the current diff: uploads buffer only up to 32 MiB and preserve streaming behavior above that threshold; the retry loop rechecks the wall-clock budget after sleeping before starting another attempt; and error logging truncates the caller-controlled path to 200 bytes.

I also rechecked the retry/error semantics around the refactor: mkdir/open/write/close ESTALEs all remain detectable through wrapping, ENOSPC still maps to the existing 507 response, partial files are removed on write/close failure, and large uploads remain single-attempt/streaming rather than becoming memory-sized buffers. The proxy response-body peek is bounded and restores the consumed bytes before returning the response to the client.

Terraform Checks are green. CI is still running on this exact head, so I am leaving this as a non-blocking comment rather than approving until that check completes. I do not see a remaining code blocker on the current head.

pavitrabhalla:
@codex review

PR #375, opened 2026-08-13
boxd: bound aggregate ESTALE retry buffering, not just per-request

pavitrabhalla:
## Summary
- Follow-up to #374, which merged before this fix landed (Codex's P1 review comment arrived after the merge).
- The per-request 32 MiB buffer cap for ESTALE retries wasn't enough on its own: the proxy allows up to 200 concurrent connections to a single sandbox (`internal/proxy/proxy.go`), and the smallest supported sandbox has only 256 MiB of RAM total (`internal/api/handlers_template.go`). A handful of simultaneous near-cap uploads could buffer the guest's entire memory allocation between them, before boxd's own overhead or the guest OS are even counted.

## Technical decisions
- Added a shared, atomic 32 MiB budget across all in-flight retry-buffered uploads in the process, not per-request.
- A request that can't get a reservation falls back to the streamed, non-retrying path — the same degradation an oversized single body already gets, rather than blocking or introducing a new failure mode.

## Testing
- `go build`, `go vet`, `go test ./cmd/boxd/... -race` all pass.


pavitrabhalla:
@codex review

meAmitPatil (approved):

PR #377, opened 2026-08-13
backup: scope the backlog-age alert to the pause tier

pavitrabhalla:
## Summary
Prepares backfill enablement: the backfill sweep deliberately leaves a slow-draining best-effort backlog, and the priority-blind oldest-pending age gauge would hold the backlog-age alert firing for the whole enablement window. The age gauge now reports per priority tier and the alert reads only the pause tier.

## Technical decisions
- Same priority label shape the depth gauge already uses, so dashboards treat both uniformly; empty tiers report zero age.
- Age still measures from enqueue, not readiness: a nacked task in backoff is unmet backlog.
- Checkpoint and best-effort ages remain exported for dashboards; only alerting narrows to the pause tier, whose seconds-level freshness is the actual RPO signal.

## Testing
Journal test pins per-tier resolution (an hours-old best-effort task does not surface as the pause tier's age) and nack semantics; telemetry test pins the labeled points and negative-age clamping. Both suites green under race in the linux container.

### DIFF
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -819,9 +819,15 @@ func main() {
 				} else {
 					log.Warn().Err(err).Msg("backup metrics: journal pending read failed, dropping gauge from sample")
 				}
-				if oldest, ok, err := backupJournal.OldestEnqueuedAt(); err == nil {
-					if ok {
-						s.OldestPendingAge = time.Since(oldest)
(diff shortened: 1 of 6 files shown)


pavitrabhalla:
@codex review

meAmitPatil (approved):

PR #379, opened 2026-08-14
observability: use threshold conditions for backup and disk alerts

pavitrabhalla:
## Summary
Unblocks CD: the staging apply is failing on the backlog-age policy update because the managed-Prometheus policy validator rejects bare PromQL metric names when the series is absent or ambiguous at apply time. Converts all backup and host-disk alert policies to classic threshold conditions with explicit typed metric.type filters, the same escape the launch-path alerts already took.

## Technical decisions
- Thresholds, windows, severities, and documentation preserved verbatim; only the condition mechanism changes.
- The pause-tier backlog scoping becomes a metric.labels.priority filter; the failure-rate threshold divides to per-second for ALIGN_RATE; the pause-hook p99 uses ALIGN_PERCENTILE_99 on the distribution.
- Backup and host-disk resources keep their own host filter suffixes.

## Testing
terraform fmt clean; CI runs the real plans against all four roots. The staging apply exercising the previously failing policy update is the acceptance check.

### DIFF
--- a/infra/modules/observability/backup-alerts.tf
+++ b/infra/modules/observability/backup-alerts.tf
@@ -1,15 +1,20 @@
 # Backup pipeline alerting. The vmd daemon exports backup metrics through
 # the host-local OTel collector into Managed Prometheus; these policies
 # watch one cell's host via the bounded host_id label (never sandbox or
-# customer identifiers). PromQL conditions fire while the query returns a
-# series, so each query folds its threshold into the expression.
+# customer identifiers). Classic threshold conditions with explicit
+# metric.type, NOT PromQL: the managed-Prometheus policy validator
+# rejects bare metric names whenever the series is absent or ambiguous
+# at apply time (the same trap the launch-path alerts hit), so every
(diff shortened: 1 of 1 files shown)


meAmitPatil (approved):

awille-ai (approved):
Reviewed with the linear-ticket-implementation generic review workflow. No Critical or High findings found. I inspected the full one-file Terraform diff, surrounding alert policy configuration, current PR review state, CI/Terraform checks, and the generated staging plan. The typed metric filters preserve the host/result/priority/mountpoint scoping; ALIGN_RATE correctly converts the counter to per-second before comparing against the per-hour threshold divided by 3600; ALIGN_PERCENTILE_99 is valid for distribution-valued metrics; and the gauge conditions use matching min/max aligners and thresholds. The staging plan shows only the intended six alert-policy updates plus existing dashboard updates, with no creates/replacements/destroys. CI, Terraform Checks, and Terraform Plans all passed. The PR was already merged before this review.

PR #382, opened 2026-08-14
api: finalize pause synchronously before responding

pavitrabhalla:
## Summary

`PauseSandbox` returned success once vmd's `PauseVM` RPC completed, leaving the DB status flip (`pausing` -> `paused`) as fire-and-forget bookkeeping in a background goroutine. A resume issued right after a successful pause could race that goroutine, see the sandbox still in `pausing`, and 409 against an operation the client was just told succeeded. Reproduced consistently (~5% of resumes) under concurrent burst pause/resume against staging.

## Technical decisions

- Await `finalizePause` before responding instead of backgrounding it. Closes the race, at the cost of adding the DB write's latency to every pause call, not just the ones that would have raced.
- A hard finalize failure now returns 500 instead of a false-positive 204 — the sandbox is genuinely stuck in `pausing` with the VM already paused underneath it, and reporting success would just relocate the race instead of closing it.
- The soft-delete-mid-pause edge case (`pgx.ErrNoRows`, a rare race with `DeleteSandbox`) still reports success, matching prior behavior: the VM is already stopped and there's nothing to reconcile for a sandbox that no longer exists.
- Also fixes a latent test-mock bug this change exposed: the mock's SQL router checked for `'pausing'` before `upserted AS`, so `FinalizePause`'s own WHERE clause (which matches `status IN ('pausing', 'resuming')`) was misrouted to the `BeginPause` mock row. It never surfaced before because the panic happened inside the async goroutine, silently swallowed by its `recover()`.

## Testing

- `go test ./internal/api/...` — full package passes, including the previously-masked mock bug now fixed.
- Reproduced the race pre-fix with a burst script against staging (20 concurrent sandboxes: create, pause all, resume all) — consistent 1/20 resume failures across two runs. Plan to re-run the same script post-deploy to confirm 0 failures, and compare `sandbox_transition_duration_seconds{operation="pause"}` before/after to quantify the added latency.

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -2619,39 +2619,46 @@ func (h *Handlers) PauseSandbox(c *gin.Context) {
 		Msg("VMD pause complete")
 
 	// Past this point the snapshot already exists on disk — the pause has
-	// physically happened, so the bookkeeping (snapshot row upsert + status
-	// flip pausing → paused in a single CTE) is fire-and-forget. BeginPause's
-	// gate write owns the row and every other transition is status-gated, so
-	// a racing resume sees 'pausing' and 409s until this lands. Detached from
-	// cancellation so a client disconnect cannot orphan the bookkeeping;
-	// trace/span context preserved. The upsert replaces the old "insert a new
-	// row + delete the previous" flow, so there's no explicit prev-snapshot
-	// cleanup to schedule here.
-	finalizeCtx := context.WithoutCancel(c.Request.Context())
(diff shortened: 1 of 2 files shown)


pavitrabhalla:
Closing in favor of #384: the fire-and-forget pause design is deliberate (keeps pause off the hot path for a snappy client experience), so making pause synchronous here was the wrong lever. #384 instead lets resume poll through a racing 'pausing' row for a bounded settle window, reusing the same mechanism already used for the equivalent create/resume race. Pause's response time is untouched.

PR #383, opened 2026-08-14
backup: try reflink staging before applying the inline copy guards

meAmitPatil:
StagePending now attempts a FICLONE clone per file before the deadline and packed-size guards, so on reflink-capable filesystems staging is O(metadata) at any size and no pause is demoted to the marker-only path. The guards now apply only to the byte-copy fallback, unchanged.


meAmitPatil:
@codex review

PR #384, opened 2026-08-14
api: let resume settle through a racing pause finalize instead of 409ing it

pavitrabhalla:
## Summary

`PauseSandbox` responds success as soon as vmd's `PauseVM` RPC completes; the DB status flip (`pausing` -> `paused`) happens afterward as fire-and-forget bookkeeping, deliberately, so pause stays fast even on an idle host that could otherwise wait on the write. The tradeoff: an immediate follow-up resume can read the sandbox row before that write lands, see it still `pausing`, and 409 against an operation the client was just told succeeded. Reproduced consistently (~5% of resumes) under concurrent burst pause/resume against staging.

This closes the race without touching pause's latency at all: `ResumeSandbox` now polls briefly through a `pausing` row instead of failing on the first read, giving the racing bookkeeping write a chance to land.

## Technical decisions

- Reused the settle-window mechanism that already exists for the equivalent create/resume race (`activateSettleWindow`/`activateSettlePoll`, used by `loadActiveOrResumeSandbox` to poll through a `starting`/`resuming` row) instead of inventing a second one.
- The fix lives entirely on the resume side. Pause's response time is unchanged — resume already does a real VMD round trip, so absorbing a short bounded poll for the rare race is cheap relative to its existing latency.
- Any other non-`paused` status (`active`, `failed`, etc.) still 409s on the first read, same as before. A sandbox genuinely stuck in `pausing` past the settle window (finalize lost or truly stalled, not just racing) still 409s once the window expires.

## Testing

- `go test ./internal/api/...` — full package passes.
- Added `TestResumeSandbox_SettlesPausingToPaused` (row flips from `pausing` to `paused` mid-poll, resume succeeds) and `TestResumeSandbox_PersistentPausing_409` (row stays `pausing` past a shrunk window, resume still 409s).
- Reproduced the race pre-fix with a burst script against staging (20 concurrent sandboxes: create, pause all, resume all) — consistent 1/20 resume failures across two runs. Plan to re-run the same script post-deploy to confirm 0 failures.

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -139,7 +139,7 @@ type Handlers struct {
 	Encryptor secrets.Encryptor // KMS envelope used by /secrets endpoints; nil disables them
 	Signer    *SecretsSigner    // signs sandbox JWTs and serves the JWKS; nil disables both
 	Stripe    StripeBillingClient
-	Now       func() time.Time  // when set, returns the current UTC time for testable handlers
+	Now       func() time.Time // when set, returns the current UTC time for testable handlers
 
 	// asyncMu/asyncCond/asyncCount track fire-and-forget bookkeeping goroutines
 	// (ActivateSandbox, FinalizePause) so tests can wait for quiescence;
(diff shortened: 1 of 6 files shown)


pavitrabhalla:
@codex review

pavitrabhalla:
Backed off ResumeSandbox's pausing-settle poll: it was re-reading GetSandbox on a fixed 50ms interval for the full 6s window, up to ~120 reads for a sandbox genuinely stuck in 'pausing'. Now uses a separate 2s window (finalize-pause is a local write, not a guest round trip) with the poll interval doubling 50ms -> 500ms, bounding worst case to about a dozen reads. Added a log line and a dedicated sandbox_resume_settle_wait_total/duration/reads metric set, emitted only when the loop waits past the first read, so settle waits are visible separately from overall resume latency. Strengthened TestResumeSandbox_PersistentPausing_409 to assert the bounded read count.

pavitrabhalla:
@codex review

awille-ai (approved):
Reviewed current head c588b44b2e3efa6398005243948dd7ea3befff03 for correctness, security/resilience, test quality, performance/resource use, observability, data/state-transition correctness, and sandbox lifecycle latency. CI and Terraform checks are green. The remaining P2 observations on request-cancellation responsiveness and settle-metric classification are non-blocking; no Critical or High findings found.

pavitrabhalla:
Addressed both findings. The settle wait now selects on the request context: a client that disconnects or times out mid-backoff releases the handler at cancellation instead of finishing the interval, and the follow-up GetSandbox on the canceled context (previously logged as an internal DB error) is skipped entirely. Result classification is now exact: only pausing->paused records "settled"; a row that left 'pausing' for any other state (failed pause reverted to active, delete claimed it) records "diverged", and a mid-wait cancellation records "canceled". Added regression tests for both: a reverted pause must emit diverged and still 409, and a canceled request must stop at exactly one read.

pavitrabhalla:
@codex review

PR #386, opened 2026-08-14
deploy: mask vmd through the config window so socket activation cannot double-boot it

meAmitPatil:
The deploy stops vmd and rewrites its config across several steps while the socket deliberately keeps listening (connections backlog instead of refusing). A connection arriving in that window socket-activates vmd against half-written config; the script's final restart then executes that instance mid-flight — killing whatever it was serving, including in-flight RPCs, and doubling the restart churn on any deploy that receives traffic.

Mask the service (--runtime) before the stop and unmask just before the final restart: with the mask held, no instant exists in which the service is both stopped and startable, so the backlogged connections wait for the deploy's one real boot and are served with the finished config. Both recovery traps unmask before acting, a reboot clears a stranded mask on its own, and the next deploy's unmask self-heals a run that died trap-less mid-window.

### DIFF
--- a/.github/workflows/scripts/deploy-vmd.py
+++ b/.github/workflows/scripts/deploy-vmd.py
@@ -998,10 +998,9 @@ def main() -> int:
             fi
             # The config window is over: lift the mask and perform the one
             # and only boot of this deploy. The mask above makes mid-window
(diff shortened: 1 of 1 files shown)


meAmitPatil:
@codex review

PR #390, opened 2026-08-14
vmd: refuse to start without an explicit HOST_ID

meAmitPatil:
vmd's `HOST_ID` silently fell back to the literal `"default"` when unset. That value is the daemon's identity for heartbeats, control-plane routing, and reconciler scoping — two daemons sharing an identity would each reconcile the other's sandboxes against their own local state and fail them. A missing value now stops the process at startup with a clear error instead of adopting a shared placeholder; `"default"` itself is rejected as the historical fallback value.

No behavior change for any correctly configured host: the deploy path already writes `HOST_ID` into the env before the daemon restarts, so the new check only fires on a misconfigured machine — which is the point.

### DIFF
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -99,7 +99,7 @@ func loadConfig() (Config, error) {
 		HostInterface:           envOrDefault("HOST_INTERFACE", "eth0"),
 		TemplateBuilderBin:      envOrDefault("TEMPLATE_BUILDER_BIN", "/usr/local/bin/template-builder"),
 		BoxdBinaryPath:          envOrDefault("BOXD_BINARY_PATH", "/usr/local/bin/boxd"),
-		HostID:                  envOrDefault("HOST_ID", "default"),
+		HostID:                  requireEnv("HOST_ID"),
 		DatabaseURL:             os.Getenv("DATABASE_URL"),
 		ControlPlaneURL:         os.Getenv("CONTROL_PLANE_URL"),
 		SecretsProxySocket:      os.Getenv("SECRETSPROXY_SOCKET"),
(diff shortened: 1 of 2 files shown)


meAmitPatil:
@codex review

meAmitPatil:
Good catch — confirmed against `infra/envs/production/us-east4/main.tf:49-65`: that host's identity is literally `default` and its row keys on it, so the ban would have refused startup on the next deploy there. Updated: the guard now rejects only an absent `HOST_ID`. Banning the literal `default` is deferred until that identity is migrated as a deliberate operation; meanwhile the duplicate-identity case is covered at the control plane by the upcoming registration guard (a second host claiming an in-use identity is rejected at heartbeat), which protects `default` like any other value.

meAmitPatil:
@codex review

awille-ai (approved):
Reviewed current head `81e7cf9536` using the implementation-review workflow.

Review scope:
- `cmd/vmd/main.go`: HOST_ID config loading and startup validation
- `cmd/vmd/main_test.go`: coverage for missing, legacy `default`, and normal host identities
- deployment/production context confirming `default` remains a required legacy identity in us-east4

Security/resilience: PASS — missing HOST_ID now fails closed at startup, while the deployed legacy `default` identity remains compatible.
Test quality: PASS — the tests directly cover the new startup contract and the legacy compatibility case.
Correctness/maintainability: PASS — the implementation matches the current repository/deployment invariant.

Non-blocking note: the PR description is stale: it still says the literal `"default"` is rejected, while the current implementation intentionally allows it. Updating the description would keep the review context accurate.

Validation: inspected the full PR patch and surrounding config/deploy code. GitHub reports no combined status checks for the current head via the connector.

awille-ai (approved):

awille-ai (approved):

PR #393, opened 2026-08-14
api: self-register hosts from heartbeats; add hostctl and status endpoints

meAmitPatil:
Host rows are currently created by hand-written SQL: the heartbeat 404s for an unknown host id, `CreateHost` has no production caller, and the `draining` status exists only in the schema's CHECK constraint. This wires the missing lifecycle:

- **Self-registration**: a heartbeat for an unknown host id that carries a complete self-description (`vmd_addr`, `proxy_addr`, `region`, capacities) creates the row in a new `provisioning` status — invisible to the scheduler and the unhealthy-host detector, so a freshly deployed host heartbeats freely without ever taking traffic. Registration stamps `last_heartbeat_at` so capability attestation binds from the first beat. Heartbeats without a description keep today's 404.
- **Duplicate-identity guard**: a heartbeat claiming an existing id from a different `vmd_addr` is rejected with 409 while the row's holder is still live; if the holder has been silent past the unhealthy threshold, the identity is released to the new address (re-provisioned host). Two live daemons can never share an identity — each would reconcile the other's sandboxes against its own local state.
- **Operator endpoints**: `GET /internal/hosts` (every host with liveness + running/paused counts) and `POST /internal/hosts/{id}/status` for `active`/`draining`; machine-managed states are not settable. Status changes invalidate the scheduler's host cache.
- **`cmd/hostctl`**: stdlib-only operator CLI (`list`, `activate`, `drain`) that talks to the control plane's internal API, replacing hand-run SQL during host bring-up.

No behavior change for existing hosts: current heartbeats carry no self-description, so they follow the exact prior path; the migration only widens the status CHECK. Covered by integration tests for register/conflict/reclaim/status/list plus the updated handler unit tests.


meAmitPatil:
@codex review

meAmitPatil:
Confirmed against `internal/hostreg/registry.go` — `ClientFor` returns the cached client without re-reading the row, and self-eviction only fires on a dead transport, so a reclaimed address would leave lifecycle RPCs flowing to the old machine (silently, if the old daemon still answers). Fixed: `HostRegistry` gains `Invalidate`, and the heartbeat handler evicts the cached client after a commit that rewrote the address. Unit test pins both paths: stale reclaim evicts exactly once; a rejected live-conflict claim evicts nothing.

meAmitPatil:
@codex review

meAmitPatil:
Fixed. An address-mismatch claim is now treated as what it is — a re-registration — so it requires the complete self-description. A partial one gets 400 with **no writes at all**: not just skipping the reclaim, but also skipping the heartbeat update, because bumping `last_heartbeat_at` on a row whose `vmd_addr` still points at the silent old machine would mark the wrong address live. Conforming vmds are unaffected (the sender already refuses to emit partial descriptions). Test extended: partial claim → 400, no eviction, no row writes.

meAmitPatil:
@codex review

meAmitPatil:
Both P1s addressed:

**Provisioning host vs the default fallback** — confirmed and fixed in `43bdf5eb`. The empty-candidate fallback in `SelectHost` now reads the default host's row first: a genuinely missing row (unpopulated table, the bootstrap case) still falls back, an `active` row still falls back (capability-filtered case, unchanged behavior — the create-time capability gate enforces), and any non-active status (`provisioning`/`draining`/`unhealthy`) refuses with a clear error instead of routing. Test covers all five cases. The lookup only runs on the empty-candidate edge, so the placement hot path is untouched.

**vmd never sends the description** — correct, and intentional: the sender half is **#394**, which wires `vmd_addr`/`proxy_addr`/`region`/detected capacity through `HeartbeatConfig` behind opt-in env vars. It's split from this PR because the sender must not deploy ahead of this handler (older control planes bind heartbeat bodies strictly), and the env opt-in makes merge order free. Until a host sets those vars, the 404-for-unknown-hosts behavior here is exactly today's, by design.

Also renamed the migration to `20260817000002` — main merged `20260816000001`/`20260817000001` after this branch cut, so the old `20260815...` stamp sorted before applied migrations and failed the Migration Order check.

meAmitPatil:
@codex review

meAmitPatil:
Fixed in `f5e45ece`. The fallback status now rides the existing SWR cache instead of being queried per create: `fillEntry` resolves the default host's row **only when the candidate list comes back empty, once per cache fill** (blocking load and background refresh both), and `SelectHost` decides from the cached entry with zero DB I/O. Freshness follows the same machinery the candidate sets already use — TTL + stale-grace, and `Invalidate()` on every operator status change and heartbeat recovery — so an activation is visible immediately, same as before. Test pins it: five creates in bootstrap mode perform exactly one `GetHost`.

meAmitPatil:
@codex review

meAmitPatil:
**Registry invalidation across replicas** — fixed in `2122ce3a`. Cross-replica busses don't exist in this stack by design, so the registry now converges through the store that *is* shared: cached clients carry their dialed address and re-verify it against the host row at most once per 30s interval (matching the heartbeat cadence), stale-while-revalidate so the request path never blocks — a recheck is one `GetHost` per host per interval per replica, ~2 QPS cell-wide at current scale. Address changed → evict, next call re-dials; unchanged → refresh the clock; lookup failure → keep the client (RPC failure + dead-peer eviction stay the backstop). The heartbeat-serving replica still evicts immediately; every other replica converges within one interval instead of never. Race-detector-clean tests cover both directions.

**Drain fencing across replicas** — intentionally not addressed with new machinery in this PR, and here's the reasoning I'd ask reviewers to weigh: unlike the registry case (unbounded staleness — a cached client would *never* converge), drain staleness is already bounded by the existing TTL+grace windows, and behind them the create path's capability admission re-checks `status = 'active'` authoritatively per its own short cache. Worst case, a just-drained host can accept placements for a few tens of seconds — for an operator-driven graceful drain (the host is healthy, just being emptied), a sub-minute convergence tail is the accepted semantics, and shortening it would mean an uncached status read per create, taxing the latency-critical path to speed up a rare ops action. The authoritative zero-cost fence is already designed in the capacity-admission work: the placement reservation is an atomic conditional `UPDATE ... WHERE status = 'active'` fused into the create's existing round trip, which makes drain enforcement per-create and instant with no added I/O. Happy to escalate if we want a hard fence before that lands.

meAmitPatil:
@codex review

meAmitPatil:
Escalation accepted — fixed in `66fe0dbe`. The recheck is now **blocking, not stale-while-revalidate**: a cached client past its 30s verification window is never dispatched through. `ClientFor` verifies the row first (singleflight — concurrent callers share one ~ms read, so the cost stays one row read per host per interval, never per operation), and on an address change it dials the **new** address and returns that client, so not even the first post-expiry call can reach the old machine. If the new address won't dial, the call **fails closed** rather than falling back to the old-address client — a loud error beats executing on the machine that lost the identity. The one deliberate softness: if the row *read* itself fails, the cached client is returned (hazard requires a reclaim and a DB failure in the same window; refusing to dispatch on any read blip trades that for routine unavailability). Race-clean tests pin all three behaviors: re-dial-before-dispatch, no re-dial when unchanged, fail-closed on undialable new address.

meAmitPatil:
@codex review

meAmitPatil:
Fixed. Activation now requires a heartbeat fresher than the unhealthy threshold, enforced **inside the UPDATE's predicate** so there is no check-then-act window — a provisioning host that died before the operator got to it returns 409 (`host has no live heartbeat; refusing to activate a host the fleet cannot see`) and its status is untouched. Draining stays unconditional (taking a dead host out of rotation is legitimate). Zero-rows is disambiguated from not-found so the operator sees the right error. Integration test covers stale-activate-409 / stale-drain-200 / fresh-reactivate-200.

While validating, the new test surfaced a test-isolation bug worth noting: it was the first host test to leave an **active** host row behind, and other tests' schedulers could then randomly place sandboxes on that undialable host (reproduced as an intermittent billing-test failure). All host tests now delete their rows via cleanup; five consecutive full-suite runs green.

meAmitPatil:
@codex review

meAmitPatil:
Fixed in the latest commit. `Invalidate` now bumps a per-host generation alongside the delete, and a verification snapshots the generation before reading the row: if it moved while the dial was in flight — a reclaim landing mid-verification — the dialed client is **discarded and the row re-read** (bounded retry, then a loud error on pathological churn), instead of the store resurrecting the just-invalidated address as freshly verified. Same generation idiom the scheduler's candidate cache already uses for exactly this write-after-invalidate hazard. Deterministic race test: dial to the intermediate address blocks while the invalidation lands → dial sequence must be old → raced → newest, the raced result never cached, and the settled entry serves without re-dialing. Race-detector-clean at -count=5.

meAmitPatil:
@codex review

meAmitPatil:
Fixed in `4f716505`. The error path now honors the invalidation state instead of bypassing it: on a failed row read, the cached client is returned **only while the snapshotted entry is demonstrably not invalidated** — still present in the map at the same address. Entry gone (invalidated for cause) → retry the read once, then **fail closed** with the read error; entry replaced by another path → return that fresh client. The availability softness survives only in the case it was meant for: a read blip with no invalidation in flight. Also addressed the canceled-leader hazard from the finding: the verification read runs on a detached context (own 5s timeout), so one caller's cancellation can't decide the singleflight result for everyone behind it. Two new tests pin the split: read-blip-without-invalidation serves cached; invalidated-plus-unreadable-row errors with nothing dialed. Race-clean at -count=5.

meAmitPatil:
@codex review

meAmitPatil:
Three findings addressed across `4f716505` and `1ed55216`:

**Fail closed when the row is unreadable after an invalidation** (`4f716505`) — the read-error softness no longer bypasses the invalidation state: the cached client is returned only while an entry actually exists in the map; entry gone (invalidated for cause) → retry once, then fail closed with the read error. The verification read also runs on a detached context with its own 5s timeout, so one caller's cancellation can't decide the singleflight result for the followers sharing it. Tests pin the split: read-blip-without-invalidation serves cached; invalidated-plus-unreadable errors with nothing dialed.

**Return the live entry, not the verification snapshot** (`1ed55216`) — the branches restructured around one rule: *return whichever client the cache currently holds for the row's address; only dial when there is none*. That covers the reported case (replacement client dialed by another caller after a dead-transport invalidation — previously the known-bad snapshot was returned while its replacement sat in the map) and also upgrades the invalidated-but-unmoved-host case from "hand back the dead client" to "dial fresh". The restructure is net-simpler than the branch pair it replaces.

**Transitional sandboxes in drain progress** (`1ed55216`) — `ListHostsAdmin` gains `transitional_count` (`pausing`/`resuming`, i.e. lifecycle RPCs still using the host), surfaced as a BUSY column in `hostctl list` with the drain rule stated in the header comment: a host is not drained until RUNNING and BUSY are both zero. Previously a host mid-pause could read 0/0 and invite premature retirement, losing the in-flight snapshot.

Full unit + hostreg race suites and two consecutive fresh integration runs green. (Note for reviewers: the integration suite is not `-count=2`-safe — seven billing/Stripe tests fail on same-process rerun on clean main too; pre-existing fixture non-idempotence, unrelated to this PR.)

meAmitPatil:
@codex review

meAmitPatil:
Fixed in `8b90743c`. The cold path was indeed the last unguarded window — and worse, its mutex-across-I/O structure *guaranteed* the hazard, since a concurrent `Invalidate` couldn't even land until after the stale client had been returned. Rather than adding a second generation guard, `dialAndCache` is gone: first-use dials and due re-verifications now share **one** `resolveClient` path — read the row, return the cache's current entry when it matches the row's address, dial otherwise — with the generation snapshot/re-check around every read and dial, the mutex never held across I/O, and singleflight deduping concurrent resolutions (cold dial stampedes included, which the old write-lock serialized anyway). One resolution raced by a reclaim discards its result and re-reads, so the caller always receives a client for the newest committed address.

New deterministic test: cold dial to the pre-reclaim address blocks while the reclaim commits and invalidates → the dialed client is discarded, the row re-read, and the caller receives the newest address (dial sequence pinned, settled entry serves without re-dialing). Worth noting the rewrite's own test suite caught a regression before push — the read-error path initially consulted the pre-read entry snapshot instead of re-checking the map after the failure — which is the same snapshot-vs-current class as the last three findings; fixed and covered. Race-clean at -count=5, full unit + integration suites green.

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (commented):
Reviewing against SS-330 and the multi-VMD active/active design, I found one High-severity issue: self-registration currently publishes raw physical RAM/CPU from `/proc/meminfo` and `runtime.NumCPU()` into the host capacity fields. Those fields are becoming schedulable/admission limits for multi-host placement, so raw machine totals can overstate usable capacity and cause over-admission once placement begins consuming them. The registration path should use explicitly configured schedulable capacity/limits (or leave these fields to the later capacity-publication contract) rather than auto-detecting physical totals. Otherwise the host-row API shape looks directionally sound for multiple VMD instances: new hosts stay provisioning, activation is operator-gated, duplicate live host IDs are fenced, and reclaim demotes back to provisioning.

awille-ai (approved):

meAmitPatil:
@codex review

meAmitPatil:
@codex review

PR #394, opened 2026-08-14
vmd: advertise host self-description in heartbeats behind opt-in env

meAmitPatil:
Sender half of host self-registration (control-plane half: #393). When `VMD_ADVERTISE_ADDR`, `PROXY_ADVERTISE_ADDR`, and `HOST_REGION` are all set, the heartbeat carries the host's self-description (`vmd_addr`, `proxy_addr`, `region`, detected memory/vCPU capacity), letting a new host register itself as `provisioning` on first contact. A 409 from the control plane (identity held by another live host) is logged at error level as a provisioning problem; the daemon keeps running.

**Strictly opt-in, and why that matters:** control planes that predate #393 bind heartbeat bodies with `DisallowUnknownFields`, so a vmd that sent new fields to an old control plane would fail every heartbeat and get its host marked unhealthy within two minutes. The description is therefore attached only when the env opts in **and** is complete — with none of the vars set (every host today), the marshaled payload is byte-identical to the current format, pinned by a test that compares the exact bytes. Capacity detection failing (no `/proc/meminfo`) also disables advertising rather than registering an invented capacity.

Rollout order: this can merge and deploy in either order relative to #393, because nothing sends until the env vars are set per host — which happens only during new-host bring-up, after the cell's control plane has the registration handler.


meAmitPatil:
@codex review

meAmitPatil:
Folded into #393 (cherry-picked as-is): review on that PR correctly noted the registration feature couldn't engage end-to-end without the sender half, and the env opt-in already makes deploy order free — so the original reason for splitting (sender must not deploy ahead of the handler) no longer required two PRs. No changes to the commit; the byte-identical-payload guarantee and its tests carry over unchanged.

PR #397, opened 2026-08-15
vmd: close two revival races found after #387 merged

pavitrabhalla:
## Summary
- Two P1 races in the ReviveVM path, found by a local Codex review round after #387 merged.
- Startup reattach's interrupted-revival cleanup checked the VM op lock and released it immediately, so a fresh revive could start the instant the check passed and race the rest of the cleanup (stopping the newly launched replacement, overwriting its record). The lock now holds through the whole cleanup via `defer`.
- The idempotent shortcut (a retry matching a live, verified replacement's salvage path) lacked a destroy-wins recheck, and my first attempt at one had a torn-read window of its own (two separate reads, a destroy could complete between them). Fixed by taking the record-owner mutex `DestroyVM` holds for its entire run, making the tracked-ness check atomic with the whole destroy lifecycle.

## Technical decisions
- Both fixes reuse existing synchronization primitives from #387 (`tryLockVMOp`, `lockRecordOwner`) rather than introducing new ones.
- The destroy-wins recheck trades a small serialization window (blocking on the record-owner mutex if a destroy is mid-flight) for correctness; DestroyVM already does this for the same tradeoff.

## Testing
- `go vet` clean, full `internal/vm` suite green under `-race` in the Linux container.
- Local Codex holistic review pass (`codex exec review --base main`) came back clean: "No actionable correctness issues were found in the diff."

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -3086,15 +3086,20 @@ func (m *Manager) reattachRecord(ctx context.Context, rec VMRecord, cleanupStale
 		// first step) and already owns this record's fate; stopping its
 		// newly launched replacement or overwriting its state out from
 		// under it would be exactly the corruption this cleanup exists
-		// to prevent for a CRASHED attempt. Skip when the lock is held
-		// (non-blocking: this pass must not stall on a live revival) and
-		// let that attempt's own commit or rollback resolve the record.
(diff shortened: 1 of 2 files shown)


pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

meAmitPatil (approved):

PR #400, opened 2026-08-16
vmd: make the reconciler's per-host sandbox scan index-only

meAmitPatil:
The reconciler lists every non-destroyed sandbox on its host every 30s to detect drift. That query (`ListSandboxesByHost`) returned full sandbox rows plus a `LEFT JOIN snapshot`, which becomes expensive on hosts with large live-sandbox counts and drives sustained background DB load regardless of traffic.

This slims the hot path:

- The list query now selects only `id`, `status`, `snapshot_id` — the only fields the drift checks read — with no join. A new covering partial index (`host_id` INCLUDE `(id, status, snapshot_id)` WHERE `destroyed_at IS NULL`) lets the scan run **index-only**: no heap fetch, no join.
- Snapshot paths are needed only by the paused-snapshot drift check, so they're fetched in **one batched PK lookup** over just the paused candidates (`GetSnapshotPathsByIDs`) instead of joined across the whole inventory every pass.

Reconciler behavior is unchanged — it still builds the complete host inventory and detects orphans / dead VMs / missing snapshots identically; only the query shape and where snapshot paths come from changed. Can't just `LIMIT` the list (the drift pass needs the full set), so the win is making the full scan cheap rather than truncating it.

Tests updated (`failMissingSnapshots` now takes the resolved path map); `internal/vm` + `internal/db` pass.

### DIFF
--- a/db/queries/sandboxes.sql
+++ b/db/queries/sandboxes.sql
@@ -260,12 +260,21 @@ SELECT id FROM destroyed;
 SELECT EXISTS(SELECT 1 FROM sandbox WHERE id = $1 AND team_id = $2 AND destroyed_at IS NULL);
 
 -- name: ListSandboxesByHost :many
--- Used by the VMD reconciler. snapshot_path is joined so the paused-sandbox
--- drift check can stat the file without a per-row snapshot lookup.
-SELECT sqlc.embed(s), snap.path AS snapshot_path
+-- Used by the VMD reconciler's per-host drift pass. Returns only the fields
+-- the drift checks read (status + snapshot linkage) so the scan stays
+-- index-only via idx_sandbox_host_reconcile even on hosts with very large
+-- live-sandbox counts. Snapshot paths are fetched separately, in one batch,
+-- for just the paused candidates that need them (see GetSnapshotPathsByIDs) —
+-- avoiding a LEFT JOIN across the entire host inventory every pass.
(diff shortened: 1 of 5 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

pavitrabhalla:
Direction looks right — lean select + covering partial index matches the field-usage audit of the drift checks (id/status/snapshot_id are the only reads), and the fail-safe no-op when the batch lookup fails is the correct bias for Drift 4.

Three things:

1. The chunking comment says to split the ANY() batch "if a host ever holds so many paused sandboxes that the param gets unwieldy" — usw2 holds ~78K paused sandboxes today, so this lookup ships a ~1.2MB uuid[] and returns ~78K rows every 30s from day one. Can we chunk (e.g. 10K per call) in this PR rather than deferring?

2. `// ponytail:` in resolvePausedSnapshotPaths looks like a stray marker — intended?

3. The index pre-build runbook needs to run on staging + both prod cells before merge (CD applies migrations to every target sequentially). Has the CONCURRENTLY pre-build been executed anywhere yet? Also worth noting in the migration comment that on a table with this much churn, index-only will still heap-fetch recently modified pages until vacuum updates the visibility map — still a large win, just not literally zero heap.

pavitrabhalla (approved):

PR #402, opened 2026-08-16
api: serve the default sandbox list sort from an index

pavitrabhalla:
Every time anyone lists their sandboxes, the database re-sorts the entire list from scratch — for the biggest team that means grinding through ~80K rows and writing ~68MB of scratch files to disk on every single call, thousands of times a day. This adds a pre-sorted index so the database just reads the list in order: same results, ~200x faster on the paged path, zero disk churn.

GET /sandboxes sorts via a stack of guarded CASE ORDER BY terms, which the planner can never satisfy from an index — every call sorts the entire filtered set. For a team with ~80K live sandboxes that spills ~68MB to temp disk per call (external merge sort, measured on prod) and is the query behind the bulk of the database's cumulative temp-file volume.

- Adds static ORDER BY created_at DESC/ASC variants of the paged list query (identical filters and pagination, including the NULL-limit "return everything" compat contract) and routes the default sort to them; rare console-only name/status sorts keep the flexible CASE query.
- Adds partial index (team_id, created_at DESC) WHERE destroyed_at IS NULL, in the house guarded-migration pattern (manual CONCURRENTLY pre-build on populated DBs, INVALID-index check, lock/statement timeouts), plus a definition check that warns if a same-named index exists with the wrong shape.

Measured on prod (read-only), against the live 81K-row team:
- today's default path: external merge sort, 27MB+20MB+20MB temp disk across 3 workers, 198ms warm (p50 under load was ~1.4s, max 6.3s)
- same table with an exactly-matching partial composite index (demonstrated via the existing (team_id, status) index): LIMIT 100 → 1.15ms, no Sort node; full unpaginated list → 89ms, zero temp I/O — the planner picks the ordered index scan even for the no-LIMIT path.

Not addressed here (deliberate): the unpaginated path still returns every live row (compat contract), and a selective metadata filter may still choose the GIN index + sort. Acceptance criterion is eliminating the observed default-path spill.

Testing: sqlc regenerate idempotent; GOOS=linux build; internal/api + internal/db tests pass. Index pre-build must run on staging + both prod cells before merge (runbook in the migration header).

### DIFF
--- a/db/queries/sandboxes.sql
+++ b/db/queries/sandboxes.sql
@@ -110,8 +110,45 @@ WHERE base_path = $1 AND destroyed_at IS NULL;
 SELECT DISTINCT base_path FROM sandbox
 WHERE base_path IS NOT NULL AND destroyed_at IS NULL;
 
+-- Static created_at variants of ListSandboxesByTeamPaged. The plain ORDER BY
+-- lets the planner walk idx_sandbox_team_created_active (forward for DESC,
+-- backward for ASC) and stop at LIMIT, where the guarded-CASE ORDER BY in
+-- ListSandboxesByTeamPaged is opaque to it and forces a sort of the entire
+-- filtered set on every call. created_at is the default sort and the only one
+-- unpaginated SDK/MCP callers ever use, so these two cover the hot path;
+-- name/status sorts stay on the flexible query below. Filters and pagination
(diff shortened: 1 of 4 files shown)


pavitrabhalla:
@codex review

meAmitPatil (approved):

pavitrabhalla:
@codex review

PR #403, opened 2026-08-16
controlplane: bound db pool lifecycle and cap background pool sizes

pavitrabhalla:
When a traffic spike hit the west API, its database connection pools jammed and stayed jammed for over half an hour after the spike ended — and nobody could see it, because the west service exported no pool health data at all. This bounds every pool so bad connections get recycled instead of lingering, turns on the missing telemetry, and adds a plain-log alarm that names a wedged pool within a minute instead of after 40 minutes of cross-referencing dashboards.

- Control-plane pool: explicit lifecycle bounds (30m lifetime + 5m jitter, 5m idle, 30s health checks, 2 warm conns, 5s dial timeout). These act on idle/released connections; a held connection is out of the pool's reach, which is what the saturation log exists to surface.
- New sustained-saturation error log (not gated on metrics export): fires after 4 consecutive zero-idle samples with a canceled acquire in the episode. Cancellation is deliberately the only evidence — callers abandoning waits is unambiguous distress, where empty-acquire counts also cover routine scale-up construction and would page on cold bursts. Accepted boundaries are documented in-line (no-deadline callers don't exist here; a fully unreachable DB is already loud in per-request error logs).
- vmd and secretsproxy pools get explicit MaxConns=8: the pgxpool default is max(4, NumCPU), i.e. 192 each on the 192-core metal host, silently consuming the cell pooler's client budget.
- DB_MAX_CONNS stays 15 — deliberately. Raising it (25, then 20) was evaluated and rejected: max_instances applies per revision and a deploy under full load can overlap old+new revisions completely, so the binding ceiling is 60 x DB_MAX_CONNS against the XL pooler's 1,000 clients. 15 -> 900 + capped host services ~= 940 worst case; any raise breaks deploys-under-load. The terraform guardrail comment now carries this overlap math. Burst headroom comes from fast queries and the lifecycle bounds, not a larger cap.
- West gets the OTel metrics env + firewall rule that east already had — the west service was exporting no db_pool_* metrics, which is why the pool's state during the incident is unknowable. Noted in-line: a standby promotion must re-run the otel deploy or export silently drops (non-fatal).

Testing: unit tests for the saturation detector state machine (8 cases incl. hold/arm/re-arm semantics); gofmt/vet; GOOS=linux build of all packages; internal/telemetry, internal/api, internal/db tests; terraform fmt. cpu_idle and Cloud Run concurrency deliberately untouched.

### DIFF
--- a/cmd/controlplane/main.go
+++ b/cmd/controlplane/main.go
@@ -148,7 +148,10 @@ func run() error {
 	poolCfg.MaxConnLifetimeJitter = 5 * time.Minute
 	poolCfg.MaxConnIdleTime = 5 * time.Minute
 	poolCfg.HealthCheckPeriod = 30 * time.Second
-	poolCfg.MinIdleConns = 2
+	// Clamped to the cap: a DB_MAX_CONNS below 2 is a valid override, and
+	// pgxpool rejects (rather than clamps) a minimum above the maximum,
+	// which would fail startup instead of running with a smaller pool.
+	poolCfg.MinIdleConns = min(2, poolCfg.MaxConns)
 	poolCfg.ConnConfig.ConnectTimeout = 5 * time.Second
 	dbPool, err := pgxpool.NewWithConfig(ctx, poolCfg)
 	if err != nil {
--- a/cmd/controlplane/main.go
+++ b/cmd/controlplane/main.go
@@ -137,6 +137,19 @@ func run() error {
 			poolCfg.MaxConns = int32(n)
 		}
 	}
+	// Lifetime + jitter bound how long any one connection can persist and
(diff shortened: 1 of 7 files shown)


pavitrabhalla:
@codex review

pavitrabhalla:
Fixed: MinIdleConns now clamps to the configured cap, so an undersized DB_MAX_CONNS override runs with a smaller pool instead of failing startup.

@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

meAmitPatil (approved):

PR #405, opened 2026-08-17
vmd: persist dirty-tracking-armed state so a restart doesn't force a full pause

meAmitPatil:
A vmd restart reattaches to live Firecracker processes but loses the in-memory `DirtyTracked` flag, so each running VM's next pause writes a full memory snapshot instead of a fast diff (a slow-pause burst after every deploy, self-healing on the next resume). This persists the armed bit and re-arms it on reattach — **gated behind `VMD_PERSIST_DIRTY_TRACKING` (default off)** until we verify Firecracker's dirty bitmap survives a restart, since re-arming against a reset bitmap would corrupt the diff.

Correctness details:
- **Persist is gated on the flag** (`toRecord(inst, persistDirtyArmed)`) — so with the feature off, no `armed` bit is ever written, no stale bit can exist, and enabling is always safe (a VM whose bit was never persisted just takes one full pause, then heals).
- **Write-ahead disarm**: durably clear the armed bit before any bitmap-resetting snapshot, refuse the snapshot if the clear fails. `PauseVM` and `CreateVMSnapshot` both hold the vm-op lock across disarm+snapshot, so no concurrent lifecycle write can re-arm in the gap (`CreateVMSnapshot` is now serialized for exactly this reason).
- **Field-level clear** edits the raw JSON (`dirty_tracking_armed` key only), preserving any newer fields under rollback/mixed-version; uses `Update` (no batch-coalescing delay).
- Re-arm only for a confirmed-Running reattach.

Also adds `snapshot_type` (layered/diff/full) + `snapshot_ms`/`stop_ms`/`pause_ms` to the pause log (not gated) so full-vs-diff pauses are visible; `pause_ms` starts before the vm-op lock so it includes queueing.

Cost: with the flag off there is no added lifecycle write. With it on, the disarm is one extra bbolt commit on the pause path (the pause already does one for its final persist) — small relative to the snapshot the pause writes, and it exists to avoid the multi-second full pause. Not micro-benchmarked; the intent is qualitative (one small commit), and the new `pause_ms`/`snapshot_ms` fields make the real cost measurable once enabled.

Before enabling: verify bitmap-survival end to end (dirty known pages before and after a restart, restore base+diff, validate both sets).

### DIFF
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -450,6 +450,7 @@ func main() {
 	resumeUffdEnabled := envOrDefault("VMD_RESUME_UFFD", "false") == "true"
 	verifySnapshotEnabled := envOrDefault("VMD_VERIFY_SNAPSHOT_ENABLED", "false") == "true"
 	incrementalSnapshotEnabled := envOrDefault("VMD_INCREMENTAL_SNAPSHOT", "false") == "true"
+	persistDirtyTrackingEnabled := envOrDefault("VMD_PERSIST_DIRTY_TRACKING", "false") == "true"
 	handlerDeathAbortEnabled := envOrDefault("VMD_HANDLER_DEATH_ABORT", "false") == "true"
 	// Tri-state: "auto" (default) lets vmd enforce only after its convergence
 	// sweep proves every layered overlay has a presence side-car; "always"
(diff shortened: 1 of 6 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
Closing: the durable dirty-tracking-armed record field is the wrong mechanism. Mirroring Firecracker's dirty-bitmap state transactionally in vmd is a permanent corruption-sensitive state machine (reattach writes the record outside the vm-op lock, rollback to a binary that doesn't disarm leaves a stale armed bit, and it needs a write-ahead disarm + fsync on the pause path). The correctness invariant belongs at the component that owns the bitmap: Firecracker will expose a monotonic dirty-bitmap generation, and vmd will compare-not-mirror on reattach (re-arm only on an exact process-identity + generation match, else one conservative full pause). Superseded by that design. The pause phase-timing telemetry from this branch will be re-landed separately.

PR #408, opened 2026-08-17
backup: read-only discrepancy report over the uncovered paused fleet

pavitrabhalla:
## Summary

Adds a `discrepancy` subcommand to `cmd/backup-restore` that classifies every paused, non-destroyed sandbox lacking a verified backup generation in the audited bucket into exactly one class (`no_snapshot_row`, `stale_zero_byte`, `sweep_missed`) and emits a JSON ledger plus a per-class summary. This is the discrepancy-report deliverable of the sandbox recovery and durability effort: it turns an opaque uncovered count into named causes with per-row host attribution, so hosts whose paused fleet predates the uploader are distinguishable from sweep gaps.

## Technical decisions

- The snapshot join targets the head pointer only (`sandbox.snapshot_id`, maintained by pause finalize), never a newest-by-created_at lateral over snapshot history: a NULL head must classify as `no_snapshot_row` even when historical rows exist, since the FK is ON DELETE SET NULL and the reaper deletes snapshot rows.
- Coverage is `NOT EXISTS` over `backup_generation` scoped to the audited bucket (`-bucket` is required and recorded in the ledger): the uploader and sweep track coverage per bucket, so rows pointing at a rotated-out bucket never mask a gap. No recency ranking beyond that: the report enumerates the never-covered backlog, deliberately.
- Raw pgx with inline SQL, matching the rest of the tool: DR tooling must run when the control plane is down. DB-only is the default; `-probe-bucket` adds an optional GCS confirmation pass.
- Read-only by construction; the ledger warns that counts churn while placement is live rather than refusing to run.

## Testing

- `go test ./internal/backup/... ./cmd/backup-restore/...` and `go vet` pass (linux, Go 1.25): exhaustive classification table tests (NULL head, zero-byte old/recent, sized old, exact stale-days boundary, large stale-days horizon), aggregate counting, and a golden test pinning the ledger JSON shape.
- No SQL parser is in the module, so the query text is pinned by an invariant test (head-pointer join, bucket-scoped `NOT EXISTS`, selection predicates, no LATERAL/recency clauses). Execution verified against Postgres 16 with a fixture covering all classes plus the exclusions, including a sandbox whose only `backup_generation` rows name a different bucket, which enumerates as uncovered as required.

### DIFF
--- a/cmd/backup-restore/discrepancy.go
+++ b/cmd/backup-restore/discrepancy.go
@@ -1,0 +1,259 @@
+package main
+
+// The discrepancy subcommand classifies every paused, non-destroyed
+// sandbox lacking a verified backup generation in the audited bucket
+// into exactly one class (no_snapshot_row, stale_zero_byte,
+// sweep_missed) and emits a JSON ledger plus a human summary. Read-only
+// by construction: it runs SELECTs against the control-plane DB,
+// optionally lists the backup bucket, and writes nothing but the local
+// report file.
+//
+// Like the rest of this tool it talks raw SQL over pgx rather than the
+// control plane's query layer: DR tooling must keep working when the
+// control plane is down, so its only inputs are a DB URL and the name
+// of the bucket being audited. The default run needs no GCS
+// credentials; -probe-bucket adds an optional bucket confirmation pass
+// over the classified rows.
+
+import (
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"os"
+	"os/signal"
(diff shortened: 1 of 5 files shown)


pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

meAmitPatil (approved):

PR #411, opened 2026-08-17
backup: carry exact object paths through verified-generation reports

pavitrabhalla:
## Summary

Verified-generation reports recorded only names, sizes, and digests, so the backup_generation.files ledger could not name a generation's bucket objects (object names embed the extent-packing fingerprint, known only at upload time). This PR captures each upload's exact object path from the store write itself and carries it through the BoltDB outbox, the report wire format, and into the files jsonb, making the ledger sufficient for a future DB-driven lifecycle GC to name objects without listing the bucket. Part of the sandbox recovery and durability effort; no deletion or read path is implemented here.

## Technical decisions

- Single naming source: the recorded path is the literal value handed to the blob store's create call, returned from the same function that writes the object, so it cannot drift from the write side. Control-plane validation rebuilds the expected prefix from the same layout helpers the uploader uses.
- Durable/wire compatibility, all additive with omitempty:
  - BoltDB outbox: pre-upgrade entries decode and deliver unchanged (tested against a byte-for-byte legacy entry).
  - Old vmd + new control plane: pathless reports are accepted and stored exactly as before.
  - New vmd + old control plane: the control plane binds report JSON strictly, so the unknown field is rejected; the reporter detects that specific rejection and retries once with the paths stripped, landing the report exactly as an old vmd would instead of dropping the coverage row. Generations reported through that window keep coverage but stay pathless (their bucket manifests remain the path authority); an upsert enrichment arm upgrades a pathless row if a rich report is redelivered, without moving completed_at or the reported_at freshness cap.
- Ledger integrity for future GC: paths are validated against the owner's layout (fingerprint suffix must be hex; charset pinned, not length, so a future fingerprint widening is not a silent incompatibility), accepted all-or-none per report, and frozen once recorded, since the bucket manifest they mirror is immutable. A completion whose manifest create deduped reports no paths (the earlier attempt's published manifest is the authority), and a pathless re-ack never demotes an undelivered path-bearing outbox entry.

## Testing

- Unit: recorded paths exactly match the store keys written (both the streamed and history-skip shared-base paths); outbox round-trip with and without paths including a pre-upgrade entry; deduped-manifest completions report pathless; pathless re-ack keeps the richer outbox entry; reporter carries paths on the wire, omits the key for legacy entries, strip-retries only on the unknown-field rejection; handler accepts/persists reports with and without paths and rejects foreign, unfingerprinted, or mixed paths.
- Integration: enrichment ordering over backup_generation (pathless never demotes, equal-instant rich redelivery enriches without touching completed_at/reported_at, recorded paths frozen).
- go build ./..., go vet, and the backup/vm/api unit suites plus the backup-report integration suite run on linux.

### DIFF
--- a/db/queries/backups.sql
+++ b/db/queries/backups.sql
@@ -9,18 +9,51 @@
 -- it exceeds now()) yields to ANY smaller incoming one, sane or merely
 -- less skewed, so repair is monotone even when the correction lands
 -- while some skew remains; redelivery of either report matches neither
--- arm and stays a no-op.
+-- arm and stays a no-op. The third arm lets a report carrying object
+-- paths enrich a row that lacks them regardless of freshness: a
+-- rollout-window fallback can land a pathless copy of the same
+-- verification first (same pinned instant, so the freshness arms never
+-- fire for the redelivery), and the paths are what a lifecycle GC
+-- reasons from. Enrichment leaves completed_at where it stands (the
+-- CASE below): an older rich redelivery must not regress freshness.
 INSERT INTO backup_generation (sandbox_id, generation, bucket, completed_at, files)
 VALUES ($1, $2, $3, $4, $5)
 ON CONFLICT (sandbox_id, bucket, generation) WHERE sandbox_id IS NOT NULL
-DO UPDATE SET completed_at = excluded.completed_at, reported_at = now(),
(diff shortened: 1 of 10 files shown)


pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

meAmitPatil (approved):

PR #414, opened 2026-08-18
vmd: keep the socket up on steady-state deploys and bound daemon shutdown

meAmitPatil:
Gate the early `stop` in the vmd deploy so steady-state deploys keep the systemd socket up (connections backlog across the swap instead of a throwaway vmd being socket-activated and killed). Bound vmd's shutdown with a per-closer deadline plus `TimeoutStopSec` so a stop can never drag on.

### DIFF
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -1176,9 +1176,13 @@ func main() {
 		// adopt-in-cutover would cost.
 		reattachStart := time.Now()
 		reattached, stale := mgr.ReattachAll(ctx)
-		log.Info().Str("startup_phase", "reattach_background").
+		// Concurrent with the sequential startup partition (runs in this
+		// goroutine), so its phase_ms must NOT be summed with the partition
+		// marks. Flagged concurrent=true and given a distinct message so a
+		// grep for "startup phase timing" excludes it from the partition sum.
+		log.Info().Str("startup_phase", "reattach_background").Bool("concurrent", true).
 			Dur("phase_ms", time.Since(reattachStart)).Int("count", reattached+stale).
-			Msg("startup phase timing")
+			Msg("startup background phase timing")
 		if reattached > 0 || stale > 0 {
 			log.Info().Int("reattached", reattached).Int("stale", stale).Msg("startup reattach complete")
 		}
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -193,6 +193,11 @@ type lifecycle struct {
 	// receipt: only a deliberate, fully clean shutdown may vouch.
 	signalInitiated bool
 	closerErr       error
+	// forcedRPCStop is set when the gRPC drain hit its budget and active RPCs
+	// were force-cancelled. A cancelled create/pause/resume may have left state
+	// the next boot must reconcile, so this bars the pool receipt — the drain
+	// budget is a circuit breaker, never a proven-safe cancellation bound.
+	forcedRPCStop bool
 
 	done   chan struct{}
 	doneCh sync.Once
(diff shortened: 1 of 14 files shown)


meAmitPatil:
@codex review

cmd/vmd/main.go
```diff
@@ -268,9 +268,27 @@ func (lc *lifecycle) wait(ctx context.Context) {
...
-// shutdown runs every registered closer in reverse order, collecting
-// errors but never stopping on the first failure — we want every
-// resource to get a chance to clean up.
+// perCloserShutdownTimeout bounds each closer independently during shutdown.
+// A closer that ignores cancellation must not stall the closers after it or
+// hang the process — the loop aborts the graceful sequence if one overruns,
+// so total shutdown stays bounded. systemd TimeoutStopSec is the final backstop
+// if the process itself wedges past this.
+// ponytail: a circuit-breaker budget, not a correctness assumption — tune freely.
+var perCloserShutdownTimeout = 12 * time.Second
```
awille-ai: **High — the 12s budget is currently a correctness-affecting cutoff, not just a circuit breaker.** This same timeout is passed to the gRPC closer; when it expires, that closer calls `grpcServer.Stop()` (cancelling active RPCs) and still returns `nil`. We have not shown that create/pause/resume are all safely cancellable within 12s, and the current lifecycle model still has known crash windows that require host reconciliation. A deploy can therefore force-cancel a legitimate mutating RPC and still later classify shutdown as clean. Please either separate the RPC drain from the generic closer budget and use a justified/proven drain bound, or make a forced gRPC stop mark the shutdown dirty so it cannot be treated as a clean handoff.

awille-ai (commented):
Reviewed against SS-424 as the immediate mitigation slice, with the longer-term durable-intent / fast-handoff direction in mind.

The steady-state deploy-ordering change is good: keeping the socket up and removing the early service stop eliminates the specific throwaway-VMD/double-restart race. The exceptional socket/journal migration branches remain fail-closed, and CI + Terraform Checks are green.

Two High-severity issues remain before I would treat this as a safe production mitigation:

1. The new 12s generic closer budget also bounds the gRPC drain. When it expires, the gRPC closer force-stops active RPCs but returns nil, so the shutdown can still be classified clean. We have not established that 12s is a safe upper bound for create/pause/resume or that all of those mutation paths are crash-consistent yet. This makes the timeout a correctness assumption rather than only a circuit breaker.

2. The deploy still sleeps 3s and checks `systemctl is-active`. The unit is `Type=simple`, while measured prod startup is ~12s and `startupReady` is the actual request gate. The script can therefore report a successful deployment and disarm recovery while the new VMD is still unavailable—or even before a later startup failure. SS-424 explicitly calls out application-aware readiness as part of the immediate fix.

Test gap (Medium): the new lifecycle unit test covers closer-abort behavior, but there is no automated coverage for the deploy-ordering state machine itself: steady-state socket active -> no early stop; first socket migration/socket change -> stop both; journal-path change -> stop both; exactly one final service restart. Given that this exact shell ordering caused the incident, a small testable helper/state-machine extraction or equivalent high-signal coverage would materially reduce regression risk.

I do not see a conflict with the longer-term durable lifecycle/journal approach. #414 is still a useful short-term layer once the two safety issues above are addressed. The service/deploy configuration has no generated infrastructure-plan analogue available here; Terraform Checks passed, so I treated plan evidence as non-blocking/unverified rather than a finding.

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

PR #415, opened 2026-08-18
api: hard-fail routing when a sandbox references an unregistered host

meAmitPatil:
Removes the two remaining silent-fallback paths that route operations to the default VMD client when a host row lookup fails:

- **`vmdForHost`**: a sandbox whose `host_id` has no host row previously fell back to the default client — a migration-period convenience that becomes a misroute the moment more than one host exists, and in a cell whose real host id coincides with the default, the misroute never even fails loudly. It is now a hard error with a distinct, alertable log line (`host_not_registered`). The registry-less mode (`Hosts == nil`, test/dev wiring) is unchanged.
- **Build resolver**: the template-build supervisor's warn-and-fall-back on host lookup failure now surfaces the error; the supervisor's dispatch path already fails the build cleanly on resolver errors, which is honest — a silently misrouted build is not.

**Pre-deploy verification** (read-only, run per cell): every live sandbox must reference an existing host row before this ships —

```sql
SELECT COUNT(*) FROM sandbox s
LEFT JOIN host h ON h.id = s.host_id
WHERE h.id IS NULL AND s.destroyed_at IS NULL;
```

Expected 0 in both cells (each cell's single host row has existed since the host table shipped). If nonzero, those rows are pre-existing corruption this change would surface on first touch — which is the point, but better to know up front.

No behavior change for any correctly-referenced sandbox: the lookup path and its result are identical; only the failure handling differs.

### DIFF
--- a/cmd/controlplane/main.go
+++ b/cmd/controlplane/main.go
@@ -263,10 +263,14 @@ func run() error {
 		if hostID == "" || handlers.Hosts == nil {
 			return vmdClient, nil
 		}
+		// No fallback on lookup failure: dispatching a build through the
+		// default client sends it to whichever machine that happens to be,
+		// and with multiple hosts that is the wrong one. The supervisor's
+		// dispatch path already fails the build cleanly on resolver errors,
+		// which is honest; a silently misrouted build is not.
 		c, err := handlers.Hosts.ClientFor(rctx, hostID)
 		if err != nil {
-			log.Warn().Err(err).Str("host_id", hostID).Msg("supervisor: host lookup failed, using default client")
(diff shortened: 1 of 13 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):

PR #417, opened 2026-08-18
api: pace the reaper maintenance sweeps on their own ticker

meAmitPatil:
The reaper's DB-only sweeps are backstops for cleanup the request path already does, but they ran on every 30-second poll on every instance, each scanning accumulated history to almost always find nothing.

They now run on a separate 10-minute ticker, still once at startup so a restart after a crash cleans up promptly; expiry and auto-delete keep the poll interval.

### DIFF
--- a/internal/api/reaper.go
+++ b/internal/api/reaper.go
@@ -14,10 +14,17 @@ import (
 	"github.com/superserve-ai/sandbox/internal/telemetry"
 )
 
+const defaultSweepInterval = 10 * time.Minute
+
 // ReaperConfig controls the timeout reaper loop.
 type ReaperConfig struct {
 	// Interval is how often the reaper polls for expired sandboxes.
 	Interval time.Duration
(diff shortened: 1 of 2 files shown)


meAmitPatil:
@codex review

meAmitPatil:
Fixed in 32e5d20f. Valid on both counts, and the first one was a regression I introduced: the single-ticker loop this replaced deliberately ran auto-delete last so a slow teardown could not delay the sweeps, and folding auto-delete into runTick put it back in front of them at startup.

The sweeps now run on their own goroutine rather than as another case in the poll loop, so neither the startup batch nor a slow poll can hold them off — the select they shared was the coupling. Concurrency between the two loops is not a new interleaving: every control-plane instance already runs both against the same tables.

meAmitPatil:
@codex review

awille-ai (approved):

PR #420, opened 2026-08-18
vmd: log startup phase durations and inventory counts

meAmitPatil:
Observability-only instrumentation so a single prod-sized vmd restart answers both "which phase owns the ~12s" and "why," without a second restart. No behavior change, no new scans, no readiness- or hot-path work — every timer/counter piggybacks on work already happening.

## Top-level partition
A running-mark model logs each startup phase as the interval since the previous mark, so the segments tile the whole start→ready timeline with no gap (sum of phase_ms == since_start_ms at ready — nothing can hide between timers):

network_firewall, vm_manager_init, host_egress_block, state_store_open, cgroup_arm, backup_setup, backup_staging_scan, grpc_setup, slot_reserve, cgroup_reap, namespace_sweep, pool_start, db_connect, pre_ready, ready.

reattach_background (the O(N) reattach) runs concurrently in its own goroutine — flagged concurrent=true with a distinct message so it is NOT summed into the partition.

## Per-phase breakdowns (why a phase is slow)
- state_store_open: bolt_open_ms / policy_scan_ms / record_scan_ms + records, policies, orphans_deleted.
- network_firewall: pre_network_ms vs firewall_install_ms.
- cgroup_arm: cgroup_detect_ms (record + tree scan) split from systemd/cgroup validation.
- slot_reserve: record_load_ms vs reserve_reclaim_ms + records, reservations, reclaimed.
- cgroup_reap: scanned, recorded, recordless, reaped, protected.
- backup_staging_scan: renewed_markers (durable pending markers) + sandboxes, generations, bases, pending.
- pool adoption (AdoptOrphanSlots): duration_ms + candidates alongside its existing adopted/torn_down/skipped.

## Scale (the N to correlate against)
vm_records from reattach_background; host netns_total / host_mount_count from the existing periodic samplers. No dedicated inventory scans on the startup path.

Touches cmd/vmd/main.go plus internal {vm,backup,network}; StateStore/SweepStaging/RenewPendingStaging gained return values used only for telemetry (existing callers ignore them). gofmt clean, Linux build + affected vm/backup/network tests green.

### DIFF
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -1176,9 +1176,13 @@ func main() {
 		// adopt-in-cutover would cost.
 		reattachStart := time.Now()
 		reattached, stale := mgr.ReattachAll(ctx)
-		log.Info().Str("startup_phase", "reattach_background").
+		// Concurrent with the sequential startup partition (runs in this
+		// goroutine), so its phase_ms must NOT be summed with the partition
+		// marks. Flagged concurrent=true and given a distinct message so a
+		// grep for "startup phase timing" excludes it from the partition sum.
+		log.Info().Str("startup_phase", "reattach_background").Bool("concurrent", true).
 			Dur("phase_ms", time.Since(reattachStart)).Int("count", reattached+stale).
-			Msg("startup phase timing")
+			Msg("startup background phase timing")
 		if reattached > 0 || stale > 0 {
 			log.Info().Int("reattached", reattached).Int("stale", stale).Msg("startup reattach complete")
 		}
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -317,6 +317,34 @@ func runDrainCheck() int {
 	return 3
 }
 
+// startupTimer emits one structured log line per startup phase — the phase's
+// own wall-clock duration, the cumulative time since the daemon began starting,
+// and, where cheap to compute, the aggregate count the phase processed.
+// Observability only: it changes no control flow and runs only on the startup
+// path, so a prod-sized restart reveals which phase dominates and what scales
+// with fleet size.
+type startupTimer struct {
+	log   zerolog.Logger
+	start time.Time
+}
+
+func newStartupTimer(log zerolog.Logger) *startupTimer {
+	return &startupTimer{log: log, start: time.Now()}
+}
+
+// done records a phase that began at t0. Pass count >= 0 to log an aggregate
+// size (VM records, staged entries); pass -1 to omit it.
+func (s *startupTimer) done(phase string, t0 time.Time, count int) {
+	e := s.log.Info().
+		Str("startup_phase", phase).
+		Dur("phase_ms", time.Since(t0)).
+		Dur("since_start_ms", time.Since(s.start))
+	if count >= 0 {
+		e = e.Int("count", count)
+	}
+	e.Msg("startup phase timing")
+}
+
 func main() {
 	// Maintenance subcommands run before any daemon setup and exit. They must
 	// not open the state store in write mode or start services.
@@ -391,6 +419,11 @@ func main() {
 
 	lc := newLifecycle(log)
 
+	// Startup phase timing — observability only (see startupTimer). Started as
+	// early as the daemon path allows so since_start_ms approximates the full
+	// time to readiness.
+	st := newStartupTimer(log)
+
 	// ---- Egress blocklist (optional) ----
 	// VMD_EGRESS_BLOCKLIST_CONFIG points at the operator-supplied config
 	// (feeds, pinned domains/CIDRs, blocked ports). Unset = no global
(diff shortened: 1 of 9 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

cmd/vmd/main.go
```diff
@@ -649,7 +684,9 @@ func main() {
 		// the first feed fetch (which may block up to feedFetchTimeout per
 		// feed). Otherwise seeded CIDRs go unenforced on non-proxied ports
 		// during the startup window.
-		hostBlock.UpdateCIDRs(blockList.CIDRs())
+		seedCIDRs := blockList.CIDRs()
+		hostBlock.UpdateCIDRs(seedCIDRs)
+		st.mark("host_egress_block", len(seedCIDRs))
 		lc.addCloser("host egress block", func(_ context.Context) error { return hostBlock.Close() })
```
awille-ai, line 775: **High — skipped optional branches make the phase labels non-comparable.** `mark()` measures from the previous mark, but this mark only runs when `blockList != nil`. If the branch is disabled, `state_store_open` includes the time since `vm_manager_init`; the same pattern exists for backup, namespace sweep, and DB-connect branches. For the startup partition to be useful across prod hosts/configs, emit each boundary even when skipped (e.g. `enabled=false`, zero work/count), or use explicit start/end timers per phase.

cmd/vmd/main.go
```diff
@@ -1103,13 +1163,24 @@ func main() {
 	// started after StartPool so its first read observes an initialized pool.
 	mgr.StartNetnsLeakSampler(ctx, time.Minute)
```
awille-ai: **High — the startup record does not actually capture the fleet-size correlates needed to interpret the phase timings.** These samplers start on a one-minute ticker and do not sample immediately, so the restart log can reach `ready` without a `netns_total`/`host_mount_count` value from the same startup. We specifically need those Ns to distinguish fixed work from fleet-scaled work. Please emit one immediate sample when starting the samplers (or attach counts from already-required startup enumerations) without adding a new fleet-sized readiness scan.

awille-ai (commented):
Reviewed specifically for whether one prod-sized restart gives us enough signal to make the architectural decision around durable lifecycle state / fast handoff.

The PR is directionally very good: the breakdowns for Bolt open/scans, firewall install, slot reservation, cgroup reap, backup staging, and pool adoption are exactly the kinds of signals we need, and they piggyback on existing work rather than adding dedicated scans. CI and Terraform Checks are green.

I found two High signal-quality gaps that would make a single restart less conclusive than the PR claims:

1. Optional startup branches do not always emit a phase mark. Because `startupTimer.mark()` measures time since the previous mark, skipping `host_egress_block`, `backup_setup`/`backup_staging_scan`, `namespace_sweep`, or `db_connect` causes the next emitted phase to absorb that interval. That means a phase name does not have stable semantics across hosts/configurations, and the claimed contiguous partition is not reliably attributable. Please emit every phase boundary unconditionally (with enabled/skipped/count fields as appropriate), or otherwise make phase start/end explicit so skipped branches contribute zero rather than changing the meaning of the next phase.

2. The fleet-size correlates we care about are not guaranteed to be present in the startup sample. `StartMountCountSampler` and `StartNetnsLeakSampler` use ticker-driven samplers whose first sample occurs after the interval; with the current 1-minute cadence, the startup event itself has no guaranteed `host_mount_count` / `netns_total`. Since the whole architectural question is whether phase latency scales with fleet-sized host state, emit a one-shot count during startup (ideally from existing enumerations / cheap proc or readdir work, without adding another fleet-sized scan), or make the samplers emit once immediately when started.

There is also an existing P2 comment on this head that the timer starts after preliminary startup work, so `since_start_ms` is not truly daemon-start→ready. I agree with that: either move timer initialization to the earliest point after logging is available and create an explicit preliminary phase, or narrow the stated contract.

With those fixed, I think this instrumentation should answer the questions we actually need: which synchronous phase owns the ~12s, what its relevant N is, and whether that work is something durable intent/provenance could let us skip or defer. I do not see a reason to add CPU pprof/runtime trace to this PR yet; phase timers + counts are the right first measurement because they preserve the real wall-clock timing we care about.

awille-ai (commented):
Reviewed latest head with the usual bar. Exact-head CI and Terraform Checks are green, and the earlier duplicate inventory/staging scans were removed. I still see two High observability-correctness issues: (1) phase boundaries are skipped when optional branches are disabled, so the next phase absorbs that time and labels are not comparable across configurations; (2) netns_total/host_mount_count are only emitted by one-minute samplers, so a startup can reach ready without the fleet-size correlates this instrumentation is intended to capture. There are also two lower-severity existing P2s around the timer starting after preliminary startup work and adoption timing not being emitted on all completion paths. No Critical issue, so COMMENT rather than request changes.

awille-ai (commented):
Reviewed latest head with the usual bar. I found two High findings that are both material to this PR's stated goal of making a single restart sufficient for diagnosis: (1) phase boundaries are skipped when optional branches are disabled, so later `phase_ms` values absorb unrelated work and become non-comparable across configurations; (2) the fleet-size correlates (`netns_total` / `host_mount_count`) are produced by one-minute samplers that do not sample immediately, so a restart can reach ready without the Ns needed to interpret startup phase timings. Exact-head CI and Terraform Checks are green. Per our review policy these are comment-only High findings, not blocking.

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

PR #422, opened 2026-08-18
vmd: keep the socket up on steady-state deploys and bound daemon shutdown

meAmitPatil:
Deploy-safety fixes for the vmd CD path. Fresh PR — supersedes the previous one, whose head got stuck after a branch rewind; identical content at `edcb151a`. The startup-observability work lives in its own PR and is intentionally NOT here, so this can deploy once on its own.

**Deploy ordering.** The early service stop is gated to the exceptional cases only (first socket migration / socket-definition change), and stops both the socket and the service so no interim vmd can socket-activate during the window. Steady-state deploys keep the socket up and do a single restart with connections backlogging — no throwaway-vmd double-restart.

**Bounded, abort-safe shutdown.** Each closer runs under its own deadline off the main goroutine; any error or overrun aborts the remaining dependency closers (no use-after-close) and leaves reclamation to process exit. A forced gRPC stop (drain budget exceeded) marks the shutdown unclean so it can't vouch the pool receipt, and returns an error so it aborts too — `grpc.Server.Stop()` doesn't wait for handlers, and detached (`WithoutCancel`) handlers like ResumeVM can still be using the store/pool/DB. `TimeoutStopSec` is the systemd backstop.

**Application-aware deploy readiness.** After the restart the script waits for vmd's real request gate (the `startupReady` log line) rather than `is-active` (which, for a `Type=simple` unit, only proves the process forked). The wait is scoped to the current systemd invocation (so a crashed-and-replaced process can't false-pass) and drives off `journalctl` exit status, not captured stdout (which would read `-- No entries --` as a false ready).

Tests: lifecycle/shutdown unit tests (incl. forced-stop, abort-on-error) and Python deploy-ordering tests (guarded stop, single restart, exit-status readiness with a no-match case) — now run in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,9 @@ jobs:
       - name: Run unit tests
         run: go test -v -short -count=1 -race ./...
 
+      - name: Run deploy-script tests
+        run: python3 -m unittest discover -s .github/workflows/scripts -p 'test_*.py'
+
   integration-tests:
     name: Integration Tests
     runs-on: ubuntu-latest
--- a/.github/workflows/scripts/deploy-vmd.py
+++ b/.github/workflows/scripts/deploy-vmd.py
@@ -1053,14 +1053,16 @@ def main() -> int:
             READY=0
             for i in $(seq 1 90); do
                 INVOCATION=$(systemctl show -p InvocationID --value {service} 2>/dev/null || true)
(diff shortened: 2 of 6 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):

PR #423, opened 2026-08-18
api: carry preview access in the sandbox list query instead of a team-wide fetch

meAmitPatil:
Listing sandboxes fetched preview policies for the team's entire sandbox set and then used that map to decorate a single page, so the cost scaled with the tenant rather than the page — and it did so even for callers that paginate correctly. On the largest teams that is six figures of rows read per request.

The three list queries now carry the effective preview access through a LEFT JOIN, matching the shape GetSandboxWithPreviewPolicy already uses for single reads, and the team-wide query is removed. The join is 1:1 on the policy table's primary key, so it cannot multiply rows and the LIMIT still pushes down: EXPLAIN on the largest team shows a nested loop doing one primary-key lookup per returned row, reading 20 rows rather than the full set.

No response-shape or behavior change. Reading one snapshot also removes a small race where a sandbox destroyed between the two queries fell back to the default access. Integration test covers both the no-policy default and an explicit policy across the static and flexible query paths.

### DIFF
--- a/db/queries/sandboxes.sql
+++ b/db/queries/sandboxes.sql
@@ -119,28 +119,38 @@ WHERE base_path IS NOT NULL AND destroyed_at IS NULL;
 -- name/status sorts stay on the flexible query below. Filters and pagination
 -- are identical to the flexible query, including the NULL @row_limit "return
 -- everything" contract.
+--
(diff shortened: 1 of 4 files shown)


meAmitPatil:
@codex review

awille-ai (approved):

PR #424, opened 2026-08-19
vmd: replace startup record scans with durable, trust-stamped projection indexes

meAmitPatil:
Removes the three synchronous full-record scans that dominate vmd startup by maintaining two sparse projection indexes (cgroup membership, vmID→namespace) inside the record's own transactions, trusted across restarts via a clean-close stamp of the store's own Bolt txid — any other writer invalidates it and the boot falls back to a single fail-closed projection rebuild. No new transaction or fsync on any lifecycle path; benchmarks at 100k records: trusted open ~0.24ms, rebuild ~374ms, <1% delta on record writes.

### DIFF
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -791,9 +791,25 @@ func main() {
 			Dur("record_scan_ms", oss.RecordScan).Dur("tx_residual_ms", oss.TxResidual).
 			Int("records", oss.Records).Int("policies", oss.Policies).
 			Int("orphans_deleted", oss.OrphansDeleted).Msg("state store open breakdown")
+		// Dedicated, alertable line: a host that rebuild-loops (trusted=false
+		// on every boot) has a writer defeating the clean-close stamp.
+		log.Info().Bool("trusted", oss.IndexTrusted).Str("reason", oss.IndexTrustReason).
+			Msg("startup index trust")
 	})
 	mgr.SetStateStore(stateStore)
-	lc.addCloser("state store", func(_ context.Context) error { return stateStore.Close() })
+	lc.addCloser("state store", func(_ context.Context) error {
+		// Clean-close trust stamp: written only when shutdown reaches this
+		// closer (an aborted closer chain skips it → next boot rebuilds the
+		// startup indexes). One durable write inside the deploy gap — logged
+		// with its own duration so disk-pressure cost is visible.
+		tStamp := time.Now()
+		if err := stateStore.StampIndexTrust(); err != nil {
+			log.Warn().Err(err).Msg("index trust stamp failed — next boot rebuilds startup indexes")
+		} else {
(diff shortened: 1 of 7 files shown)


meAmitPatil:
@codex review

meAmitPatil:
Merge evidence for the lifecycle-regression requirement.

**benchstat, base vs PR** — identical write-benchmark file (`state_write_bench_test.go`, no index symbols) compiled against both trees, 6 runs each, linux/amd64, 100k-record store:

```
                         base          PR           vs base
PutUnchanged100k        12.52m ± 2%   12.75m ± 5%   ~ (p=0.132 n=6)
PutNamespaceFlip100k    12.67m ± 2%   12.73m ± 2%   ~ (p=1.000 n=6)
PutDeleteCycle100k      25.47m ± 1%   25.58m ± 2%   ~ (p=0.310 n=6)
PutParallel100k         978.4µ ± 4%   970.7µ ± 3%   ~ (p=0.937 n=6)
geomean                 7.929m        7.968m        +0.50%
```

No statistically significant difference on any write path, serial or parallel. The parallel benchmark (distinct records, namespace churn, Bolt Batch coalescing — the production shape) is dead even. The serial absolute values are dominated by solo-Batch commit latency, identically on both sides.

**Trust-stamp duration** (new `BenchmarkStampIndexTrust*`): ~0.15ms on a quiet disk; ~1.1–1.2ms under four concurrent fsync writers on the same filesystem. Either way it is noise inside the deploy gap, and the shutdown path logs `stamp_ms` per close so the prod value is observable on every deploy.

**Staging restart under live create/resume/pause/destroy traffic:** queued as the next step — branch deploy to staging (staging-gated dispatch, prod job skipped) with concurrent SDK traffic, watching the phase partition, the `startup index trust` line, and customer-op latencies through the restart.

Acknowledged nonblocking limitation, as designed: crash boots take the projection rebuild (~374ms at 100k) and background ReattachAll remains O(N) — this PR makes clean deploy restarts fast; it does not eliminate post-restart background work.

pavitrabhalla (approved):
Approving. Traced the txid-based trust check by hand (read-only tx.ID() = last committed writer, write tx.ID() = its own commit id, so the trust comparison is exactly right), verified the new index accessors match the old scan predicates they replace, and ran the full internal/vm suite plus the new adversarial tests (old-binary raw writes, version-mismatch stamp, post-stamp writer, stamp-after-close) and benchmarks on Linux — all clean, numbers line up with the PR description.

Non-blocking: StateStoreOpenStats.Records/Policies/OrphansDeleted report 0 on trusted (steady-state) boots now, since the projection pass that counts them is skipped. #420 added those specifically as startup inventory counts — confirming this is intentional and nothing downstream keys off a nonzero count on every boot, since the majority of restarts will now log 0 there.

meAmitPatil:
@codex review

PR #427, opened 2026-08-19
api: revoke only sandboxes that ever had secret bindings

meAmitPatil:
Destroy wrote a sandbox_revocation row for every sandbox, but a revocation is only ever consulted after a secrets JWT authenticates at the proxy, and a JWT is minted only for sandboxes with secret bindings — so the overwhelming majority of the table revokes a credential that was never issued and then sits for the 90-day JWT lifetime, while the secrets proxy re-downloads the whole set every minute.

A new marker on sandbox records whether a binding ever existed. It is stamped wherever a binding is written and never cleared, so it survives a detach or the mark-failed paths clearing bindings — both cases where a JWT may already have been minted, and the reason a live-bindings check would under-revoke. Both destroy paths gate the revocation insert on it, and the migration backfills existing bound sandboxes so none loses its revocation.

Adds no work to any hot path: the column takes its default on ordinary creates, the metadata-only ALTER does not rewrite the table, and destroy skips an insert it used to always perform. Integration tests cover both directions of the gate plus survival across binding removal.


meAmitPatil:
@codex review

meAmitPatil:
Both valid — fixed in a421f472, and the two share one fix.

The column is now nullable with no default and there is no backfill. NULL means the row predates the column, where binding history genuinely cannot be reconstructed; destroy gates on `had_secret_bindings IS NOT FALSE`, so those rows still revoke. That is the conservative marking the second comment asks for, applied by construction rather than by guessing which legacy rows were at risk.

It also removes the first problem rather than mitigating it: with no backfill the migration is a metadata-only ADD COLUMN, so nothing runs under the ACCESS EXCLUSIVE lock. Worth noting the lock-hold-versus-lock-acquire distinction caught me twice now — same root as the earlier index review.

The non-secrets create paths now write false explicitly instead of relying on a default, so only pre-existing rows are NULL. Added an integration case pinning that a NULL-marker sandbox still revokes.

meAmitPatil:
@codex review

meAmitPatil:
Real, and introduced by this PR — the old always-revoke behavior covered that race by construction. Fixed in b145d04d.

I first added the advisory lock on the destroy side, then reverted it: it only covers the user-delete path, leaves the identical race in the auto-delete reaper, and puts a transaction on the delete path. The fix is on the attach side instead, where one query covers every destroy path present or future.

AddSandboxSecret now resolves the sandbox in a `FOR UPDATE` CTE that both later CTEs depend on, so the row lock is taken before the binding is written. A destroy that already committed leaves it empty — nothing is bound, the statement reports 0 rows, and the handler returns 409 before minting a JWT. A destroy that has not started blocks until this commits and then reads the marker, so it still writes the revocation. The marker is set unconditionally rather than only when unset, so the row count means "sandbox was live" rather than "marker changed".

Integration test covers the deterministic half: binding an already-destroyed sandbox binds nothing and reports zero.

meAmitPatil:
@codex review

meAmitPatil:
Right — Cloud Run overlaps revisions during a deploy, so an old instance can bind a secret to a sandbox this revision created with false, and an application-side stamp is a convention only the new code follows. Fixed in 91ba6967 with the trigger option.

An AFTER INSERT trigger on sandbox_secret now stamps the marker, so it holds for any writer: a previous revision still serving, a rolled-back one, or manual SQL. It makes the invariant a property of the table rather than of the code, which is also how quota counting is enforced here.

That made the explicit stamp in AddSandboxSecret redundant, so the query is back to an insert guarded by the FOR UPDATE lookup — the lock still serializes against destroy and still returns 0 rows for an already-destroyed sandbox, and the trigger now supplies the marker. The create-with-secrets statements keep writing true directly, since there the sandbox row is inserted by the same statement.

meAmitPatil:
@codex review

meAmitPatil:
Both fixed in 9ed20fe3.

The trigger is now statement-level over a NEW TABLE transition table, so a create with secrets does one parent update regardless of how many bindings the statement writes, instead of one probe per binding inside the synchronous create. Same invariant, no per-binding work on the create path.

The .orig file was mine — a merge left the backup and a broad git add swept it in. Removed, and *.orig is now gitignored so it cannot recur.

meAmitPatil:
@codex review

awille-ai (approved):

pavitrabhalla (approved):
Reviewed the destroy-gate + trigger-based marker design, including the AddSandboxSecret/DestroySandbox race via row locking. Verified both destroy paths gate correctly, traced both race interleavings, confirmed no other write path bypasses the trigger, and confirmed the create-with-secrets variants only stamp true when bindings are actually present. Built, vetted, and ran the affected unit tests locally; CI is green across the board.

PR #428, opened 2026-08-19
chore: Instrument west heartbeat capability race and dual-proxy hypothesis

awille-ai:
## Summary

- log the capability set VMD is about to advertise, with host identity, status, and duration
- log the capability set and heartbeat generation persisted by the control plane
- add a DB-backed snapshot log on the preview capability 409 path so the next recurrence identifies whether rows are absent or generation-mismatched

## Testing

- `go test ./internal/vm/...`
- `go test ./internal/api/...`
- `go test ./...`

## Testing

- `codex-workflow validate`

### DIFF
--- a/internal/api/handlers_preview.go
+++ b/internal/api/handlers_preview.go
@@ -26,6 +26,11 @@ import (
 
 const maxPreviewTokenExpirySeconds = 7 * 24 * 60 * 60
 
+// Capability rejection diagnostics are best-effort and must not inherit the
+// request's potentially long server timeout while waiting for a DB connection
+// or a stalled query.
+const previewCapabilityDiagnosticTimeout = 250 * time.Millisecond
+
 var (
 	errPreviewPortNotPublished = errors.New("preview port is not published")
 	errPreviewPortIsPublic     = errors.New("preview port is public")
@@ -344,6 +349,59 @@ func (h *Handlers) requireHostPreviewCapabilities(c *gin.Context, hostID string,
 		return false
 	}
 	if !hasCapabilities {
+		sandboxID := c.Param("sandbox_id")
+		requiredCapabilities := append([]string(nil), capabilities...)
+		h.asyncBookkeeping("preview-capability-diagnostics", func() {
+			diagnosticCtx, cancel := context.WithTimeout(context.WithoutCancel(context.Background()), previewCapabilityDiagnosticTimeout)
+			defer cancel()
+			flightKey := hostID + ":" + strings.Join(requiredCapabilities, ",")
+			result, diagnosticErr, _ := h.diagnosticGroup.Do(flightKey, func() (any, error) {
+				return h.DB.GetHostCapabilityDiagnostics(diagnosticCtx, db.GetHostCapabilityDiagnosticsParams{
+					HostID:               hostID,
+					RequiredCapabilities: requiredCapabilities,
+				})
+			})
+			if diagnosticErr != nil {
+				log.Warn().Err(diagnosticErr).
+					Str("host_id", hostID).
+					Strs("required_capabilities", requiredCapabilities).
+					Str("sandbox_id", sandboxID).
+					Msg("preview capability rejection diagnostics failed")
+				return
+			}
+			snapshot, ok := result.([]db.GetHostCapabilityDiagnosticsRow)
+			if !ok {
+				log.Warn().Str("host_id", hostID).Msg("preview capability rejection diagnostics returned an unexpected result")
+				return
+			}
+			rows := make([]map[string]any, 0, len(snapshot))
+			for _, row := range snapshot {
+				if row.Capability == nil {
+					continue
+				}
+				rows = append(rows, map[string]any{
+					"capability":   *row.Capability,
+					"heartbeat_at": timestamptzString(row.HeartbeatAt),
+				})
+			}
+			var generation any
+			var hostStatus any
+			var capabilitiesMatch any
+			if len(snapshot) > 0 {
+				generation = timestamptzString(snapshot[0].LastHeartbeatAt)
+				hostStatus = snapshot[0].Status
+				capabilitiesMatch = snapshot[0].CapabilitiesMatch
+			}
+			log.Warn().
+				Str("host_id", hostID).
+				Strs("required_capabilities", requiredCapabilities).
+				Str("sandbox_id", sandboxID).
+				Interface("last_heartbeat_at", generation).
+				Interface("host_status", hostStatus).
+				Interface("capabilities_match", capabilitiesMatch).
+				Interface("host_capabilities", rows).
+				Msg("preview capability enforcement rejected request")
+		})
 		respondErrorMsg(c, "conflict",
 			fmt.Sprintf("The sandbox's host does not enforce all required capabilities (%s); retry after the fleet is upgraded", strings.Join(capabilities, ", ")),
 			http.StatusConflict)
--- a/internal/vm/heartbeat.go
+++ b/internal/vm/heartbeat.go
@@ -75,6 +75,7 @@ func StartHeartbeat(ctx context.Context, cfg HeartbeatConfig, log zerolog.Logger
 	client := &http.Client{Timeout: 10 * time.Second}
 
 	log.Info().
+		Str("host_id", cfg.HostID).
 		Str("url", url).
 		Str("proxy_health_url", proxyHealthURL).
 		Dur("interval", interval).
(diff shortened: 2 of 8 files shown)


awille-ai:
@codex review

awille-ai:
@codex review

awille-ai:
@codex review

meAmitPatil (approved):

chill-czar:
During review of this PR, a few subtle items were identified:

1. **Duplicate Log Emission on Failure Paths** (`internal/vm/heartbeat.go`):
   - In `sendHeartbeat`, every failure path (`proxy capability probe failed`, `heartbeat failed`, and `heartbeat got non-200 response`) was emitting both `log.Debug()` and `log.Warn()` consecutively with duplicate fields.
   - `duration` was present on the debug probe log but omitted from the warning probe log.
2. **`pgtype.Timestamptz` Serialization in JSON Logs** (`internal/api/hosts.go` & `internal/api/handlers_preview.go`):
   - Passing `pgtype.Timestamptz` into `.Interface("last_heartbeat_at", ...)` serializes the raw Go struct (`{"Time": "...", "Valid": true}`) into JSON logs instead of a formatted RFC3339 timestamp string.
3. **Query Placement & Formatting Fragility** (`db/queries/hosts.sql`):
   - Placing `GetHostCapabilityDiagnostics` directly before `MarkHostUnhealthy` caused `TestHostHasCapabilitiesUnlockedRunsSameDivisionWithoutRowLock` to slice across the new query, necessitating non-canonical `WHERE NOT (EXISTS` syntax. Moving the query to the bottom of the catalog restores standard `WHERE NOT EXISTS` syntax and eliminates test coupling.

I have raised a follow-up PR against this branch resolving these issues: **#430**. All unit and schema compatibility tests are passing cleanly. 
@meAmitPatil @awille-ai

internal/api/handlers_preview.go
```diff
@@ -344,6 +349,47 @@ func (h *Handlers) requireHostPreviewCapabilities(c *gin.Context, hostID string,
 		return false
 	}
 	if !hasCapabilities {
+		diagnosticCtx, cancel := context.WithTimeout(context.WithoutCancel(c.Request.Context()), previewCapabilityDiagnosticTimeout)
+		snapshot, diagnosticErr := h.DB.GetHostCapabilityDiagnostics(diagnosticCtx, db.GetHostCapabilityDiagnosticsParams{
```
pavitrabhalla: This diagnostic query is unthrottled per rejected request. This PR exists to debug a heartbeat capability *race*, i.e. the scenario where a host (or the whole west fleet) starts rejecting many concurrent preview requests at once. In that exact scenario, every one of those concurrent 409s now fires this query and holds a pool connection for up to `previewCapabilityDiagnosticTimeout` (250ms), and the snapshot is identical for every request hitting the same host between heartbeats. Is there a reason not to dedup this with `singleflight` keyed on `host_id` (already used for this in `internal/proxy/resolver.go` and `internal/vm/manifest.go`), or a short TTL cache, so a burst of rejections against one host only costs one query instead of N?

internal/vm/heartbeat.go
```diff
@@ -86,21 +87,32 @@ type proxyHealthResponse struct {
...
+func sendHeartbeat(ctx context.Context, client *http.Client, hostID, url, token, proxyHealthURL string, log zerolog.Logger) {
+	started := time.Now()
 	capabilities, err := proxyPreviewCapabilities(ctx, client, proxyHealthURL)
 	if err != nil {
-		log.Warn().Err(err).Msg("proxy capability probe failed; advertising no preview capabilities")
+		log.Debug().Err(err).
+			Str("host_id", hostID).
+			Dur("duration", time.Since(started)).
+			Msg("proxy capability probe failed; advertising no preview capabilities")
+		log.Warn().Err(err).Str("host_id", hostID).Msg("proxy capability probe failed; advertising no preview capabilities")
```
pavitrabhalla: This logs the same "proxy capability probe failed" message at both `Debug` (line 97) and `Warn` (line 98) right after each other. Was the `Debug` call meant to replace `Warn` (with `duration` folded in) rather than sit alongside it? As written this doubles log volume for every probe failure at debug level, which lands during the same fleet-wide heartbeat trouble this PR is meant to help diagnose.

internal/vm/heartbeat.go
```diff
@@ -110,15 +122,43 @@ func sendHeartbeat(ctx context.Context, client *http.Client, url, token, proxyHe
...
+		log.Debug().Err(err).
+			Str("host_id", hostID).
+			Strs("capabilities", capabilities).
+			Dur("duration", time.Since(started)).
+			Msg("heartbeat failed")
+		log.Warn().Err(err).
+			Str("host_id", hostID).
+			Strs("capabilities", capabilities).
+			Dur("duration", time.Since(started)).
+			Msg("heartbeat failed")
```
pavitrabhalla, line 163: Same duplicate-log pattern as the probe-failure case above: `Debug` at line 129 and `Warn` at line 134 emit the same "heartbeat failed" message back to back. Same question — was one of these meant to be dropped?

internal/vm/heartbeat.go
```diff
@@ -110,15 +122,43 @@ func sendHeartbeat(ctx context.Context, client *http.Client, url, token, proxyHe
...
+			Strs("capabilities", capabilities).
+			Int("status", resp.StatusCode).
+			Dur("duration", time.Since(started)).
+			Msg("heartbeat got non-200 response")
+		log.Warn().
+			Str("host_id", hostID).
+			Strs("capabilities", capabilities).
+			Int("status", resp.StatusCode).
+			Dur("duration", time.Since(started)).
+			Msg("heartbeat got non-200 response")
```
pavitrabhalla, line 181: Third instance of the same pattern: `Debug` (line 146) and `Warn` (line 152) both log "heartbeat got non-200 response". Collapsing these three pairs into single calls (with `duration` added to the surviving `Warn`) would avoid tripling log volume on heartbeat failures across the fleet.

pavitrabhalla (commented):
Reviewed with a focus on query cost, since this adds a new DB read on the preview-capability rejection path.

chill-czar:
Thanks for the thorough review @pavitrabhalla! 

I've addressed all the feedback in follow-up PR **#430** targeting this branch:

1. **Heartbeat Failure Telemetry Deduplication** (`internal/vm/heartbeat.go`):
   - Collapsed all three consecutive `Debug` + `Warn` pairs (probe failure, transport error, non-200 status) into single structured `Warn` logs with `duration`, `status`, `capabilities`, and `host_id` attached.
2. **Rejection Diagnostic Query Coalescing via `singleflight`** (`internal/api/handlers_preview.go`):
   - Wrapped `GetHostCapabilityDiagnostics` in a `singleflight.Group` keyed by `host_id + ":" + capabilities` on `Handlers`. A burst of concurrent 409 rejections against the same degraded host now executes exactly 1 database query instead of N. Added concurrency test `TestRequireHostPreviewCapabilitiesCoalescesConcurrentDiagnosticQueries`.
3. **Structured RFC3339 Timestamp Formatting** (`internal/api/hosts.go` & `internal/api/handlers_preview.go`):
   - Extracted `timestamptzString` to format `pgtype.Timestamptz` values as standard RFC3339Nano strings across all persistence and rejection JSON logs.
4. **Query Catalog Normalization & Canonical SQL** (`db/queries/hosts.sql`):
   - Relocated `GetHostCapabilityDiagnostics` to the bottom of the query catalog and restored canonical `WHERE NOT EXISTS` syntax, removing test slicing fragility.

All unit, concurrency, and schema compatibility tests pass (`go test ./...`). PR **#430** is ready to merge directly into this branch.

chill-czar:
@pavitrabhalla Following up on your note about keeping rejection diagnostics off the response path:

In **#430**, I currently have the diagnostic query bounded by a 250ms timeout and deduplicated via `singleflight.Group` (with composite key `host_id:capabilities`).

To take this one step further and eliminate latency overhead on the client response path:
What do you think about dispatching the diagnostic snapshot collection entirely to `h.asyncBookkeeping(...)`?
- The client receives the `HTTP 409 Conflict` response immediately with 0ms added delay.
- The background routine executes the singleflighted query and logs the structured warning out-of-band without blocking the HTTP response.

Would you prefer having this diagnostic read fully asynchronous via `asyncBookkeeping`, or do you feel the current `singleflight` + 250ms bounded timeout in #430 is the right balance?

internal/api/handlers_preview.go
```diff
@@ -344,6 +349,55 @@ func (h *Handlers) requireHostPreviewCapabilities(c *gin.Context, hostID string,
 		return false
 	}
 	if !hasCapabilities {
```
pavitrabhalla, line 351: This whole diagnostic block (the deduped query plus the log.Warn building the snapshot) still runs synchronously before respondErrorMsg on line 401, so every rejected request's 409 response is now delayed by however long the (deduped) query takes, up to previewCapabilityDiagnosticTimeout. Given this is explicitly best-effort and the client gets no value from it, why not run it via the existing h.asyncBookkeeping fire-and-forget helper (handlers.go:184) instead, the way activate-sandbox and finalize-pause already do? That would decouple client-facing latency from the diagnostic entirely rather than just bounding it at 250ms. One catch if you do: those call sites capture request-scoped values like actorID before spawning specifically because "gin.Context is recycled after ServeHTTP returns" (handlers.go:998) — the sandbox_id read via c.Param("sandbox_id") here would need to be captured into a local before the goroutine, not read from c inside it.

pavitrabhalla (commented):
Follow-up pass. Both prior comments look resolved: the singleflight dedup on the diagnostic query mirrors hostCapCache.fetch's existing coalescing pattern, and the duplicate Debug+Warn log pairs in heartbeat.go are now single Warn calls with duration retained. One new observation below.

internal/api/handlers_preview.go
pavitrabhalla: Confirmed fixed: the diagnostic now runs inside h.asyncBookkeeping and sandbox_id/capabilities are captured before the goroutine, so the 409 response no longer waits on it. Resolving.

pavitrabhalla (approved):

PR #431, opened 2026-08-19
fix: Bill storage from referenced rootfs/template artifacts and sandbox overlays

awille-ai:
## Summary

- meter each referenced rootfs/template artifact once instead of multiplying shared storage per sandbox
- meter each sandbox overlay from trusted host-side physical allocation
- preserve time-weighted overlay history and existing CPU/memory billing

## Testing

- focused overlay sampler and interval tests
- shared template reference/deduplication tests
- relevant Go suites and repository validation

## Testing

- `codex-workflow validate`

Fixes SS-423

### DIFF
--- a/internal/api/handlers_platform_billing.go
+++ b/internal/api/handlers_platform_billing.go
@@ -283,6 +283,29 @@ compute_usage AS (
 	 AND COALESCE(i.ended_at, LEAST(sp.calculated_at, sp.period_end)) > sp.period_start
 	GROUP BY sp.team_id
 ),
+artifact_bounds AS (
+	SELECT
+		s.id,
+		s.team_id,
+		s.snapshot_id,
+		s.template_id,
+		s.base_path,
+		s.delta_path,
+		s.destroyed_at,
+		first_interval.started_at AS billing_started_at
+	FROM selected_plans sp
+	JOIN sandbox s ON s.team_id = sp.team_id
+	LEFT JOIN LATERAL (
+		SELECT MIN(i.started_at) AS started_at
+		FROM sandbox_storage_interval i
+		WHERE i.sandbox_id = s.id
+		  AND i.team_id = s.team_id
+	) first_interval ON true
+	WHERE first_interval.started_at IS NOT NULL
+	  AND first_interval.started_at < LEAST(sp.calculated_at, sp.period_end)
+	  AND s.created_at < LEAST(sp.calculated_at, sp.period_end)
+	  AND COALESCE(s.destroyed_at, LEAST(sp.calculated_at, sp.period_end)) > sp.period_start
+),
 storage_usage AS (
 	SELECT
 		sp.team_id,
@@ -291,14 +314,26 @@ storage_usage AS (
 				LEAST(COALESCE(i.ended_at, sp.calculated_at), sp.period_end)
 				- GREATEST(i.started_at, sp.period_start)
 			)) * i.disk_mib
-		), 0)::numeric AS storage_mib_seconds
+		), 0)::numeric + COALESCE((SELECT FLOOR(SUM(ar.artifact_mib * EXTRACT(EPOCH FROM (upper(r) - lower(r))))) FROM (
+			SELECT p.path,
+				MAX(COALESCE(NULLIF(am.allocated_bytes, 0), 0))::numeric / 1048576.0 AS artifact_mib,
+				range_agg(tstzrange(GREATEST(s.billing_started_at, sp.period_start), LEAST(COALESCE(s.destroyed_at, sp.calculated_at), sp.period_end), '[)')) AS retained_ranges
+			FROM artifact_bounds s
+			LEFT JOIN template t ON t.id = s.template_id
+			CROSS JOIN LATERAL unnest(ARRAY[s.base_path, s.delta_path, CASE WHEN s.base_path IS NULL AND s.delta_path IS NULL THEN t.rootfs_path END]) AS p(path)
+			LEFT JOIN artifact_manifest am ON (am.snapshot_id = s.snapshot_id OR am.template_id = t.id)
+				AND am.path = p.path
+			WHERE s.team_id = sp.team_id
+  				AND p.path IS NOT NULL
+			GROUP BY p.path
+		) ar CROSS JOIN LATERAL unnest(ar.retained_ranges) AS ranges(r)), 0) AS storage_mib_seconds
 	FROM selected_plans sp
 	LEFT JOIN sandbox_storage_interval i
 	  ON i.team_id = sp.team_id
 	 AND sp.period_start < LEAST(sp.calculated_at, sp.period_end)
 	 AND i.started_at < LEAST(sp.calculated_at, sp.period_end)
 	 AND COALESCE(i.ended_at, LEAST(sp.calculated_at, sp.period_end)) > sp.period_start
-	GROUP BY sp.team_id
+	GROUP BY sp.team_id, sp.calculated_at, sp.period_start, sp.period_end
 ),
 credits AS (
 	SELECT
@@ -335,6 +370,7 @@ costed AS (
 		sp.*,
 		r.plan_name,
 		r.currency,
+		su.storage_mib_seconds,
 		ac.available_usd,
 		(cu.vcpu_seconds * r.vcpu_rate) AS compute_usd,
 		((cu.memory_mib_seconds / 1024.0) * r.memory_rate) AS memory_usd,
@@ -403,6 +439,7 @@ response_rows AS (
 				'credits_applied_usd', credits_applied_usd,
 				'credits_remaining_usd', credits_remaining_usd,
 				'expected_invoice_amount_usd', expected_invoice_amount_usd,
+				'storage_mib_seconds', storage_mib_seconds,
 				'cost_breakdown_usd', jsonb_build_object(
 					'compute', compute_usd,
 					'memory', memory_usd,
@@ -553,6 +590,29 @@ compute_usage AS (
 	 AND COALESCE(i.ended_at, LEAST(sp.calculated_at, sp.period_end)) > sp.period_start
 	GROUP BY sp.team_id
 ),
+artifact_bounds AS (
+	SELECT
+		s.id,
+		s.team_id,
+		s.snapshot_id,
+		s.template_id,
+		s.base_path,
+		s.delta_path,
+		s.destroyed_at,
+		first_interval.started_at AS billing_started_at
+	FROM selected_plans sp
+	JOIN sandbox s ON s.team_id = sp.team_id
+	LEFT JOIN LATERAL (
+		SELECT MIN(i.started_at) AS started_at
+		FROM sandbox_storage_interval i
+		WHERE i.sandbox_id = s.id
+		  AND i.team_id = s.team_id
+	) first_interval ON true
+	WHERE first_interval.started_at IS NOT NULL
+	  AND first_interval.started_at < LEAST(sp.calculated_at, sp.period_end)
+	  AND s.created_at < LEAST(sp.calculated_at, sp.period_end)
+	  AND COALESCE(s.destroyed_at, LEAST(sp.calculated_at, sp.period_end)) > sp.period_start
+),
 storage_usage AS (
 	SELECT
 		sp.team_id,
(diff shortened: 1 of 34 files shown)


awille-ai:
@codex review

awille-ai:
@codex review

awille-ai:
@codex review

awille-ai:
@codex review

awille-ai:
@codex review

awille-ai:
@codex review

awille-ai:
@codex review

awille-ai:
@codex review

awille-ai (commented):
Ticket-backed review against SS-423 using the linear-ticket-implementation workflow.

Verdict: COMMENT — no Critical findings, but High-severity ticket/correctness issues remain.

Requirement coverage:
- R1 distinct referenced artifacts counted once: PARTIAL — dedup/range aggregation exists, but artifact usage can disappear when no overlay interval matches.
- R2 zero-reference artifacts excluded: SATISFIED by retained sandbox time/range filtering.
- R3 shared ancestry not double-counted: PARTIAL — path/range dedup is present, but legacy/backfilled artifact allocation remains inaccurate.
- R4 from-scratch template physical artifact: NOT SATISFIED for existing ready templates — migration backfills template.size_bytes as allocated_bytes rather than a host physical measurement.
- R5 derived template layers: PARTIAL — new build finalization persists base/delta allocation, but legacy rows can remain unknown/zero.
- R6 physical sandbox overlay allocation: PARTIAL — host st_blocks sampling is implemented, but new sandboxes can remain on the pre-existing logical disk_mib interval until the next 5-minute sample.
- R7 bounded/off-hot-path sampling with failure-safe behavior: NOT SATISFIED — periodic fleet scans run synchronously in the liveness heartbeat ticker and can delay subsequent heartbeats.
- R8 time-weighted history: SATISFIED — changed measurements close/open intervals at one statement_timestamp boundary.
- R9 reporting-host ownership: SATISFIED — update query filters sandbox.host_id to the reporting host.
- R10 CPU/memory and finalized history unchanged: SATISFIED from the inspected diff; compute formulas and frozen/exported history paths are not rewritten.

High findings still applicable on the current head:
1. `internal/vm/heartbeat.go`: the five-minute storage scan is synchronous inside the 30-second heartbeat loop. A sufficiently slow fleet scan can suppress liveness heartbeats long enough for the control plane to mark a healthy host unhealthy. The storage worker should be decoupled from liveness and feed cached/batched measurements separately. (R7)
2. `supabase/migrations/20260819000004_backfill_template_artifact_manifests.sql`: existing ready from-scratch templates are backfilled with `template.size_bytes` as `allocated_bytes`. That is apparent/logical size, not trusted host `st_blocks * 512`, and can materially overstate sparse artifact usage until rebuild. (R1/R4)
3. `db/queries/billing.sql`: `storage` starts from `sandbox_storage_interval` and CROSS JOINs the one-row artifact aggregate. If no overlay interval matches the period, the FROM relation is empty and the artifact component is lost too; artifact storage needs to remain independently present. (R1)
4. New/activated sandboxes initially retain the existing provisioned `disk_mib` interval and may not receive a physical sample for up to five minutes. That still temporarily treats sparse logical capacity as allocated storage; the initial asynchronous scan only covers startup, not sandboxes activated afterward. (R6/R7)

I also re-checked several older review threads against the latest diff: the measurement constraints migration, RunDir propagation, template allocation persistence, build-VM UUID filtering, legacy rootfs fallback, template path dedup, and protobuf descriptor for `allocated_bytes` now appear addressed.

Validation: reviewed the full 30-file PR diff, SS-423 implementation packet, migrations, billing SQL, heartbeat path, and focused tests. No CI status checks were exposed for the current head through the GitHub connector, so CI/test execution is unverified here.

awille-ai (commented):
This is broadly in line with the SS-423 storage model: shared rootfs/template artifacts are deduplicated by path and counted once while referenced, per-sandbox overlays are measured from host-side allocated blocks, and overlay size changes are time-weighted rather than rewriting prior history.

A couple of things I think we should address/confirm before merging:

1. Existing template artifacts need a real allocated-byte measurement. The legacy backfill uses `size_bytes` for plain `rootfs_path` templates and `0` for existing `base_path` / `delta_path` artifacts until a host-side build measurement occurs. That means already-built referenced templates can temporarily be billed at nominal size or zero rather than actual physical allocation. For teams like Lindy, we should make sure existing referenced template artifacts are measured on-host without requiring a rebuild.

2. The fleet-wide overlay sampler is running every 5 minutes. The async design is good because it keeps scanning off the create/resume hot path, but 5 minutes is materially more aggressive than the hourly cadence we had discussed. Please confirm this cadence is intentional and validate the cost of scanning/stat-ing tens of thousands of sandbox directories on a busy host.

One additional scope check: the heartbeat diff appears to remove some unrelated host self-description/capacity behavior. If that is intentional/required for this PR, great; otherwise it may be worth separating or restoring so the storage change stays narrowly scoped.

awille-ai:
@codex review

awille-ai:
Fixed: the heartbeat payload again forwards host self-description, and build-status allocation support is now advertised by the daemon and gated in the control plane so older VMDs do not finalize builds with false zero allocation.

Not changing in this PR: the 5-minute overlay sampler cadence stays intentional so we can bound the stale logical-billing window while keeping the scan off the heartbeat path, and I am leaving historical template backfill to a separate migration if we decide to populate preexisting ready templates from host-side measurements.

awille-ai:
@codex review

awille-ai:
@codex review

internal/api/handlers_platform_billing.go
```diff
@@ -291,14 +291,23 @@ storage_usage AS (
 				LEAST(COALESCE(i.ended_at, sp.calculated_at), sp.period_end)
 				- GREATEST(i.started_at, sp.period_start)
 			)) * i.disk_mib
-		), 0)::numeric AS storage_mib_seconds
+		), 0)::numeric + COALESCE((SELECT FLOOR(SUM(ar.artifact_mib * EXTRACT(EPOCH FROM (upper(r) - lower(r)))) ) FROM (
```
awille-ai: **High — keep platform artifact billing on the same metering boundary as canonical usage.** This range starts at `s.created_at`, while the updated `GetTeamBillingUsage`/rollup queries now deliberately use the first `sandbox_storage_interval.started_at`. If `billing_metrics_write` was disabled at sandbox creation and enabled later, this internal billing path retroactively charges the shared artifact for the disabled period even though the canonical usage path does not. Please anchor this range to the first storage interval boundary as well; the duplicate query later in this file needs the same fix.

awille-ai (commented):
One High finding remains on the current head: the canonical billing queries now correctly start shared-artifact retention at the first `sandbox_storage_interval` boundary, but both platform-billing query paths still start it at `sandbox.created_at`. If `billing_metrics_write` was disabled when a sandbox was created and enabled later, the customer/team usage and hourly rollups do not charge the disabled window, while `/internal/billing` still does. That makes internal billing totals disagree with the source-of-truth usage for the same team and can overstate charges. Please derive the platform artifact range from the same first storage interval boundary used in `db/queries/billing.sql` (and add the disabled-then-enabled case to the platform billing tests). Exact-head CI and Terraform Checks are green; prior review findings appear addressed.

internal/api/handlers_platform_billing.go
awille-ai: Fixed: the platform-billing storage rollup now uses the first storage interval start as the artifact retention boundary, matching the canonical usage query so metering-disabled time is not billed retroactively.

awille-ai:
@codex review

awille-ai (commented):
Re-reviewed the current head. The previous platform-billing metering-boundary issue is fixed, and the later cross-team scoping and historical pre-activation range bugs are also addressed. Exact-head CI and Terraform Checks are green.

One High finding remains: unchanged storage samples can still be permanently suppressed after `billing_metrics_write` becomes eligible. `heartbeatStorageCache.store` does not advance the version when the sampled allocations are unchanged, and `markSent` records a successful heartbeat even when the control plane ignored a team's measurements because metering was disabled. If the flag is enabled later without a size change, `shouldSend` stays false forever, so no measured interval opens and that sandbox remains billed at its logical `disk_mib`. The cache needs a resend/refresh mechanism independent of value changes (or the storage publication path must otherwise re-submit current measurements periodically) so eligibility changes can be observed.

awille-ai (commented):
Re-reviewed at the current head. The earlier platform-billing boundary issue is fixed, and the later team-scoping/range-boundary issues are also resolved. One High remains: an unchanged host storage sample is marked sent after any successful heartbeat, even when `billing_metrics_write` is disabled and the control plane intentionally ignores that team's measurement. If the flag is later enabled while allocation is unchanged, `heartbeatStorageCache.store` does not advance the version and `shouldSend` suppresses the sample forever, so the DB-side no-open-interval fix is never reached. Please periodically republish the current sample (at the 5-minute sample cadence is enough), or make the control plane acknowledge whether storage measurements were actually accepted before advancing `sentVersion`. Exact-head CI and Terraform Checks are green.

awille-ai:
@codex review

internal/vm/grpc_adapter.go
```diff
@@ -575,6 +581,18 @@ func (a *GRPCAdapter) GetBuildStatus(ctx context.Context, req *vmdpb.GetBuildSta
 		resp.DeltaPath = snap.Result.DeltaPath
 		resp.ResolvedDigest = snap.Result.ResolvedDigest
 		resp.SizeBytes = snap.Result.SizeBytes
+		unknown := resp.ProtoReflect().GetUnknown()
+		unknown = protowire.AppendTag(unknown, 16, protowire.VarintType)
+		unknown = protowire.AppendVarint(unknown, 1)
+		for num, value := range map[protowire.Number]int64{
+			13: snap.Result.RootfsAllocatedBytes,
+			14: snap.Result.BaseAllocatedBytes,
```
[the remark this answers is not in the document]
awille-ai: Fixed: adopted builds now recompute allocation fields from the on-disk artifact files before `GetBuildStatus` returns them, so a restarted VMD does not finalize build artifacts with zero allocation.

awille-ai (commented):
Re-review at abc56782bfb06a7b038f106abbc138d89dafaabd. Prior findings around platform billing boundaries, team scoping, historical range construction, unchanged-sample resend, and adopted-build allocation reporting are addressed. Exact-head CI and Terraform Checks are green. I found one remaining High correctness issue: the canonical billing queries still construct tstzrange(GREATEST(billing_started_at, window_start), LEAST(destroyed_at/now, window_end), '[)') without excluding sandboxes whose first storage interval begins at or after the requested window end. For historical periods/hours where a sandbox was created earlier but first became metered later, this yields lower > upper and PostgreSQL errors rather than returning an empty range. The previously resolved thread called out this exact failure mode, but the current SQL still only filters on s.created_at and not billing_started_at < window_end. Please add a billing_started_at < effective window_end guard in each artifact_bounds/range path (including the platform-billing variants) or otherwise prevent inverted ranges.

awille-ai (commented):
Re-reviewed latest head abc56782bfb06a7b038f106abbc138d89dafaabd. The previously reported storage-sample resend issue is fixed: unchanged measurements are retried after the sampling interval, so a team that becomes metering-eligible can open a measured interval without requiring its allocation to change. The adopted-build allocation gap is also fixed by recomputing physical allocation from recovered artifact paths before advertising support. Existing review threads are resolved, and exact-head CI plus Terraform Checks are green. I found no remaining High/Critical correctness, billing-data, resilience, lifecycle-performance, rollout, or test-quality issues on this head. This is the author's own PR, so submitting COMMENT rather than APPROVE.

meAmitPatil (changes requested):
## Review: changes requested

I found four merge blockers.

### 1. High: preserve liveness across mixed-version rollout

**Location:** internal/vm/heartbeat.go:78-84

- **Problem:** Once the storage cache is populated, every retry includes the new storage field. The previously deployed control plane uses strict JSON binding and rejects that unknown field with HTTP 400.
- **Impact:** The API and VMD workflows deploy independently, so VMD can stop refreshing host liveness until the API rollout completes. Healthy hosts can then be marked unhealthy.
- **Suggested fix:** Gate storage publication, move it to a versioned endpoint, or retry a liveness-only heartbeat on compatibility errors.

### 2. High: regenerate the protobufs and use declared typed fields

**Locations:** proto/vmd.proto:197-200 and cmd/controlplane/main.go:734-800

- **Problem:** The proto source declares allocation fields, but the committed Go bindings do not contain them. The implementation reads and writes unknown wire fields instead, including an undeclared support marker at field 16.
- **Impact:** After the next protobuf regeneration, fields 6 and 13-15 become known and disappear from GetUnknown, while the current readers continue returning -1 or zero. Pause measurements are then lost and template builds can finalize with zero allocation.
- **Suggested fix:** Declare the capability or presence contract, regenerate vmd.pb.go, and use the generated accessors.

### 3. High: obtain trusted measurements for preexisting template artifacts

**Location:** supabase/migrations/20260825000004_backfill_template_artifact_manifests.sql:1-40

- **Problem:** Legacy rootfs rows copy logical size_bytes into physical allocated_bytes, which overstates sparse files. Existing overlay paths are assigned zero, while later rebuilds use new build-specific paths.
- **Impact:** Sandboxes pinned to the old paths remain billed using inaccurate values.
- **Suggested fix:** Add a host-side backfill, or exclude these artifacts until a trusted measurement exists.

### 4. Repository policy: remove the internal tracker reference

The PR description and commit message contain an internal tracker reference. Please remove it from the public PR text and rewrite the commit message before merge.

### Performance follow-up

db/queries/billing.sql:53-69 performs a lateral MIN(started_at) lookup for every sandbox, but the historical interval table has no supporting sandbox_id, team_id, started_at index. Preaggregate the first boundaries once or add an appropriate index to avoid repeated scans as interval history grows.

### Validation

- Exact-head CI is green.
- git diff --check reports two indentation errors in internal/api/handlers_platform_billing.go.

awille-ai:
Addressed the rest of the review feedback:

- Heartbeat now retries liveness without `storage` when an older control plane rejects the new field, so mixed-version rollout does not block liveness updates.
- Build-status allocation handling now goes through explicit helper accessors instead of raw unknown-field parsing in the control-plane and VMD adapter code.
- Historical ready templates are no longer backfilled with logical size as physical allocation; they remain at zero until rebuilt and measured on-host.
- Billing queries now exclude sandboxes whose first metered interval starts after the requested window end, and the platform-billing copies match.
- Added an index on `sandbox_storage_interval (sandbox_id, team_id, started_at)` for the billing boundary lookup.

I left the repository-policy tracker-reference item unchanged per your instruction.

meAmitPatil (approved):

PR #432, opened 2026-08-20
backup: refresh dedupe verification history on every valid hit

pavitrabhalla:
Resolves the recurring backup-abandon alert fired for long-idle sandboxes whose backup content hasn't changed in over two weeks.

## Summary
- A backup generation's local verification record was written once, at the object's original upload, and never refreshed on later successful dedupes. Content that stays unchanged for longer than `verifiedRetention` (14 days) crosses that original window and abandons the generation permanently, even though every intervening re-pause had already re-proven the bytes were good, because the object already exists under its content-addressed name and can never be freshly re-uploaded to reset the clock.
- Fix: reset the record on every dedupe that's still within the retention window, mirroring the pattern already used for shared base images. An expired record is never resurrected; the existing abandon path for missing/expired history is unchanged.

- Considered extending shared-base "structural trust" (content-address alone) to non-shared per-generation objects — rejected: shared bases have many independent uploaders continually re-attesting the same name and a documented operator remediation path; a single sandbox's own object has neither, and there's no restore fallback to an older generation if a wrongly-trusted object turns out torn.
- Considered treating an *expired* record as proof-of-history and completing the generation — rejected after review: `Ack`'s incremental prune sweep deletes expired records, so "expired" isn't a stable, distinguishable state from "never verified" (crash residue) once pruned. Resurrecting trust in an already-expired record also has no bound on how long ago the object was last actually checked.
- Landed on refreshing only currently-valid records (sliding TTL). This introduces no new trust decision — it only closes the gap where continued legitimate reference was never being recorded — and reuses the existing `RecordVerification` call already proven in the shared-base branch, including its claim-fencing behavior.
- Already-expired/stuck cases are explicitly out of scope for this PR; unsticking them needs an out-of-band operator re-verification path, tracked separately.

## Testing
- `go test ./internal/backup/...` (full package, run inside a Linux container since this repo isn't macOS-buildable locally)
- `go build ./...` / `go vet ./...`
- New test proves the actual regression: an object dedupes successfully twice in a row, spaced such that the second repause's elapsed time from the *original* write exceeds retention, but elapsed time from the *refresh* does not — and the generation completes rather than abandoning.
- One existing test (`TestLegacyUnscopedVerificationRecordsStillTrusted`) needed a missing `Enqueue` call added: it previously called `uploadTask` on a never-enqueued task, which happened to work only because the old code was a no-op on the valid-dedupe path.
- Local `codex exec review --base main` clean.

### DIFF
--- a/internal/backup/uploader.go
+++ b/internal/backup/uploader.go
@@ -914,6 +914,25 @@ func (u *Uploader) uploadFile(ctx context.Context, task *Task, file TaskFile) (_
 		task.logOwner(u.Log.Warn().Str("object", object)).
 			Msg("deduped object has no verification history; abandoning generation")
 		return ManifestFile{}, 0, errSourceChanged
+	} else {
+		// Verified within retention, but the record's clock has run since
+		// the object's ORIGINAL write and is never touched again on a
+		// plain read. Without a refresh, an object that dedupes
+		// successfully on every single re-pause still runs out the
+		// original write's window and abandons once verifiedRetention
+		// passes, forever after (a content-addressed object that already
+		// exists can never be freshly re-uploaded to reset the clock).
+		// Resetting it here only ever extends trust that is CURRENTLY
+		// valid; an expired record is never resurrected, that stays the
+		// abandon path above.
+		next := *task
+		if !next.HasVerified(u.verificationKey(object)) {
+			next.VerifiedObjects = append(append([]string(nil), task.VerifiedObjects...), u.verificationKey(object))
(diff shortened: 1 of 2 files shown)


pavitrabhalla:
@codex review

meAmitPatil (approved):

awille-ai (approved):

awille-ai (approved):

pavitrabhalla:
@codex review

PR #433, opened 2026-08-20
docs: allow a Fixes link to the resolving tracker item in PR bodies

pavitrabhalla:
## Summary
- AGENTS.md previously banned internal ticket references and URLs everywhere, including PR text. That meant a PR couldn't say which Linear issue or Sentry alert it resolves, even for the team merging it.
- Carve out a narrow exception: a PR body may open with `Fixes: <Linear issue URL>` or `Fixes: <Sentry issue URL>`. Everything else is unchanged — code, comments, commits, and fixtures still describe changes in plain words with no ticket IDs or internal URLs, and a PR body must still explain the change in prose rather than relying on the link alone.
- Add `.github/PULL_REQUEST_TEMPLATE.md` with the placeholder so it's the default rather than something to remember each time.

## Testing
- Docs-only change; local `codex exec review --base main` clean.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,11 @@
+Fixes: <!-- Linear issue or Sentry alert URL, if any. Delete this line if none. -->
+
+## Summary
+<!-- What changed and why, in plain language. See AGENTS.md: no ticket IDs
+     or internal URLs anywhere below this line, no customer names. -->
+
+## Technical decisions / tradeoffs
+<!-- Alternatives considered and why they were rejected, if any. -->
+
+## Testing
+<!-- How this was verified: tests run, manual checks, etc. -->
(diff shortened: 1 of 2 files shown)


pavitrabhalla:
@codex review

meAmitPatil (approved):

awille-ai (commented):
Two High findings remain on the current head:

1. AGENTS.md no longer bans raw internal ticket IDs in PR bodies outside the approved tracker-link line. That broadens the policy beyond the stated single-URL exception, especially for CLI/API-created PRs that never use the template. Keep the body-wide ticket-ID ban and exempt only the permitted top-of-body URL line.

2. The PR template puts the editable tracker-link placeholder inside an HTML comment, so replacing the URL in place leaves the link invisible in the rendered PR body. Make the editable line visible or explicitly require replacing/uncommenting the whole block.

Exact-head CI and Terraform Checks are green.

awille-ai (commented):
Two High findings remain on the current head: (1) AGENTS.md no longer explicitly bans raw internal ticket IDs elsewhere in PR descriptions, so the intended single-link exception is broader than stated; (2) the PR template keeps the tracker placeholder inside an HTML comment, so replacing the URL alone leaves the link invisible. CI is green, but these should be corrected to make the exception narrow and usable.

PR #434, opened 2026-08-20
ci: deploy-trigger coverage; vmd: stall forensics and in-band boot verdicts

meAmitPatil:
**CI — deploy-trigger coverage.** A push touching only paths absent from a deploy workflow's filters runs no job at all — the change merges and never ships. Adds the packages each binary imports to its workflow's filters, plus a CI check deriving them with go list -deps that fails when one is missing. The check also requires each workflow to self-trigger (deploy-api didn't), counts only real trigger entries (not commented-out paths), and sees dot-prefixed paths.

**vmd — make failing creates diagnosable.** A guest that loads its snapshot and never becomes ready currently dies with no evidence and a misleading error. This PR: captures the console on readiness timeout — a content-free line summary in the logs (any output? reached the agent? kernel panic?) with the raw file quarantined root-only and capped, or a launch-scoped journald summary for unit-supervised VMs; stamps io pressure on restore logs beside cpu/mem (now cached off the hot path); and returns the honest "guest not ready" verdict by clamping both readiness gates (restore and resume) to the caller deadline minus the bounded error path, so the definitive error always beats the client deadline instead of racing it into a doomed transient retry.

Supporting hardening from review: failure-path durable writes are deferred off the reply path with lifecycle-lock serialization, replacement repair, and destroy compensation; in-memory state flips before replying; the crash-safety-critical cgroup teardown marker deliberately remains synchronous (tested invariant). No successful-path latency is added anywhere; the restore path loses two /proc reads.


meAmitPatil:
@codex review

meAmitPatil:
Right — deploy-vmd builds and ships four binaries and I derived dependencies from one. Fixed in c62cc784: the check now takes the union across every binary a workflow ships.

That was not theoretical. The union adds internal/secretsproxy, which only the secretsproxy binary imports and vmd's graph never reaches — so the previous version could not have caught its removal from the filters. Verified by deleting that path locally and confirming the check fails on it, then restoring.

Both workflows still pass with the union, so no additional path entries were needed beyond the ones already in this PR.

meAmitPatil:
@codex review

meAmitPatil:
Right again, and this is the third time the same shape of mistake bit this PR — a hand-maintained list missing an entry. Fixed in 29883717 by removing the list.

deploy-proxy.yml had the gap you describe for internal/analytics, internal/preview, internal/sentrylog and internal/vmdclient; its filters now list them. But the more useful change is that the check no longer names workflows or binaries at all: it iterates .github/workflows/deploy-*.yml and derives each one's binaries from the cmd/ entries in its own path filters. A new deploy workflow is covered the moment it exists, with no edit here.

That immediately turned up a fourth workflow I had not accounted for, deploy-otel-collector.yml, which correctly skips as shipping no Go binary. Verified the guard still fails by deleting internal/preview from the proxy filters and confirming it was caught, then restoring.

meAmitPatil:
@codex review

meAmitPatil:
Both real, both fixed in b16ec8db — and the first one would have failed the job on every run.

The pipefail catch is the one I could not have found the way I was testing: my shell is not -eo pipefail, so the unmatched grep on deploy-otel-collector.yml exited 0 locally and killed the step in CI. Extraction now ends in || true, and I re-ran the check by parsing the run block straight out of ci.yml and executing it under bash -eo pipefail, which is what I should have been doing from the start.

Broadening past internal/ then exposed a bug in my own comparison: filters are written proto/** while dependencies are proto/boxdpb, so string equality could never match them. Coverage is now a prefix test — a filter covers a package when it is that package or one of its parents — which is what the path filters actually mean. proto/** added to the proxy workflow.

Verified both directions under CI semantics: passes clean, and deleting proto/** from the proxy filters produces the expected failure naming proto/boxdpb.

meAmitPatil:
@codex review

meAmitPatil:
Both fixed in 774465fd.

The circularity is the more interesting one: binaries now come from each workflow's own go build commands — or, for the Docker-based API deploy, from the Dockerfile's build line — never from the path filters. The command directories then get checked like any other dependency, so an omitted cmd/ entry is reported instead of quietly shrinking the set that gets validated. Verified by deleting cmd/boxd/** from the vmd filters: the previous version went quiet, this one names it.

go.mod and go.sum are now required of every workflow that builds a Go binary, and were genuinely missing from both proxy and vmd — a dependency bump touching only those files would have left both stale. Only deploy-api had them.

Both directions re-run under bash -eo pipefail against the run block parsed out of ci.yml: clean pass, and each negative case fails as expected.

meAmitPatil:
@codex review

meAmitPatil:
This one is the check doing its job rather than a bug in it.

CI evaluates the branch merged into current main, and main gained an internal/backup import in the control plane while this PR was in review (#408 / #411). deploy-api.yml did not list that package, so a change to it today would merge without deploying the control plane — precisely the failure this job exists to catch, found on its first run against code it had not seen.

My local checkout predated those merges, which is why it passed here. Merged main in, added internal/backup to the control-plane filters, and re-ran the job script under GOOS=linux with bash -eo pipefail: clean.

meAmitPatil:
@codex review

meAmitPatil:
Both fixed in 2cead82f.

Glob semantics are now honored rather than normalized away: triggers keep their /** and coverage is evaluated by form. A directory of Go source is covered only by a recursive filter on it or an ancestor, since GitHub's non-recursive 'internal/db' matches a file at that exact path and would not start a deploy for internal/db/db.go. Files — go.mod, go.sum, and the Dockerfile a Docker-based workflow builds through — are covered by an exact name or by a recursive filter above them.

Fixing that surfaced two bugs of my own, both invisible to the earlier version. Keeping ** in the trigger strings meant the unquoted loop variable was glob-expanded against the working tree before comparison, so every path reported as uncovered; the workflow list is now expanded first and pathname expansion disabled for the rest. And the trigger regex only accepted lowercase-initial paths, so 'Dockerfile' was never read as a filter at all — which the new Dockerfile requirement immediately exposed.

Verified both negative cases: rewriting internal/db/** as internal/db is reported, and deleting 'Dockerfile' from the API filters is reported. Clean otherwise.

awille-ai (commented):
Re-reviewed the latest head. The previously reported deploy-trigger coverage gaps are addressed in the current diff, and I don't see a remaining high-severity code finding. I am not approving this head yet because its exact-head CI run is still red: the failing job is the integration-test job (`Run integration tests`), while the Terraform checks pass. This is non-blocking from my review side, but I’d want that CI failure understood/cleared before approval.

meAmitPatil:
@codex review

pavitrabhalla (approved):

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

PR #436, opened 2026-08-20
vmd: move host firewall rules into owned chains for cohabitated hosts

meAmitPatil:
Moves vmd's forwarding and redirect rules into vmd-owned chains entered via veth-scoped jumps, so another agent's rules (e.g. Docker's) and vmd's cannot interfere; the shared surface shrinks to three head jumps and the MASQUERADE. Startup verifies the ruleset from one iptables-save dump and skips the install when intact; otherwise shared-chain changes are planned from a locked snapshot before any mutation, with foreign policy never demoted, replaced plumbing put back where it stood, and a raced repair — detected by rule content and nf_tables handle — yielding below the other writer's rule rather than removing enforcement. A once-a-minute reconciliation repairs drift after readiness.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):

PR #429, opened 2026-08-19
vmd: verify the host firewall in one dump and skip the install when intact

meAmitPatil:
Verify the host firewall against a single `iptables-save` dump at startup and skip the install when the ruleset is provably intact (presence, multiplicity, and ordering of every vmd rule; owned chains matched exactly). On any mismatch the unchanged installer runs, followed by an ordering/dedup repair and a mandatory re-verify — startup fails rather than serve with unverified enforcement. Nothing persisted, rollback-safe, drift repair unchanged; steady-state boots replace dozens of iptables execs with one dump.

PR #437, opened 2026-08-20
Revert "vmd: verify the host firewall in one dump and skip the install when intact"

meAmitPatil:
Reverts #429: its fail-closed verification rejects hosts where another agent's unconstrained jump precedes vmd's rules, so vmd cannot start there. Restores the previous installer so main deploys cleanly everywhere; the verify-and-skip optimization returns in #436 with rules moved into vmd-owned chains.


meAmitPatil:
@codex review

awille-ai (approved):
Reviewed this as a revert, not as a standalone firewall redesign. The patch cleanly restores the pre-#429 installer path across the same six-file footprint, removing the verifier/repair machinery rather than leaving a mixed state. That intentionally gives up #429's fast-path verification and stronger ordering/drift checks, but it also removes the fail-closed startup behavior that rejects hosts with pre-existing foreign jump rules; #436 is the follow-up that moves vmd rules into owned chains before reintroducing verification. I don't see a remaining High/Critical correctness, security/resilience, test, or rollout issue in the revert itself. Exact-head CI and Terraform Checks are green.

PR #439, opened 2026-08-21
billing: make the rollup interval-overlap scan sargable

meAmitPatil:
The hourly rollup and backfill schedulers find the teams active in a window with `COALESCE(ended_at, LEAST(now(), $2)) > $1`. Wrapping the column hides it from the planner, so both interval scans fall back to a full table read whose cost grows with total billing history rather than with the window.

Given the `$1 < LEAST(now(), $2)` guard already in the same conjunction, that is exactly equivalent to `ended_at IS NULL OR ended_at > $1` — which the partial indexes already on these tables can serve, turning the sequential scan into a bitmap OR across them. No schema change; the two schedulers' duplicated CTE is also collapsed into one shared constant.

Since this is a billing path the equivalence is proved against Postgres rather than argued: an integration test runs both spellings over the same rows across windows hitting every boundary (open intervals, zero-length, exact start/end, straddling now, future) and requires identical row sets.

### DIFF
--- a/internal/billing/rollup.go
+++ b/internal/billing/rollup.go
@@ -237,12 +237,20 @@ RETURNING true`
 	return claimed, err
 }
 
-func enqueueHour(ctx context.Context, pool *pgxpool.Pool, hourStart, hourEnd time.Time, refreshCompleted bool) (int64, error) {
-	// interval_team_set is MATERIALIZED so the flag check runs against the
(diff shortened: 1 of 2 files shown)


meAmitPatil:
@codex review

awille-ai (approved):

PR #441, opened 2026-08-24
Revert "billing: make the rollup interval-overlap scan sargable (#439)"

meAmitPatil:
Reverts #439. The rewrite made the predicate indexable, but indexable is not the same as selective: every live sandbox holds an *open* storage interval, so the `ended_at IS NULL` arm matches essentially the whole live fleet rather than a narrow slice. Measured against production data the new plan reads more rows than the sequential scan it replaced and comes out slower, not faster.

The change never took effect — `internal/billing/**` is absent from deploy-api's path filter, so no revision ever ran it — but it should not sit on main waiting for an unrelated deploy to carry it out.

The underlying cost is real and still worth fixing, but the right shape is not settled and is not a query rewrite. A storage interval stays open for as long as its sandbox exists, so it overlaps every hour in between; deriving the active-team set by scanning intervals is therefore proportional to the live fleet no matter how it is indexed. Reverting first so main reflects what actually runs.


awille-ai (approved):
Reviewed at head 8e772ed1826824131976aa73947db55f9c35b1c7. This cleanly reverts #439's two-file footprint and restores the pre-#439 query shape. The revert avoids carrying an optimization that is indexable but not selective for long-lived open storage intervals, and #439 was never deployed so there is no runtime compatibility transition to manage. No High/Critical correctness, billing-data, performance, security/resilience, or test-quality findings. Exact-head CI and Terraform Checks are green.

PR #443, opened 2026-08-25
infra: null out backup_coverage_alerts to break a Terraform CD deadlock

pavitrabhalla:
## Summary
- Merging #409 revealed a deploy-pipeline deadlock: the new `google_monitoring_alert_policy.backup_coverage` resources require `backup_uncovered_paused_sandboxes` to already exist as a Cloud Monitoring metric type, but the controlplane code that emits it deploys through this same workflow's api job, which is gated on this apply succeeding first. Apply fails on the missing metric -> api job never runs -> metric never appears -> apply keeps failing, indefinitely.
- Verified concretely: staging has 18 real paused sandboxes and it's been well over an hour since the original merge, but the metric shows zero data points anywhere in Cloud Monitoring, consistent with the emitting code never having reached Cloud Run.
- Fix: null `backup_coverage_alerts` in all three environments (its documented default — "Null omits the policy entirely") so `terraform apply` succeeds with zero coverage alert_policy resources, unblocking the gated api-deploy job so the sampler code actually ships.

- Considered `-target` to apply everything except the three alert_policy resources — rejected: still needs the same api-deploy job to run afterward, and this is simpler and reversible with an obvious diff.
- Left the original `backup_coverage_alerts` blocks commented out in place (not deleted) so the follow-up re-enabling them is a one-line uncomment, not a re-author.

## Testing
- `terraform validate` clean on all three environments (this is a config-only change; no state/credentials available locally to `plan`).
- Local Codex review: clean, no findings.

Follow-up once the metric is confirmed live in Cloud Monitoring: uncomment `backup_coverage_alerts` in all three files to re-enable the disabled-by-default alert policies.

### DIFF
--- a/infra/envs/production/us-east4/main.tf
+++ b/infra/envs/production/us-east4/main.tf
@@ -380,11 +380,21 @@ module "observability" {
   # under us-central1 labeling before the host swap, so the legacy
   # value stays scoped: host rows relabeled or restored under it must
   # not fall outside the alert.
-  backup_coverage_alerts = {
-    enabled        = false
-    display_prefix = "Backup coverage / us-east4"
-    regions        = ["us-east4", "us-central1"]
-  }
+  #
+  # Left null for now (was the object below): creating these alert_policy
+  # resources requires backup_uncovered_paused_sandboxes to already exist
+  # as a Cloud Monitoring metric type, but the controlplane code that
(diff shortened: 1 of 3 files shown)


pavitrabhalla:
@codex review

meAmitPatil (approved):

PR #444, opened 2026-08-25
api: cache the per-create billing eligibility gate and coalesce its background rechecks

meAmitPatil:
Fronts the per-team billing eligibility check with a TTL cache (eligible 30s / ineligible 3s, singleflighted, hard-capped), so create and resume stop paying a DB round trip per request; enforcement is unchanged — the 30s sweep and the coalesced post-activation recheck still bound how long an ineligible team can run.
Also dedupes the sweep's scan so the eligibility function runs once per active team instead of once per active sandbox row.

### DIFF
--- a/db/queries/billing.sql
+++ b/db/queries/billing.sql
@@ -803,13 +803,22 @@ ORDER BY s.team_id
 LIMIT sqlc.arg(batch_limit);
 
 -- name: ListTeamsWithActiveIneligibleSandboxes :many
-SELECT DISTINCT s.team_id
-FROM sandbox s
-WHERE s.destroyed_at IS NULL
-  AND s.status = 'active'
-  AND NOT team_sandbox_billing_eligible(s.team_id)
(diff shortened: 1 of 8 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
Reviewed the current head with the usual correctness, security/resilience, test-quality, performance, lifecycle, and rollout bar. The prior coalescing issue is fixed by sharing only the eligibility read while keeping each activation's pause claim independent. The request-path cache is TTL-bounded, singleflighted, hard-capped, does not cache errors, lets canceled callers return without canceling the shared bounded DB read, and preserves enforcement via the post-activation reconciliation plus periodic sweep. The sweep query now deduplicates active teams before evaluating eligibility. Exact-head CI and Terraform Checks are green; no remaining High/Critical findings.

PR #445, opened 2026-08-25
controlplane: place sandboxes by published host capacity with an atomic admission fence

meAmitPatil:
Places new sandboxes by the capacity each host publishes, with a per-create reservation that concurrent control-plane replicas cannot both spend. Today's scheduler picks between two random active hosts by sandbox count and cannot see memory, CPU or network slots at all; two replicas can pile onto the same host in the gap between reports.

Stacked on the capacity publication branch (it consumes `host_pressure`); the base retargets to `main` once that merges.

## Prod-inert until opted in

`SCHEDULER_CAPACITY_ADMISSION` defaults off, and with it off the control plane constructs the same least-loaded scheduler as before and takes the same code path. Enabling it changes behavior only for hosts that advertise `capacity_pressure_v1`.

**Do not enable it before template artifacts exist on every host in the cell.** Template snapshot/mem paths are host-local files resolved before placement, so unconstrained placement would dispatch a restore to a host that does not hold it. `PlacementRequest.AllowedHosts` is the hook for that constraint; its producer is separate work.

## How admission stays correct

Three host states, decided by the capability on the current heartbeat:

- **capable + fresh report** — fenced capacity admission.
- **capable + stale/missing report** — excluded. A broken publisher under a healthy heartbeat must not take unbounded placements on counts the fence cannot see.
- **not capable** — legacy least-loaded placement, so a mixed fleet rolls through safely.

The fence is one statement on the create path. It increments charge counters **on the pressure row** and inserts a ledger row keyed by sandbox id. The counters have to live on that row: under READ COMMITTED a statement's snapshot is taken before it blocks, so a replica waiting on a row lock would still count the ledger as it stood before the winner inserted — both would pass the same limit check. An UPDATE instead re-evaluates its WHERE against the committed row version after the lock is granted, so the loser sees the winner's charge. An integration test races 20 admissions against room for 3 and asserts exactly 3 win; before this shape it let 8 through.

Reservations retire on their own evidence, never on report-generation resets: a failed create deletes its row (capacity back immediately), a successful one is stamped when the vmd RPC returns and keeps charging until a report lands slack-clear of that stamp — because a report sampled before a VM materialized can arrive after it. Every report recomputes the counters from the ledger, so any drift self-heals within one heartbeat.

Ranking is cached (same TTL and stale-grace contract as today): hosts score by their worst pressure ratio across memory, vCPU, slots and sandbox count; everything within a tolerance band is shuffled, warm-slot hosts first within the band. The randomization is load-bearing — a globally-best-first order would make one host's row the fleet's serialization point.

## Also included

- `capacity_pressure_v1` advertised by vmd under the same condition it publishes pressure, deliberately not gated on the startup readiness gate.
- Live reservations counted by `hostctl list` (`RSV`) and `drain --wait`: a host reads zero everywhere else during the reservation-to-dispatch gap, and retiring it then would strand a launching VM.
- `sandbox_placement_duration_seconds` with bounded mode/result/reason/attempts labels — this is a new synchronous component of create latency and should be visible as one.

## Latency

With the flag off: unchanged. With it on: one single-row statement before dispatch (~1-3ms uncontended), plus up to two more only when a candidate rejects. Ranking rides the existing cache; the create path runs no fleet-wide query.

## Tests

18 scheduler unit tests (fail-closed staleness, legacy fallthrough, bounded retries, ranking, in-band randomization, constraint handling, cache) and 12 integration tests against a real Postgres (the 20-way race, slot ceiling vs warm inventory, idempotent retries, slack retirement, lease expiry, double-release, drift healing, drain visibility, cascade on host reclaim).


meAmitPatil:
CI does not run on this PR — the workflow triggers only on PRs targeting `main`, and this one is stacked on `amit/capacity-publication`. Verified by running the CI commands in a linux container against this exact tree instead (`go build ./...`, `go vet ./...`, `go test -race`), including the packages that cannot build on darwin:

```
ok  internal/scheduler   internal/api   internal/db   internal/vm (45s)
ok  internal/integration (race, real postgres)
```

Full unit sweep across every package also passed. Retargeting to `main` after #438 merges will run the real pipeline.

meAmitPatil:
Superseded by #460.

The Postgres reservation fence here costs a synchronous database round trip per create and serializes bursts on one `host_pressure` row. Enforcement is moving into vmd instead — an in-memory admission gate inside the existing create RPC, where the check is a mutex rather than a network round trip and contention scopes per host process rather than per database row.

#460 carries the reservation-free half: ranking, freshness, capability handling, region preference, under-described tiering, the template-locality hook and metrics — evaluated in shadow, influencing nothing, with no schema migration.

Closing rather than reworking: roughly half this diff is the design being dropped, and the branch has already been rebased and rewritten several times. Left in history because the reservation design and its review are worth being able to read back.

PR #448, opened 2026-08-26
vmd: diagnose a stalled guest while it is still running

meAmitPatil:
The teardown capture answered where the stall is *not*: the memory handler sits idle with zero unresolved faults, so the guest itself is what hangs — one vCPU spinning in guest mode, another halted waiting for an interrupt. Answering *why* needs signals that exist only while the VM runs.

This fires diagnostics from inside the readiness window instead of at teardown. A healthy guest answers in tens of milliseconds, so after a few seconds the outcome is already decided, yet vmd currently sits idle until the deadline. Using that dead time costs the request nothing — it is not on the reply path — and leaves most of a minute for probes a teardown path could never afford:

- KVM counters sampled twice, reported as deltas per vCPU: whether the spinning vCPU is exiting at all, and whether the halted one is ever woken. A flat exit count means a tight guest loop; a flat halt_wakeup means nothing is arriving.
- guest instruction pointers, sampled with perf: which code the guest is actually looping in. Raw addresses, resolved offline against a symbol dump, since the shipped guest kernel is stripped.
- thread stacks and fault counters re-read for correlation.

Read-only with respect to the VM — nothing interferes with a guest that might still recover. Concurrency is capped and saturated attempts are skipped rather than queued, the whole battery is bounded well inside the readiness window, and on a successful restore the entire cost is a timer that never fires.

### DIFF
--- a/internal/vm/manager.go
+++ b/internal/vm/manager.go
@@ -3079,7 +3079,15 @@ func (m *Manager) restoreVMSnapshot(ctx context.Context, vmID, snapshotPath, mem
 			readiness = remaining
 		}
 	}
-	if err := m.waitForBoxd(ctx, hostIP, readiness); err != nil {
+	// Diagnostics fire from inside the readiness window if the guest is still
+	// unready after a few seconds — the teardown capture proved the guest
+	// itself is what stalls, and answering why needs signals that exist only
+	// while it runs. Cancelled on the healthy path, where the whole cost is a
+	// timer that never fires.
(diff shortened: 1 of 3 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):

PR #452, opened 2026-08-27
vmd: read stalled-guest RIPs from the kvm_entry tracepoint

meAmitPatil:
## What

Replaces the guest instruction-pointer sampling in the stall diagnostics: instead of `perf kvm --guest record` (PMU sampling), record the `kvm:kvm_entry` tracepoint and histogram the guest RIP per vCPU.

## Why

Guest PMU sampling is not virtualized on our hosts — `perf kvm --guest record` runs cleanly but captures zero guest samples, which is why the `guest_ips` field of every stall specimen so far has been empty. The `kvm_entry` tracepoint is host-side: KVM logs the guest RIP on every VM entry, no guest PMU involved. A stalled vCPU still enters thousands of times per second (the host timer tick alone forces exits), so the address it spins at clusters sharply — validated against a live guest before writing this change.

Reporting is now per-vCPU (`v<cpu>:<rip>:<count>`), which separates the spinning vCPU's cluster from the halted one's — the distinction the diagnosis rests on.

Also: a perf probe that reaches the distribution stub is no longer cached negatively, so installing the real perf under a running daemon takes effect on the next stall instead of requiring a restart.

## Testing

- The parsing regex was validated against real `perf script` output before writing the fixture — the tracepoint's print format varies across kernel versions (`vcpu N rip 0x...` vs `vcpu N, rip 0x...`), and the test covers both.
- Full `internal/vm` suite in a linux container: pass.
- `GOOS=linux go vet` and full build: clean.

### DIFF
--- a/internal/vm/stall_diag.go
+++ b/internal/vm/stall_diag.go
@@ -35,11 +35,9 @@ const (
 	stallDiagSettle     = 2 * time.Second
 	stallDiagPerfWindow = 3 * time.Second
 
-	// A stall means the host may already be struggling, so the profile is
-	// deliberately coarse: ~100 Hz is ample to identify a loop the guest has
-	// been spinning in for seconds, and the size cap bounds the write even if
-	// the rate assumption is ever wrong.
-	stallDiagPerfHz      = "97"
+	// kvm_entry fires per VM entry — a few thousand per second on a spinning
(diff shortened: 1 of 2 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):

PR #453, opened 2026-08-27
ci: deploy a published Firecracker release to the sandbox hosts

meAmitPatil:
Installs a published Firecracker release across a cell — fetching a fixed, digest-verified artifact rather than building, so every host provably runs the same bytes.

Per host: refuse if the rollback binary is missing, verify the digest, loader pre-check while still in `/tmp`, back up (written atomically, named by the digest it preserves, and verified before anything is overwritten), install by atomic rename, re-verify. Sequential, stopping on the first failure.

Snapshots have to survive the swap in both directions, so production is gated on `staging_verified`: a fresh create under the new version, a pre-install pause resuming after, and a pause under the new version resuming under the rollback target.

### DIFF
--- a/.github/workflows/deploy-firecracker.yml
+++ b/.github/workflows/deploy-firecracker.yml
@@ -28,6 +28,13 @@ on:
           - staging
           - production
 
+# One rollout at a time, and never cancelled: two overlapping dispatches for
+# different versions could interleave their per-host loops and leave a cell on
+# mixed digests, with both runs reporting success.
+concurrency:
+  group: deploy-firecracker
+  cancel-in-progress: false
+
 permissions:
   contents: read
   id-token: write
@@ -47,15 +54,19 @@ jobs:
 
       - uses: google-github-actions/setup-gcloud@v2
 
-      - name: Install ${{ inputs.version }}
+      # The dispatch input reaches the script through the environment, never
+      # through the run body: `${{ }}` is substituted before bash parses the
+      # script, so an interpolated tag could close the quoting and run
+      # arbitrary commands on a job that has already authenticated to GCP.
+      - name: Install
         env:
           GCP_PROJECT: ${{ vars.GCP_PROJECT }}
           GCP_REGION: ${{ vars.GCP_REGION }}
           VMD_LABEL: ${{ vars.VMD_LABEL }}
+          VERSION: ${{ inputs.version }}
           GH_TOKEN: ${{ github.token }}
         run: |
-          python3 .github/workflows/scripts/deploy-firecracker.py \
-            --version '${{ inputs.version }}'
+          python3 .github/workflows/scripts/deploy-firecracker.py --version "$VERSION"
 
   deploy-production:
     name: Production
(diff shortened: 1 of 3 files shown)


meAmitPatil:
@codex review

meAmitPatil:
All four addressed. The gate is now `staging_verified` and names all three directions explicitly — fresh create under the new version, pre-install pause resuming after, and a pause under the new version resuming under the rollback target — with the failure message stating that skipping the third makes the rollout forward-only.

Backups are now written via temp-file-and-rename and their full digest is checked against the live binary before anything is overwritten, so a partial file can no longer be inherited. Region matching is exact (`zone.rsplit('-', 1)[0] == region`), and non-running labelled instances are now listed at deploy time so a cold standby is visible rather than silently skipped.

meAmitPatil:
Two of three addressed; the standby one is deliberately unchanged.

**Rollback binding** — the script now preflights the whole cell before touching any host: it reads each host's currently-installed digest, refuses if they disagree with each other (so the rollback target staging tested is actually the one present), and accepts an optional `EXPECTED_CURRENT_SHA256` to pin it explicitly. Drift now stops the rollout instead of being discovered halfway through it.

**Rollback floor** — `test -f` is no longer the whole check; the stock binary must also run (`--version`) before the live one is touched, so a truncated or non-executable floor fails closed.

**Standby discovery** — leaving this as-is, matching `deploy-vmd.py` exactly (`component=vmd AND status=RUNNING`). The finding is right for west, where the component label swaps on promotion so the standby is structurally invisible — but not for east, where the standby carries the same `component=vmd` and is excluded only by being stopped. That asymmetry is pre-existing and applies equally to the vmd deploy; it is better fixed once in the label scheme and then reflected in both deploys than special-cased here. I removed the enumeration I had added, since it could never see west's standby and gave false assurance.

meAmitPatil:
Right on both counts — including that the check was passing vacuously, since one active host per region makes "all hosts agree" trivially true.

Production now takes a required `rollback_sha256` input, validated as 64 lowercase hex before anything else runs, and exported as `EXPECTED_CURRENT_SHA256` to every cell — so `staging_verified` is bound to the digest production actually runs rather than an unstated one. The script rejects a malformed value independently, covered by a test.

Both cells are also preflighted before either is installed to (`--preflight-only`), so a second cell found on an unexpected binary halts the rollout instead of being discovered after the first has already been swapped.

meAmitPatil:
Real deadlock, and in the worst possible situation — a split fleet is exactly when finishing the rollout matters most. Fixed.

Preflight is no longer a uniformity check. Each host is now accepted in either of two states: still on the verified rollback binary, or already on the target **provided it still holds a correct backup of that rollback binary**, so it has not lost the way back. Any third digest is a binary nobody verified against and still stops the rollout. Already-updated hosts then no-op on install and the rest converge, so a retry completes rather than being rejected from both directions.

The live digest and the backup check now come back in one round trip. The decision itself is a pure function, covered by five tests: not-yet-updated, already-updated-with-good-backup (the retry case), updated-with-missing-backup, updated-with-wrong-backup, and unknown binary. 29 tests total.

meAmitPatil:
@codex review

meAmitPatil:
Both real, fixed.

**Rollback direction** — the backup requirement was asymmetric: rolling back from B to A, an untouched host is already on A and never ran B, so it can hold no backup of B and was rejected, blocking the rollback of the hosts that *were* changed. A host already at the target now passes in either direction; a missing local backup is reported rather than fatal, since rollback installs a release and the `.bak` is break-glass. Covered by a test that asserts a rollback across a partly-updated cell passes for both hosts.

**Target binding** — right that comparing the tarball against the checksum published beside it only proves the pair is self-consistent, and that each cell downloads independently. Production now requires `target_sha256` — the digest verified on staging — validated as 64 lowercase hex and enforced inside the script, so a release whose assets were replaced after verification is refused rather than installed.

meAmitPatil:
@codex review

meAmitPatil:
Correct — I conflated "does this cell exist" with "is its region configured", so an unset `GCP_REGION_USW` skipped the cell and reported success with it still on the old binary.

Both usw steps now gate on `CLOUD_RUN_SERVICE_USW`, matching `deploy-vmd.yml`, and a missing region falls through to the script's required-region check, which already refuses. Cell absent → skipped deliberately; cell present but misconfigured → aborts.

meAmitPatil:
@codex review

meAmitPatil:
Fixed — the staging step passed neither pinned digest, so on a production dispatch nothing held staging to the predecessor production is actually moving off.

It now receives both `EXPECTED_CURRENT_SHA256` and `TARGET_SHA256`. They are empty on a staging-only dispatch, which is the run that discovers them, and set on a production dispatch — so staging on some third binary now aborts instead of silently becoming the proving ground for a transition production will never make.

One residual worth naming rather than implying otherwise: this verifies staging's *current state*, not its history. If staging is already at the target from the earlier run it passes without re-proving which predecessor it came from — that part still rests on `staging_verified`. Tightening it further would mean requiring staging to hold a `.pre-<rollback>.bak`, which is the same requirement that deadlocked rollback earlier, so I would rather not reintroduce it without a clearer need.

meAmitPatil:
@codex review

awille-ai (approved):
## Verdict
APPROVE

I did not find any remaining Critical or High issues in the current head (`8def8be`). The earlier rollout-safety findings are addressed: production covers both configured cells, the live binary swap is atomic, dispatch input is kept out of shell source, deploys are serialized, partial-rollout rollback is retryable, production is pinned to both the staging-verified target digest and the production predecessor digest, staging is bound to that predecessor, and the usw cell fails closed on missing region configuration.

## Review scope
Reviewed the full PR diff (`deploy-firecracker.yml`, `deploy-firecracker.py`, and `test_deploy_firecracker.py`), the surrounding `deploy-vmd` topology/discovery pattern, current review threads, and current CI wiring. Independent passes covered hostile/adversarial failure modes, security/resilience, test quality, correctness/maintainability, observability, compatibility/rollout, infrastructure-deployment safety, and sandbox lifecycle performance.

## Validation
- Confirmed the deploy-script tests are wired into the repository CI via `python3 -m unittest discover -s .github/workflows/scripts -p 'test_*.py'`.
- Confirmed current workflow topology matches the existing use4/usw production deployment pattern.
- Traced forward deploy, partial failure + retry, and reverse rollback states through `host_state`, preflight, and the per-host install path.
- Checked artifact digest verification, target digest pinning, predecessor digest pinning, backup creation/verification, loader/capability pre-checks, and atomic rename semantics.
- Checked malformed/missing region/digest handling and the manual staging-compatibility gate.

At the time of this review, the latest CI run was still in progress; completed jobs observed so far were green, while the Unit Tests job (which includes the new deploy-script tests) had not yet completed. That is a CI-status caveat rather than a code-review finding.

PR #455, opened 2026-08-27
feat: 2. Publish VMD host capabilities and heartbeat to Supabase

awille-ai:
## Summary

- publish explicit lifecycle operational capabilities from the running VMD service
- expose proxy/files operational state through the existing proxy health probe
- preserve preview protocol capabilities and host self-description
- keep status/capacity separate from technical capability
- keep lifecycle capabilities gated until VMD is ready to serve lifecycle RPCs

## Testing

- extend focused proxy health and VMD heartbeat tests
- verify Linux compilation with `GOOS=linux go test -c ./internal/vm`

### DIFF
--- a/internal/vm/heartbeat.go
+++ b/internal/vm/heartbeat.go
@@ -39,7 +39,6 @@ type HeartbeatConfig struct {
 	Region            string
 	CapacityMemoryMib int32
 	CapacityVcpus     int32
-
 	// Pressure, when set, is sampled after each SUCCESSFUL heartbeat and
 	// published to the separate best-effort pressure endpoint — never
 	// inside the heartbeat body, so no pressure outcome can ever affect
@@ -57,6 +56,10 @@ type HeartbeatConfig struct {
 	// limits published alongside pressure; 0 means unset (no cap).
 	MaxSandboxes    int32
 	MaxNetworkSlots int32
+	// LifecycleReady gates lifecycle capability publication until VMD is ready.
+	LifecycleReady func() bool
+	// ResolverReady gates proxy, file, and preview capability publication.
+	ResolverReady func() bool
 }
 
 // pressureProbeEvery is how many beats a publisher that found the
@@ -274,8 +277,26 @@ type heartbeatRequest struct {
 }
 
 type proxyHealthResponse struct {
-	Capabilities []string `json:"capabilities"`
+	Capabilities  []string `json:"capabilities"`
+	FilesEnabled  bool     `json:"files_enabled"`
+	ResolverReady bool     `json:"resolver_ready"`
 }
+
+type proxyHealthState struct {
+	PreviewCapabilities []string
+	FilesEnabled        bool
+	ResolverReady       bool
+}
+
+const (
+	capabilityCanCreate       = preview.HostCapabilityCanCreate
+	capabilityCanResume       = preview.HostCapabilityCanResume
+	capabilityCanPause        = preview.HostCapabilityCanPause
+	capabilityCanDestroy      = preview.HostCapabilityCanDestroy
+	capabilityCanProxyTraffic = preview.HostCapabilityCanProxyTraffic
+	capabilityCanReadFiles    = preview.HostCapabilityCanReadFiles
+	capabilityCanWriteFiles   = preview.HostCapabilityCanWriteFiles
+)
 
 type heartbeatStorageCache struct {
 	mu           sync.RWMutex
@@ -336,12 +357,28 @@ func (c *heartbeatStorageCache) store(measurements []heartbeatStorageMeasurement
 
 func sendHeartbeat(ctx context.Context, client *http.Client, cfg HeartbeatConfig, url, token, proxyHealthURL string, storage []heartbeatStorageMeasurement, log zerolog.Logger) (bool, bool) {
 	started := time.Now()
-	capabilities, err := proxyPreviewCapabilities(ctx, client, proxyHealthURL)
+	proxyState, err := proxyHealthCapabilities(ctx, client, proxyHealthURL)
 	if err != nil {
 		log.Warn().Err(err).
+			Str("host_id", cfg.HostID).
 			Dur("duration", time.Since(started)).
-			Msg("proxy capability probe failed; advertising no preview capabilities")
-		capabilities = nil
+			Msg("proxy health probe failed; advertising no proxy or file capabilities")
+		proxyState = proxyHealthState{}
+	}
+	capabilities := make([]string, 0, 7)
+	lifecycleReady := cfg.LifecycleReady != nil && cfg.LifecycleReady()
+	resolverReady := cfg.ResolverReady != nil && cfg.ResolverReady()
+	if lifecycleReady {
+		capabilities = append(capabilities, capabilityCanCreate, capabilityCanResume, capabilityCanPause, capabilityCanDestroy)
+	}
+	if err == nil {
+		if resolverReady && proxyState.ResolverReady {
+			capabilities = append(capabilities, capabilityCanProxyTraffic)
+			if proxyState.FilesEnabled {
+				capabilities = append(capabilities, capabilityCanReadFiles, capabilityCanWriteFiles)
+			}
+		}
+		capabilities = append(capabilities, proxyState.PreviewCapabilities...)
 	}
 	return postHeartbeat(ctx, client, cfg, url, token, capabilities, storage, log, started)
 }
@@ -498,34 +535,39 @@ func measureOverlayStorage(runDir string) ([]heartbeatStorageMeasurement, error)
 }
 
 func proxyPreviewCapabilities(ctx context.Context, client *http.Client, healthURL string) ([]string, error) {
+	state, err := proxyHealthCapabilities(ctx, client, healthURL)
+	return state.PreviewCapabilities, err
+}
+
+func proxyHealthCapabilities(ctx context.Context, client *http.Client, healthURL string) (proxyHealthState, error) {
 	probeCtx, cancel := context.WithTimeout(ctx, 2*time.Second)
 	defer cancel()
 	req, err := http.NewRequestWithContext(probeCtx, http.MethodGet, healthURL, nil)
 	if err != nil {
-		return nil, fmt.Errorf("build proxy health request: %w", err)
+		return proxyHealthState{}, fmt.Errorf("build proxy health request: %w", err)
 	}
 	resp, err := client.Do(req)
 	if err != nil {
-		return nil, fmt.Errorf("request proxy health: %w", err)
+		return proxyHealthState{}, fmt.Errorf("request proxy health: %w", err)
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
 		_, _ = io.Copy(io.Discard, resp.Body)
-		return nil, fmt.Errorf("proxy health returned %d", resp.StatusCode)
+		return proxyHealthState{}, fmt.Errorf("proxy health returned %d", resp.StatusCode)
 	}
 	var health proxyHealthResponse
 	if err := json.NewDecoder(io.LimitReader(resp.Body, 4<<10)).Decode(&health); err != nil {
 		if err == io.EOF {
-			return nil, nil
+			return proxyHealthState{}, nil
 		}
-		return nil, fmt.Errorf("decode proxy health: %w", err)
+		return proxyHealthState{}, fmt.Errorf("decode proxy health: %w", err)
 	}
 	advertised := make(map[string]bool, len(health.Capabilities))
 	for _, capability := range health.Capabilities {
 		advertised[capability] = true
 	}
 	if !advertised[preview.HostCapabilityPorts] {
-		return nil, nil
+		return proxyHealthState{FilesEnabled: health.FilesEnabled, ResolverReady: health.ResolverReady}, nil
 	}
 	out := []string{preview.HostCapabilityPorts}
 	if advertised[preview.HostCapabilityPortAccess] {
@@ -537,5 +579,5 @@ func proxyPreviewCapabilities(ctx context.Context, client *http.Client, healthUR
 			}
 		}
 	}
-	return out, nil
+	return proxyHealthState{PreviewCapabilities: out, FilesEnabled: health.FilesEnabled, ResolverReady: health.ResolverReady}, nil
 }
--- a/internal/vm/heartbeat_test.go
+++ b/internal/vm/heartbeat_test.go
@@ -24,7 +24,7 @@ func TestSendHeartbeatAdvertisesVerifiedPreviewCapabilities(t *testing.T) {
 		switch r.URL.Path {
 		case "/health":
 			w.Header().Set("Content-Type", "application/json")
-			_ = json.NewEncoder(w).Encode(proxyHealthResponse{Capabilities: []string{
+			_ = json.NewEncoder(w).Encode(proxyHealthResponse{FilesEnabled: true, ResolverReady: true, Capabilities: []string{
 				preview.HostCapabilityPorts, preview.HostCapabilityPortAccess,
 				preview.HostCapabilityPortTokens, preview.HostCapabilityPortBrowserAuth,
 			}})
@@ -43,6 +43,8 @@ func TestSendHeartbeatAdvertisesVerifiedPreviewCapabilities(t *testing.T) {
 
 	sendHeartbeat(context.Background(), server.Client(), HeartbeatConfig{
 		HostID:            "host-a",
+		LifecycleReady:    func() bool { return true },
+		ResolverReady:     func() bool { return true },
 		VMDAddr:           "10.0.0.2:50051",
 		ProxyAddr:         "10.0.0.2:5007",
 		Region:            "region-a",
@@ -57,6 +59,9 @@ func TestSendHeartbeatAdvertisesVerifiedPreviewCapabilities(t *testing.T) {
 		t.Fatalf("authorization = %q", gotAuthorization)
 	}
 	want := []string{
+		capabilityCanCreate, capabilityCanResume, capabilityCanPause, capabilityCanDestroy,
+		capabilityCanProxyTraffic,
+		capabilityCanReadFiles, capabilityCanWriteFiles,
 		preview.HostCapabilityPorts, preview.HostCapabilityPortAccess,
 		preview.HostCapabilityPortTokens, preview.HostCapabilityPortBrowserAuth,
 	}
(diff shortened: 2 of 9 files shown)


awille-ai:
@codex review

awille-ai:
@codex review

internal/vm/heartbeat.go
```diff
@@ -165,12 +181,21 @@ func (c *heartbeatStorageCache) store(measurements []heartbeatStorageMeasurement
...
 	if err != nil {
 		log.Warn().Err(err).
+			Str("host_id", cfg.HostID).
 			Dur("duration", time.Since(started)).
 			Msg("proxy capability probe failed; advertising no preview capabilities")
-		capabilities = nil
+		proxyState = proxyHealthState{}
+	}
+	capabilities := []string{capabilityCanCreate, capabilityCanResume, capabilityCanPause, capabilityCanDestroy}
+	if err == nil {
```
awille-ai, line 374: **High — lifecycle capabilities can be true before lifecycle RPCs are serveable.** `StartHeartbeat` sends immediately, while `cmd/vmd/main.go` starts the heartbeat before `startupReady.Store(true)`. The gRPC interceptors return `Unavailable` until that flag flips, so this line can persist `can_create/can_resume/can_pause/can_destroy=true` for a host that is still rejecting every lifecycle call. On an unhealthy→recovering host, that heartbeat also reactivates the host before the request gate opens. Tie these capabilities to the same readiness boundary (or delay capability-bearing heartbeat publication until readiness) so the published contract is actually true.

internal/vm/heartbeat_test.go
```diff
@@ -57,6 +57,8 @@ func TestSendHeartbeatAdvertisesVerifiedPreviewCapabilities(t *testing.T) {
 		t.Fatalf("authorization = %q", gotAuthorization)
 	}
 	want := []string{
+		capabilityCanCreate, capabilityCanResume, capabilityCanPause, capabilityCanDestroy,
```
awille-ai, line 62: **Low — the new file-capability branch is not covered.** This test's proxy health response never sets `files_enabled=true`, so the code path that appends `can_read_files` and `can_write_files` is untested. Please add an end-to-end heartbeat case for healthy proxy + files enabled, plus the files-disabled case, so R5/R6 cannot regress silently.

awille-ai (commented):
Hostile review against SS-390 found one High correctness issue plus two non-blocking gaps.

High: lifecycle capabilities are advertised before VMD is actually ready to serve lifecycle RPCs. `StartHeartbeat` fires immediately, but `cmd/vmd/main.go` starts heartbeat before `startupReady.Store(true)`. During that window the heartbeat persists `can_create/can_resume/can_pause/can_destroy=true` while the gRPC interceptors return `Unavailable` for every request. This is especially dangerous on recovery because the heartbeat can also reactivate an unhealthy host before the request gate opens. Make lifecycle capability publication depend on the same readiness boundary, or start capability-bearing heartbeats only after the gate opens.

Low/test gap: heartbeat tests do not exercise `files_enabled=true` end-to-end, so the branch that appends `can_read_files`/`can_write_files` is currently untested. Add explicit healthy-files-enabled and healthy-files-disabled heartbeat cases.

Low/repo hygiene: the PR description still says `SS-430`, which violates `AGENTS.md`'s explicit rule against internal ticket IDs in public PR descriptions. Replace that with a plain description of the synchronization behavior.

Spec-level caveat, not counted as PR noncompliance: the ticket defines successful proxy `/health` as the source for `can_proxy_traffic`, but that endpoint does not prove the proxy's VMD resolver/local HTTP dependency can resolve sandboxes. A proxy can therefore report `can_proxy_traffic=true` while all actual sandbox routing fails. That is a weakness in the chosen SS-390 signal rather than a deviation by this patch.

internal/api/hosts.go
```diff
@@ -90,6 +90,14 @@ func (h *Handlers) HostHeartbeat(c *gin.Context) {
 		return
 	}
 
+	log.Debug().
```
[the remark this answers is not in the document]
awille-ai: Removed the internal ticket reference from the squashed commit message.

internal/vm/heartbeat.go
awille-ai: Addressed in 4ebaed9c: lifecycle capabilities are now published only after the existing VMD readiness gate reports ready; the heartbeat still runs before readiness for liveness.

internal/vm/heartbeat_test.go
awille-ai: Addressed in 4ebaed9c: heartbeat coverage now exercises files_enabled=true and the files-disabled path.

awille-ai:
Addressed the review feedback in 4ebaed9c: lifecycle capability publication is readiness-gated, files-enabled/disabled heartbeat coverage was added, and the PR description no longer contains internal ticket references.

awille-ai:
@codex review

internal/vm/heartbeat.go
```diff
@@ -165,12 +183,24 @@ func (c *heartbeatStorageCache) store(measurements []heartbeatStorageMeasurement
...
 			Dur("duration", time.Since(started)).
 			Msg("proxy capability probe failed; advertising no preview capabilities")
-		capabilities = nil
+		proxyState = proxyHealthState{}
+	}
+	capabilities := make([]string, 0, 7)
+	if cfg.LifecycleReady == nil || cfg.LifecycleReady() {
+		capabilities = append(capabilities, capabilityCanCreate, capabilityCanResume, capabilityCanPause, capabilityCanDestroy)
+	}
+	if err == nil {
```
awille-ai, line 374: **Low — make the readiness gate fail closed.** `LifecycleReady == nil` currently advertises all lifecycle capabilities. Production wires the callback today, but this is a correctness/safety boundary: any future caller that omits the newly added field silently regresses to unconditional `can_create/can_resume/can_pause/can_destroy`. Prefer `nil => false`, or make the readiness signal required by the API so omission cannot produce a false attestation.

internal/vm/heartbeat.go
```diff
@@ -165,12 +183,24 @@ func (c *heartbeatStorageCache) store(measurements []heartbeatStorageMeasurement
...
+	proxyState, err := proxyHealthCapabilities(ctx, client, proxyHealthURL)
 	if err != nil {
 		log.Warn().Err(err).
+			Str("host_id", cfg.HostID).
 			Dur("duration", time.Since(started)).
 			Msg("proxy capability probe failed; advertising no preview capabilities")
-		capabilities = nil
+		proxyState = proxyHealthState{}
+	}
+	capabilities := make([]string, 0, 7)
```
awille-ai, line 368: **Low — this warning now understates the failure.** A failed proxy probe no longer suppresses only preview attestations; it also removes `can_proxy_traffic`, `can_read_files`, and `can_write_files`. Please describe the broader operational-capability impact so this log is useful when diagnosing degraded hosts.

awille-ai (commented):
Second hostile review against the current SS-390 head. The prior lifecycle-readiness High, file-capability test gap, and public-artifact hygiene issue are addressed. CI and Terraform checks are green.

One High correctness/resilience issue remains: the initial heartbeat can publish `can_proxy_traffic` / file capabilities before VMD's local resolver is serving. The heartbeat service starts before the local HTTP resolver; `StartHeartbeat` sends immediately; a separately running proxy can return 200 even though its VMD resolver still cannot reach localhost:9090. That creates a false operational attestation for up to the next heartbeat interval. The existing open inline P2 thread identifies this same root cause.

Low maintainability/safety issue: `LifecycleReady == nil` currently means ready. Production wires it correctly today, but this is a safety gate whose zero value fails open. A future/test caller that constructs `HeartbeatConfig` without the callback silently advertises lifecycle capabilities. Prefer nil => not ready, or make readiness non-optional in the heartbeat API.

Low observability issue: on proxy probe failure the warning still says only `advertising no preview capabilities`, but the failure now suppresses `can_proxy_traffic`, `can_read_files`, and `can_write_files` as well. Update the message so operators understand the operational capability impact.

Requirement coverage: R1 satisfied; R2 satisfied; R3 partial because proxy/file capability truth can be false during startup; R4 satisfied; R5 satisfied; R6 satisfied; R7 satisfied; R8 satisfied; R9 satisfied; R10 partial for the same resolver-startup false-positive window.

internal/vm/heartbeat.go
```diff
@@ -165,12 +183,24 @@ func (c *heartbeatStorageCache) store(measurements []heartbeatStorageMeasurement
...
+	if cfg.LifecycleReady == nil || cfg.LifecycleReady() {
+		capabilities = append(capabilities, capabilityCanCreate, capabilityCanResume, capabilityCanPause, capabilityCanDestroy)
+	}
+	if err == nil {
+		capabilities = append(capabilities, capabilityCanProxyTraffic)
+		if proxyState.FilesEnabled {
+			capabilities = append(capabilities, capabilityCanReadFiles, capabilityCanWriteFiles)
```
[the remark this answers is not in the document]
awille-ai: Addressed in 8de35415: proxy, file, preview, and lifecycle capabilities are now gated until both VMD readiness and the local resolver readiness signal are true.

internal/vm/heartbeat.go
awille-ai: Addressed in 8de35415: a missing readiness callback now fails closed, so no operational capabilities are advertised by default.

internal/vm/heartbeat.go
awille-ai: Addressed in 8de35415: the warning now states that proxy and file operational capabilities are suppressed when the health probe fails.

awille-ai:
Addressed the latest review feedback in 8de35415: all operational capability publication is gated on VMD and local resolver readiness, the readiness callback fails closed, and the probe-failure warning reflects the broader capability impact.

awille-ai:
@codex review

awille-ai (commented):
Re-review of current head `8de35415` against SS-390. The previous resolver-startup race, fail-open readiness default, and misleading warning are addressed, and I resolved those threads. CI and Terraform are green.

One High correctness issue remains: the resolver readiness fix now conflates lifecycle readiness with data-plane resolver readiness. SS-390 explicitly defines `can_create`, `can_resume`, `can_pause`, and `can_destroy` from the initialized/running lifecycle service, while `can_proxy_traffic`/file capabilities use proxy/data-plane signals. Production now sets `LifecycleReady` to `startupReady && localHTTPReady`, so if the local resolver exits/fails after startup, the next heartbeat removes all four lifecycle capabilities even though the gRPC lifecycle service can still be healthy. Keep separate lifecycle and resolver/data-plane readiness signals: gate lifecycle capabilities on `startupReady`; gate proxy/file/preview capabilities on `startupReady && localHTTPReady` (plus proxy health/files state).

No additional blocking findings found in this pass.

awille-ai:
Addressed in 0936df68: lifecycle capabilities now use startupReady alone, while proxy, file, and preview capabilities require both startupReady and local resolver readiness. Added regression coverage for this separation.

awille-ai:
@codex review

internal/vm/heartbeat.go
```diff
@@ -165,12 +185,26 @@ func (c *heartbeatStorageCache) store(measurements []heartbeatStorageMeasurement
...
+	lifecycleReady := cfg.LifecycleReady != nil && cfg.LifecycleReady()
+	resolverReady := cfg.ResolverReady != nil && cfg.ResolverReady()
+	if lifecycleReady {
+		capabilities = append(capabilities, capabilityCanCreate, capabilityCanResume, capabilityCanPause, capabilityCanDestroy)
+	}
+	if resolverReady && err == nil {
+		capabilities = append(capabilities, capabilityCanProxyTraffic)
+		if proxyState.FilesEnabled {
+			capabilities = append(capabilities, capabilityCanReadFiles, capabilityCanWriteFiles)
+		}
```
awille-ai, line 380: **High — preserve preview feature semantics separately from resolver availability.** This append is inside the `resolverReady` block, so a resolver-only outage removes `preview_ports_v1` / access / token / browser-auth attestations even though those values describe what protocol the healthy proxy implements. SS-390 explicitly keeps those feature/version capabilities distinct from `can_proxy_traffic` operational availability and requires their existing semantics to be preserved. Keep `can_proxy_traffic` and file capabilities gated on resolver readiness, but publish the validated `proxyState.PreviewCapabilities` whenever the proxy `/health` probe succeeds. This is not metadata-only today: create placement still requires `preview_ports_v1`, so suppressing it also changes admission behavior during a resolver-only outage.

awille-ai (commented):
Hostile re-review against SS-390: the prior lifecycle/resolver over-coupling is fixed, but one High semantic issue remains.

High: existing `preview_*` protocol/version attestations are now suppressed whenever `ResolverReady` is false. SS-390 explicitly defines these as feature/protocol attestations that remain distinct from operational availability (`can_proxy_traffic`) and requires their existing semantics to be preserved. Before this PR they were derived from the proxy `/health` capability list; resolver readiness was not part of their meaning. Keep `can_proxy_traffic` and file operational capabilities behind resolver readiness, but append the validated `proxyState.PreviewCapabilities` whenever the proxy health probe succeeds. This matters today because create placement still requires `preview_ports_v1`, so coupling that feature bit to resolver liveness also turns a resolver-only outage into implicit create-admission enforcement that SS-390 does not own.

The previous findings (lifecycle startup race, resolver startup race, fail-open readiness default, file capability coverage, logging, and public ticket-reference hygiene) are all addressed on the current head. The current CI and Terraform runs were still in progress at the time of this review.

internal/vm/heartbeat.go
awille-ai: Addressed in e58e321a: preview protocol attestations are now preserved whenever the proxy health probe succeeds, independent of resolver readiness; only operational proxy/file capabilities require resolver readiness.

internal/vm/heartbeat.go
```diff
@@ -165,12 +185,26 @@ func (c *heartbeatStorageCache) store(measurements []heartbeatStorageMeasurement
...
+		proxyState = proxyHealthState{}
+	}
+	capabilities := make([]string, 0, 7)
+	lifecycleReady := cfg.LifecycleReady != nil && cfg.LifecycleReady()
+	resolverReady := cfg.ResolverReady != nil && cfg.ResolverReady()
+	if lifecycleReady {
+		capabilities = append(capabilities, capabilityCanCreate, capabilityCanResume, capabilityCanPause, capabilityCanDestroy)
+	}
+	if resolverReady && err == nil {
+		capabilities = append(capabilities, capabilityCanProxyTraffic)
```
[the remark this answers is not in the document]
awille-ai: Addressed in e58e321a: the proxy health response now verifies its configured VMD resolver endpoint, and operational proxy/file capabilities require that resolver probe to succeed.

awille-ai:
Addressed the latest review feedback in e58e321a: preview feature attestations remain independent from resolver readiness, while operational proxy/file capabilities require a successful probe of the proxy’s configured VMD resolver.

awille-ai:
@codex review

meAmitPatil:
Reviewed `e27cf903`: one P2 remains — `VMDResolver.Ready` accepts any HTTP 404, so a wrong service can falsely publish proxy/file capabilities; the synthetic `/instances` probe also enters lazy reattach/BoltDB on every LB health check.
Please use a dedicated VMD readiness endpoint with an exact service marker and tests for real VMD, arbitrary 404, and timeout. Lifecycle paths otherwise gain no synchronous work or new create-burst serialization.

internal/vm/heartbeat.go
```diff
@@ -165,12 +187,28 @@ func (c *heartbeatStorageCache) store(measurements []heartbeatStorageMeasurement
...
+		proxyState = proxyHealthState{}
+	}
+	capabilities := make([]string, 0, 7)
+	lifecycleReady := cfg.LifecycleReady != nil && cfg.LifecycleReady()
+	resolverReady := cfg.ResolverReady != nil && cfg.ResolverReady()
+	if lifecycleReady {
+		capabilities = append(capabilities, capabilityCanCreate, capabilityCanResume, capabilityCanPause, capabilityCanDestroy)
+	}
+	if err == nil {
+		capabilities = append(capabilities, proxyState.PreviewCapabilities...)
```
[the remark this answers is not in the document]
awille-ai: Already fixed in e27cf903: operational capabilities are appended before preview protocol capabilities, matching the expected payload order. The subsequent CI run passes.

awille-ai:
@codex review

internal/vm/heartbeat.go
```diff
@@ -336,12 +187,28 @@ func (c *heartbeatStorageCache) store(measurements []heartbeatStorageMeasurement
...
+		capabilities = append(capabilities, capabilityCanCreate, capabilityCanResume, capabilityCanPause, capabilityCanDestroy)
+	}
+	if err == nil {
+		if resolverReady && proxyState.ResolverReady {
+			capabilities = append(capabilities, capabilityCanProxyTraffic)
+			if proxyState.FilesEnabled {
+				capabilities = append(capabilities, capabilityCanReadFiles, capabilityCanWriteFiles)
+			}
+		}
+		capabilities = append(capabilities, proxyState.PreviewCapabilities...)
```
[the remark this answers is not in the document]
awille-ai: This conflicts with the SS-390 contract we explicitly settled on. `preview_*` values are protocol/version attestations, not runtime availability bits; resolver/data-plane availability is represented separately by `can_proxy_traffic`. Gating `preview_*` on resolver readiness would also change current create admission because placement consumes `preview_ports_v1`, which SS-390 deliberately does not own. The current code is correct to preserve preview attestations when proxy `/health` succeeds even if resolver readiness is false.

awille-ai (commented):
PR-review verdict: APPROVE (GitHub does not allow this account to formally approve its own PR).

Reviewed current head against SS-390 and the decisions from the prior review loop.

Requirement coverage:
- R1/R2 host identity + metadata: satisfied; existing heartbeat/self-registration path is preserved.
- R3 operational capability vocabulary: satisfied; create/resume/pause/destroy/proxy/read/write are published and no `can_stop`/`can_preview_url` is introduced.
- R4 status vs technical capability separation: satisfied.
- R5 heartbeat-generation / non-destructive capability sync: satisfied; existing persistence path remains intact.
- R6 cadence/freshness: satisfied; heartbeat cadence is unchanged.
- R7 stale/non-active fencing: satisfied / unchanged by this PR.
- R8 capacity/pressure separation: satisfied; lifecycle capability values do not depend on schedulable capacity.
- R9 preview protocol attestations: satisfied; `preview_*` remains a feature/protocol contract and is not suppressed by resolver-only unavailability.
- R10 authoritative signals: satisfied for the intended behavior; lifecycle uses VMD readiness, proxy/file operational capability requires healthy proxy plus resolver availability, and files use the existing `filesEnabled` state.

Reviewer execution (sequential fallback; no subagent runtime available):
- Requirements: PASS
- Hostile/adversarial: PASS
- Security/resilience: PASS
- Test quality: PASS with a non-blocking coverage note: a focused regression test for preview attestations surviving resolver-unready state would make the R9 boundary more explicit.
- Correctness/maintainability: PASS
- Observability: PASS
- Compatibility/rollout: PASS
- Data correctness: PASS

The remaining open Codex thread asking to gate `preview_*` on resolver readiness conflicts with SS-390's explicit feature-vs-operational separation, so I replied there rather than treating it as a defect. Both CI and Terraform Checks pass on this head.

meAmitPatil (approved):

PR #456, opened 2026-08-27
infra: re-enable backup_coverage_alerts now that the metric is confirmed live

pavitrabhalla:
## Summary
- #409 shipped the coverage sampler + disabled-by-default `backup_coverage_alerts` policies. Applying it deadlocked Terraform CD: the alert_policy resources require `backup_uncovered_paused_sandboxes` to already exist as a Cloud Monitoring metric type, but the controlplane code that emits it deploys through this same gated apply.
- #443 nulled `backup_coverage_alerts` in all three environments to break the deadlock and let the emitting code ship, leaving the original blocks commented out in place.
- This PR uncomments them now that the metric is confirmed live: verified fresh, continuous data points (15s cadence) in Cloud Monitoring for staging (rayai-dev) and both production cells (rayai-prod, us-east4 and usw2) before restoring.

- Values restored unchanged from the original #409 config (`enabled = false`, same display_prefix/regions) — this is a pure re-enable, not a config change. Policies stay disabled by default; flipping them on is a separate step gated on the cell's backfill backlog converging to zero.

## Testing
- `terraform validate` clean on all three environments.
- Confirmed via the Cloud Monitoring API that `backup_uncovered_paused_sandboxes` has recent points in all three projects/regions before restoring the alert_policy blocks that reference it.
- Local Codex review: clean, no findings.

### DIFF
--- a/infra/envs/production/us-east4/main.tf
+++ b/infra/envs/production/us-east4/main.tf
@@ -380,21 +380,11 @@ module "observability" {
   # under us-central1 labeling before the host swap, so the legacy
   # value stays scoped: host rows relabeled or restored under it must
   # not fall outside the alert.
-  #
-  # Left null for now (was the object below): creating these alert_policy
-  # resources requires backup_uncovered_paused_sandboxes to already exist
-  # as a Cloud Monitoring metric type, but the controlplane code that
-  # emits it deploys through this same Terraform CD run's api job, which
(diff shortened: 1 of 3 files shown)


pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

meAmitPatil (approved):

PR #458, opened 2026-08-28
vmd: move backup staging onto a dedicated background-data disk

pavitrabhalla:
## Summary
Staged pause generations currently live on the local-SSD array that also serves live VM disk I/O (`/mnt/localssd`). The uploader reads each staged file twice before it leaves the host (digest pre-check, then the upload stream), which puts drain traffic in direct contention with tenant workloads — the sharpest remaining risk in the backup pipeline as upload concurrency rises. This attaches a dedicated persistent disk per production vmd host (staging already has one, reused here) and moves that read traffic off the local-SSD array. Part of the sandbox recovery/durability effort.

## Technical decisions
- **Split staging into two roots.** The pause RPC path keeps its own always-local tree (`pauseStagingRoot`, unconditionally on `SNAPSHOT_DIR`'s filesystem) so reflink and the base-pin hard link it creates never have to cross filesystems on the synchronous pause path — moving that copy cross-device would have added real I/O to a sandbox-lifecycle hot path. A detached worker promotes each finished generation into the configurable upload-visible tree (`backupStaging`, `BACKUP_STAGING_DIR`) before enqueueing; on hosts with no override configured this collapses to the existing single-root behavior (the promotion step is then a no-op existence check, not a copy).
- **Promotion doesn't delete the local copy inline.** The journal dedupes same-generation enqueues by content and won't adopt a new path onto an already-`Staged` row, so a successful enqueue isn't proof the journal points at the promoted copy rather than the local one. Cleanup is left entirely to the existing staging sweep, which only reclaims a generation once the journal has no pending row for it at all — correct regardless of which path a row ends up referencing.
- **Disk type is `hyperdisk-balanced`, not `pd-balanced`.** These are Z3 metal hosts, which reject standard Persistent Disk types the same way they already reject the API-default boot disk type.
- **Sized at 1024GB** for real headroom rather than steady-state drain: peak pause rate runs on the order of 1.3k generations/hour at ~12MB packed each, so a fully stalled uploader fills roughly 15.6GB/hour — 1024GB absorbs ~65 hours (~2.7 days) of that before the disk itself becomes the constraint.
- **Every host that can carry the `vmd` deploy label gets a disk**, including the us-west2 cold standby: the deploy step (and its `BACKUP_STAGING_DIR` config) follows `active_sandbox_host`, and the new mount precondition in `deploy-vmd.py` would otherwise block a deploy to a freshly-promoted standby.
- **No `deletion_policy` on the attachment resources**: unsupported under this repo's pinned `~> 6.0` provider constraint for production (confirmed against the provider schema directly, not staging's `~> 7.0` block). The disk resources' own `prevent_destroy` is what actually protects the data; the attachments themselves are deliberately left replaceable so the documented host-recreation flow still works.
- **Production env wiring is deferred.** Terraform only attaches the raw disks — formatting and mounting is a manual step, same as the existing local-SSD array (documented in `host-dr-runbook.md`). `BACKUP_STAGING_DIR` is left set on staging only (whose disk is already mounted and proven for the journal path) until an operator confirms `/mnt/sandbox-data` is mounted on use4 and usw2; enabling it before that would trip the new deploy-time mount precondition and block every subsequent deploy to that host.

## Testing
- `CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ./...` and `go vet ./...` clean.
- Full `go test ./...` (via a Linux container, since this repo builds Linux-only packages) passes, including new coverage for cross-root promotion and the journal-dedup interaction with local-copy cleanup.
- `go test -race ./internal/backup/... ./internal/vm/... ./cmd/vmd/...` clean.
- `terraform fmt -check -recursive` and `terraform init -backend=false && terraform validate` clean in `infra/envs/production/us-east4`, `infra/envs/production/us-west2`, and `infra/envs/staging/us-central1`.
- Local `codex exec review --base main` iterated to a clean pass (no findings) across several rounds.


pavitrabhalla:
@codex review

meAmitPatil (approved):

awille-ai (commented):
## Verdict
COMMENT

One High rollout/compatibility issue remains. The cross-root promotion/data-durability design itself looks sound, and the Terraform/CI checks are green, but `BACKUP_STAGING_DIR` is sticky in `/etc/sandbox/vmd.env`: the deploy script only removes/re-appends the line when the incoming value is non-empty. If staging needs to roll this feature back by removing `BACKUP_STAGING_DIR` from the workflow, the old value remains on the host.

That is especially risky here because the pre-PR vmd already understands `BACKUP_STAGING_DIR` as the *single* staging root. So a code/workflow rollback can restart the old binary with the stale dedicated-PD path and move inline pause staging onto the slower PD—the exact lifecycle regression this PR deliberately avoids. If the mount is unavailable, the rollback deploy also skips the new mount precondition because its incoming value is empty, while the stale env value can make vmd recreate the path on the root filesystem.

**High — make an unset `BACKUP_STAGING_DIR` actively clear stale host config.** The deploy path should have explicit disable semantics (for example, always remove the existing line, then append only when non-empty), and this rollback case should be covered by the deploy-script tests.

## Requirement coverage (SS-382)
- R1 — keep synchronous pause staging on the snapshot/NVMe filesystem when moving uploader-visible staging: **satisfied**.
- R2 — asynchronously promote completed generations to the dedicated uploader-visible disk without losing in-flight journal coverage: **satisfied**.
- R3 — provision a dedicated background-data disk per production host, including west standby: **satisfied** by Terraform.
- R4 — validate/mount before enabling the dedicated staging root: **satisfied** for staging and enforced by the deploy precondition when the setting is present.
- R5 — activate `BACKUP_STAGING_DIR` on production hosts: **partial by design**. This PR attaches the disks but intentionally leaves production wiring disabled pending manual format/mount + staging validation. I do not consider that itself a merge blocker for a staged rollout, but SS-382 should remain open until production activation is completed.
- R6 — protect pause latency: **satisfied by design**; the cross-filesystem copy runs in the detached worker, not on the pause RPC path.

## Reviewer execution
| --- | --- | --- |
| Requirements | sequential fallback | FINDINGS |
| Test quality | sequential fallback | FINDINGS |
| Correctness | sequential fallback | FINDINGS |
| Infrastructure plan safety | sequential fallback | PASS with caveat |

## Validation
- Inspected the full PR diff and surrounding backup staging/journal code, including `StagePending`, `FinishPendingStage`, `StageTask`, shared-base staging, journal dedupe behavior, periodic sweeps, and the new cross-root tests.
- Verified the pre-PR binary already reads `BACKUP_STAGING_DIR`, which makes stale env state a real rollback hazard rather than a theoretical configuration-cleanliness issue.
- Traced promotion failure, enqueue failure/retry, dedup against an already-Staged generation, local-copy cleanup, restart recovery, and mount-loss/rollback behavior.
- Current CI, Terraform Checks, and Terraform Plans all complete successfully on head `1cf907e9`.
- Terraform plan jobs ran successfully for production use4, production usw2, staging, and the other configured roots. The connector exposes the successful plan jobs but not the uploaded human-readable plan summaries, so I could not independently inspect every planned resource action; the changed Terraform itself was reviewed directly.

pavitrabhalla:
@codex review

PR #459, opened 2026-08-28
vm/fc: pin the swagger generator and resync the vendored Firecracker spec

meAmitPatil:
The vendored spec had drifted from the one we build against and the client generator was unpinned, so regenerating dropped a field production code uses. This resyncs the spec to a verbatim copy, pins the generator to the newest version whose output compiles against the go-openapi/runtime already in go.mod, and regenerates.

No wire change: JSON tags are identical apart from one addition nothing sets yet, and regeneration is byte-idempotent. Most files gain only a generator banner — the real changes are the spec, generate.go, and four generated files.

Two optional booleans now generate as `*bool`, so both call sites route through a helper keeping them omitted when false; these params use deny_unknown_fields, where an explicit false would break rollback.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
## Verdict
APPROVE

I did not find any remaining Critical or High issues in the current head (`1c282972`). The substantive generated-client changes are consistent with the Firecracker wire contract, and the regeneration path is now pinned and CI-enforced.

## Review scope
Generic review (no explicit ticket/spec was linked in the PR). Sequential passes covered hostile/adversarial behavior, security/resilience, test quality, correctness/maintainability, compatibility/rollout, and sandbox lifecycle performance.

## Validation
- Inspected the full PR diff and the meaningful generated-model changes rather than treating the generator banner churn as signal.
- Verified `SnapshotCreateParams.Flatten` changes from `bool` to `*bool`, and the only runtime caller now uses `omitFalse`, preserving the pre-field wire format when disabled.
- Verified `MemoryBackend.AbortOnHandlerDeath` changes from `bool` to `*bool`, and its runtime caller likewise omits false, so older Firecracker binaries with `deny_unknown_fields` continue to accept requests when the feature is off.
- Verified `SnapshotLoadParams.ClockRealtime` is `*bool`, preserving the intended three states (omitted / false / true) needed by the newer Firecracker contract.
- Checked the vendored spec changes: the added `flatten` and `clock_realtime` definitions match the generated types and the existing runtime behavior; the remaining spec edits are descriptive clarifications.
- Checked generator pinning: `go generate` uses `go-swagger@v0.33.1` explicitly rather than an unversioned tool.
- Checked the CI guard: it runs `go generate ./internal/vm/fc/...` and uses `git status --porcelain`, so it catches both modified generated files and newly generated untracked files. The earlier review thread about untracked output is resolved in the current head.
- Current CI is green, including Build, Vet, Integration Tests, and the new `FC Client Check` regeneration job.
- No new network/RPC/filesystem work is added to sandbox create/resume/pause hot paths; the runtime change is only pointer construction/JSON omission semantics.

## Reviewer execution
| --- | --- | --- |

The main residual risk is ordinary generated-code churn: future spec changes can still alter public Go identifiers while preserving JSON tags, but build/tests plus the deterministic regeneration check are the appropriate controls for this internal client package.

PR #460, opened 2026-08-28
controlplane: evaluate capacity ranking in shadow alongside real placement

meAmitPatil:
Groundwork for capacity-based placement: ranks hosts by the pressure they publish and reports what it **would** have chosen, without influencing any create.

Supersedes #445, which built the same ranking on a Postgres reservation fence. That fence cost a synchronous database round trip per create and serialized bursts on one `host_pressure` row. Enforcement is moving into vmd instead — an in-memory gate inside the existing create RPC, where the check is a mutex rather than a network round trip. This PR is the half with no enforcement in it.

**Why shadow, not a flag that places.** Ranking alone cannot enforce a limit: two control-plane replicas can still exceed a host's cap between reports. So there is deliberately no configuration in which this decides placement.

**Cost to a create:** one bounded, non-blocking channel send. Ranking, queries and metrics all happen on a background worker; a full queue drops rather than waits. `TestShadowOfferNeverBlocksARequest` proves this against a deliberately wedged evaluator. The queue is small (64) because the queue is the sampler.

**What it reports:** whether ranking's first choice agrees with what the live scheduler actually did, plus fleet composition — described, under-described (reporting VMs they could not size, so their allocation is an undercount), stale, legacy. That last is the readiness signal for enforcement.

**Ranking:** dominant-resource scoring across memory, vCPUs, slots and sandbox count; hosts at their operator limit dropped; fully-described hosts ahead of under-described ones; comparable hosts shuffled with warm-slot hosts first. `AllowedHosts` is the template-locality hook, producer separate.

**Not included:** no migration, reservation table, lifecycle hooks, `RSV` column, or placement change. The candidate query reads only existing columns.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

pavitrabhalla (approved):

meAmitPatil:
@codex review

meAmitPatil:
@codex review

PR #463, opened 2026-08-31
vmd: hold fleet-sized background work until the daemon is ready to serve

meAmitPatil:
Slot refill, slot adoption, and the full reattach each walk the whole inventory, and every netlink operation they issue takes the kernel's single global RTNL lock — so running them during startup delays the readiness they compete with, and their per-record logs can push the readiness line past journald's rate limit.

A `postReady` barrier now holds all three until readiness is announced, released after the readiness line is logged. Requests can't be served during the wait (the gRPC interceptors reject everything with `Unavailable` until `startupReady` flips), so nothing is delayed that a caller was waiting on. Behind the gate the pool no longer declares itself as producing, so an early create takes the bounded inline path instead of stalling on a worker that hasn't started.

Also resolves the host interface once instead of twice, and off the startup goroutine: both advertised endpoints derive from the same lookup, which dumps interface and address tables that grow with the fleet.

### DIFF
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -1455,9 +1455,8 @@ func main() {
 	go func() {
 		defer sentrylog.Recover("startup reattach")
 		// Held until readiness: requests load their VM on demand, so nothing
-		// waits on this pass, while running it during startup both contends
-		// for RTNL and emits one line per record — enough, on a large host,
-		// to push the readiness line past journald's rate limit.
+		// waits on this pass (see the fleet-sized background work comment
+		// above for why it must not run during startup).
 		select {
 		case <-postReady:
 		case <-ctx.Done():
@@ -1748,11 +1747,9 @@ func main() {
 	startupReady.Store(true)
 	st.mark("ready", true, -1)
 	log.Info().Msg("startup complete — gRPC serving requests")
-	// Released only after the readiness line is emitted: the work behind it
-	// logs per record, and on a large host that burst is enough to exhaust
-	// journald's rate limit. Anything published after this point can be
-	// dropped; the readiness line cannot, because the deploy readiness check
-	// reads it from the journal.
+	// Released only after the readiness line is emitted: the deploy
+	// readiness check reads that line from the journal, so nothing the
+	// background work logs may come ahead of it.
 	close(postReady)
 
 	// Leak gauge for network namespaces — independent of the launcher path.
--- a/cmd/vmd/main.go
+++ b/cmd/vmd/main.go
@@ -213,21 +213,38 @@ func publishesCapacityPressure(advertiseAddr, controlPlaneURL string) bool {
 	return advertiseAddr != "" && controlPlaneURL != ""
 }
 
-func advertisedVMDAddr(interfaceName string, grpcPort int, explicit string) (string, error) {
+// hostIPOnce memoizes hostInterfaceAddress. Both advertised endpoints derive
+// from the same interface, and the lookup dumps the host's interface and
+// address tables — tables that grow with every VM and pooled slot on the box
+// — so it runs at most once per process, and not at all when both endpoints
+// are configured explicitly.
+func hostIPOnce(interfaceName string) func() (string, error) {
+	var (
+		once sync.Once
+		ip   string
+		err  error
+	)
+	return func() (string, error) {
+		once.Do(func() { ip, err = hostInterfaceAddress(interfaceName) })
+		return ip, err
+	}
+}
+
+func advertisedVMDAddr(hostIP func() (string, error), grpcPort int, explicit string) (string, error) {
 	if explicit != "" {
 		if _, _, err := net.SplitHostPort(explicit); err != nil {
 			return "", err
 		}
 		return explicit, nil
 	}
-	ip, err := hostInterfaceAddress(interfaceName)
+	ip, err := hostIP()
 	if err != nil {
 		return "", err
 	}
 	return net.JoinHostPort(ip, strconv.Itoa(grpcPort)), nil
 }
 
-func advertisedProxyAddr(interfaceName, proxyHealthURL, explicit string) (string, error) {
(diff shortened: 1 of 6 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
## Verdict
APPROVE

I did not find any remaining Critical or High defect in the current head (`7130c807`). The change safely moves fleet-sized refill/adoption/reattach work behind readiness, preserves slot ownership/fail-closed behavior during the new post-ready race window, and bounds/ramp-limits the RTNL-heavy background producers once serving begins.

## Requirement coverage — SS-362
- R1 — move non-critical fleet-sized reconciliation off the startup critical path where correctness permits: **satisfied for network refill/adoption and full VM reattach**.
- R2 — preserve crash-safety and never reuse resources belonging to a live/uncertain VM: **satisfied**. The authoritative allocator checks ownership and kernel namespace presence independently of adoption, so a create immediately after readiness cannot claim an abandoned/live slot merely because the gated adoption worker has not published itself yet.
- R3 — keep post-ready background recovery bounded so it does not stampede the host: **satisfied for refill/adoption** via a shared gradually-ramped background token budget plus bounded foreground yielding. Full reattach remains outside that semaphore, but it was already designed as background/on-demand-safe and this PR does not make its individual operations more aggressive.
- R4 — requests must remain correct while proactive reattach is incomplete: **satisfied** by the existing on-demand VM/state loading path; the full pass is proactive population/GC rather than a prerequisite for lifecycle RPC correctness.
- R5 — measure/document startup across 1k/5k/10k/20k+ and establish an explicit recovery SLO: **not satisfied by this PR**. I consider this remaining SS-362 project work, not a blocker for this isolated optimization.

## Review scope / validation
- Traced the readiness ordering: interceptors stay closed until `startupReady`, the readiness log is emitted, then `postReady` releases the gated workers.
- Traced the narrow race where requests can start between `startupReady.Store(true)` and gated producer activation. Empty-pool callers intentionally use the bounded inline path; `claimSlotIndex` independently skips owned/kernel-present indices, so adoption lag cannot cause slot reuse or collision with abandoned state.
- Reviewed `StartGate`, adoption pending/scanning transitions, producer trust, `ClaimWait`, shutdown-before-ready behavior, and duplicate adoption starts.
- Reviewed the shared background concurrency ramp and `withBGSlot` token lifecycle, including panic return and shutdown/cancellation paths.
- Reviewed foreground priority accounting for create/resume and the anti-priority-inversion exception while `ClaimWait` is actively waiting on a producer.
- Reviewed host-interface resolution memoization and the move into the heartbeat goroutine; explicit advertise addresses continue to avoid the fleet-sized lookup entirely.
- Rechecked the prior P2 test finding. The current test initializes `mgr`, verifies the intentional `panic("boom")`, and confirms the token is returned, so that thread is correctly resolved.

## Reviewer execution
| --- | --- | --- |
| Requirements | sequential fallback | FINDINGS (ticket remains broader than PR) |

The main follow-up is measurement rather than a code correction: keep SS-362 open until restart performance is actually measured at the ticket's target cardinalities and an explicit recovery SLO is documented.

awille-ai (approved):
## Verdict
APPROVE

I did not find any remaining Critical or High defect in the current head (`7130c807`). The change safely moves fleet-sized refill/adoption/reattach work behind readiness, preserves slot ownership/fail-closed behavior during the new post-ready race window, and bounds/ramp-limits the RTNL-heavy background producers once serving begins.

## Requirement coverage — SS-362
- R1 — move non-critical fleet-sized reconciliation off the startup critical path where correctness permits: **satisfied for network refill/adoption and full VM reattach**.
- R2 — preserve crash-safety and never reuse resources belonging to a live/uncertain VM: **satisfied**. The authoritative allocator checks ownership and kernel namespace presence independently of adoption, so a create immediately after readiness cannot claim an abandoned/live slot merely because the gated adoption worker has not published itself yet.
- R3 — keep post-ready background recovery bounded so it does not stampede the host: **satisfied for refill/adoption** via a shared gradually-ramped background token budget plus bounded foreground yielding. Full reattach remains outside that semaphore, but it was already designed as background/on-demand-safe and this PR does not make its individual operations more aggressive.
- R4 — requests must remain correct while proactive reattach is incomplete: **satisfied** by the existing on-demand VM/state loading path; the full pass is proactive population/GC rather than a prerequisite for lifecycle RPC correctness.
- R5 — measure/document startup across 1k/5k/10k/20k+ and establish an explicit recovery SLO: **not satisfied by this PR**. I consider this remaining SS-362 project work, not a blocker for this isolated optimization.

## Review scope / validation
- Traced the readiness ordering: interceptors stay closed until `startupReady`, the readiness log is emitted, then `postReady` releases the gated workers.
- Traced the narrow race where requests can start between `startupReady.Store(true)` and gated producer activation. Empty-pool callers intentionally use the bounded inline path; `claimSlotIndex` independently skips owned/kernel-present indices, so adoption lag cannot cause slot reuse or collision with abandoned state.
- Reviewed `StartGate`, adoption pending/scanning transitions, producer trust, `ClaimWait`, shutdown-before-ready behavior, and duplicate adoption starts.
- Reviewed the shared background concurrency ramp and `withBGSlot` token lifecycle, including panic return and shutdown/cancellation paths.
- Reviewed foreground priority accounting for create/resume and the anti-priority-inversion exception while `ClaimWait` is actively waiting on a producer.
- Reviewed host-interface resolution memoization and the move into the heartbeat goroutine; explicit advertise addresses continue to avoid the fleet-sized lookup entirely.
- Rechecked the prior P2 test finding. The current test initializes `mgr`, verifies the intentional `panic("boom")`, and confirms the token is returned, so that thread is correctly resolved.

## Reviewer execution
| --- | --- | --- |
| Requirements | sequential fallback | FINDINGS (ticket remains broader than PR) |

The main follow-up is measurement rather than a code correction: keep SS-362 open until restart performance is actually measured at the ticket's target cardinalities and an explicit recovery SLO is documented.

PR #464, opened 2026-09-01
ci: drop the unusable stock-binary precondition from the Firecracker deploy

meAmitPatil:
The deploy refused to run on a host that has no `firecracker.v1.15.0.bak`, which turns out to be the wrong precondition.

That file is pre-fork upstream Firecracker, kept as a "deep rollback floor". It cannot serve as one: it has no UFFD-internal backend, no layered restore and no flatten, so it refuses every snapshot the fleet holds — falling back to it means destroying sandboxes and re-seeding templates, which is a rebuild rather than a rollback. It is also absent on any host not hand-provisioned during the early fork work, since no bring-up runbook creates it.

The rollback that does work is the digest-named backup this script already writes and verifies immediately before each swap, plus redeploying an earlier release. Neither depends on that file, so the check only blocked deploys without adding safety.

Also traps the staged copy so an aborted run clears it, rather than leaving the uploaded binary in `/tmp` as the failed run did.

### DIFF
--- a/.github/workflows/scripts/deploy-firecracker.py
+++ b/.github/workflows/scripts/deploy-firecracker.py
@@ -25,9 +25,6 @@ import tempfile
 RELEASE_REPO = "superserve-ai/firecracker"
 
 INSTALL_PATH = "/usr/local/bin/firecracker"
-# The pre-fork stock binary. It is the floor of any rollback, so a host that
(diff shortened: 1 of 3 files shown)


meAmitPatil:
@codex review

pavitrabhalla (approved):

PR #471, opened 2026-09-03
boxd: wall-clock correction and workload freezer primitives

meAmitPatil:
Guest-side primitives for restoring a snapshot with its monotonic clock frozen: boxd can correct the guest's wall clock from the host's PTP clock, and can freeze its workload in a cgroup around a snapshot and release it on an explicit, token-carrying wake from the supervisor. Every command boxd spawns joins that cgroup and confirms it before the spawn completes, and no command starts while the workload is frozen.

Dormant: the cgroup is mounted, and announced to boxd, only by images built with the freezer switch on, which nothing sets yet. Without it boxd probes nothing at start, the exec path is unchanged, the new routes are never called, the PTP device is never opened, and health returns exactly the response it always has.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai:
@codex review

awille-ai (approved):
## Verdict
APPROVE

I did not find any remaining Critical or High issue in the current head (`972869bb`). This is a ticket-backed review against SS-460's first stacked PR: guest-side wall-clock correction + workload-freezer primitives, with the guest-init freezer block present but dormant because no current builder path enables `FreezeWorkload`.

## Requirement coverage — SS-460 / stacked PR 1
- R1 — provide a trusted guest-side realtime correction source using KVM PTP and fail closed when a frozen-clock restore cannot be corrected: **satisfied**. `/wake` only thaws after correction/verification succeeds when `clock_frozen=true`.
- R2 — keep customer workload stopped across the stale-realtime window and release it only by explicit supervisor protocol: **satisfied**. boxd-spawned commands join the cgroup before exec; freeze excludes concurrent spawns; `/health` never thaws; `/wake` and `/thaw` are token-bound and idempotent.
- R3 — bound pause-side freeze work and roll back to a running workload on a normal timeout/failure: **satisfied**. The context budget covers spawn-guard acquisition plus cgroup transition; confirmed rollback returns the workload to THAWED, while an unconfirmed rollback remains fail-closed and recoverable under the same token.
- R4 — preserve legacy behavior when the guest image does not opt into the freezer: **satisfied**. With `BOXD_WORKLOAD_FREEZER` absent, `/health` retains the old constant response and the spawn path takes no new lock/wrapper work.
- R5 — make the guest-init freezer mount opt-in and do not activate it in this PR: **satisfied**. `builder.Config.FreezeWorkload` defaults false and no production builder/deployment path in this PR sets it.
- R6 — expose enough validation for a later builder/activation PR to prove PTP settime + freezer readiness before shipping an image: **satisfied** via `/verify-clock`, health state, protocol metadata, and the gated init construction.

## Review scope / validation
- Re-ran the previously reported race/failure cases against the current code rather than assuming resolved threads were correct: freeze while spawn placement is in flight, budget expiry while waiting for the spawn guard, cgroup stuck in FREEZING, failed/unconfirmed thaw rollback, repeated freeze/thaw/wake tokens, malformed/trailing JSON, and blocked diagnostics.
- Traced process placement from `beginSpawn` through wrapper `cgroup.procs` join, fd-3 confirmation, and `endSpawn`, including Start failure and placement timeout/kill.
- Traced frozen restore ordering: token validation → PTP read/correction → correction verification → confirmed THAWED → successful wake response. A PTP/settime failure leaves the workload stopped.
- Checked `/health`: legacy images retain the old fast path exactly; freezer-enabled images report 503 while freezing/frozen without touching PTP or thawing anything.
- Checked rollback semantics: a freeze that cannot confirm its rollback retains `frozen=true` and the active token, so later `/thaw` can retry rather than losing control of a possibly stopped workload.
- Checked lifecycle logging: `diagf` is bounded/non-blocking, avoiding the earlier request-latency and unbounded-goroutine problems.
- Checked host-only routing guard and its fail-closed behavior. The interface-address cache is warmed asynchronously at boxd initialization; there remains a very small first-request race where the request may perform the same enumeration if warm-up has not completed, but I consider that Low/non-blocking rather than a correctness or rollout issue.
- Current CI and Terraform Checks are green.

## Reviewer execution
| --- | --- | --- |
| Requirements | sequential fallback | PASS |
| Performance/resource efficiency | sequential fallback | PASS with Low caveat |

The important activation boundary remains outside this PR, as intended: do not enable the image freezer or frozen-clock supervisor policy until the later manifest/deployment/vmd integration PRs and the staging end-to-end harness are in place.

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

PR #472, opened 2026-09-03
test: size the backfill-pool test's wait to the retry it waits for

meAmitPatil:
The pool-starvation test uses a two-second backoff so a held slot is visible, but the first retry after a failed probe is due at a doubled backoff plus jitter, four to six seconds later, and the wait for it was five. Only that wait changes; the default budget for millisecond retries stays.


meAmitPatil:
@codex review

meAmitPatil:
Folded into #476 as an identical commit, so it merges with that PR.

PR #474, opened 2026-09-03
resume: emit lookup, claim, vmd, post, and revert phases; skip the binding read when there were never bindings

meAmitPatil:
Resume now emits lookup, claim, vmd, post, and revert phases the way create does, so the control-plane share of resume latency is measurable per stage and a failed resume's compensation is reported on its own rather than inflating the stage it follows. On the explicit endpoint the stage phases partition total; on the auto-resume path total is unchanged and measures the resume work alone.
The secret-binding read is skipped when had_secret_bindings is false on the claimed row; the flag is trigger-set, never cleared, and read under the same advisory lock binding mutations take.
The post-boot egress replay is unchanged: the daemon's retry-adoption branch returns a running VM without applying rules and its proxy rules do not survive a restart, so dropping the replay waits for the daemon to apply rules on adoption and acknowledge it in the response.

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -762,6 +762,37 @@ func (h *Handlers) resumePausedSandbox(c *gin.Context, sandbox *db.Sandbox, team
 	sandboxID := sandbox.ID
 	l := sandboxLogger(sandboxID.String(), sandbox.HostID)
 
+	// Phase stamps, registered before ANY return so failed resumes land in
+	// the histograms too (create's discipline): claim is everything before
+	// the boot RPC (policy reads, the resuming claim, snapshot lookup); vmd
+	// is the boot RPC with its retry and the stateless fallback; post is
+	// the policy/egress/secret reapply. A stage that errored before its
+	// end-stamp reports its elapsed time. total stays with the callers,
+	// which own the request-level clock.
+	tStart := time.Now()
+	var tVmdStart, tVmdEnd, tPostDone time.Time
+	defer func() {
+		phases := map[string]time.Duration{}
+		if !tVmdStart.IsZero() {
+			phases["claim"] = tVmdStart.Sub(tStart)
+		} else {
+			phases["claim"] = time.Since(tStart)
+		}
+		switch {
+		case !tVmdEnd.IsZero():
+			phases["vmd"] = tVmdEnd.Sub(tVmdStart)
+		case !tVmdStart.IsZero():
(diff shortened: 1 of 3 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
## Verdict
APPROVE

I reviewed the current head (`87d0d01b`) in two independent passes: first manually, then with the structured PR-review workflow. Both converge on no remaining Critical or High defects.

## Review scope
This is a generic review (no explicit ticket/spec is linked in the PR). The change adds resume phase telemetry (`lookup`, `claim`, `vmd`, `post`, `revert`, `total`) and skips the secret-binding metadata read when the claimed sandbox row proves it has never had bindings.

## Key validation
- Traced explicit resume and auto-resume timing boundaries. Explicit resume partitions the synchronous path as intended; pre-claim failures record `lookup`, and compensation starts a separate `revert` phase rather than inflating `post`.
- Traced VMD ResumeInstance → NotFound → RestoreSnapshot fallback: the `vmd` phase includes the entire boot/fallback attempt and is closed before post-restore work starts.
- Traced post-stage failures through `failPost`, capability rejection, network replay, and secret reapply. The VM is re-paused before returning and the compensation duration is reported separately.
- Rechecked the secret-read shortcut against the actual database invariant rather than relying on the PR description. New sandboxes start with `had_secret_bindings=false`; the insert trigger makes it permanently true on first binding; legacy rows remain NULL and conservatively take the read.
- Checked the concurrency boundary: production wires `handlers.Pool = dbPool`; resume acquires `LockSandboxForSecretWrites` and runs `BeginResume` in the same transaction. Attach/detach use the same advisory lock and re-read status, so either the binding commits first (resume sees `true`) or resume claims first (the mutation sees `resuming` and rejects). A stale false cannot skip a concurrently-added binding.
- Rechecked the retained post-boot network replay. It remains necessary because retry adoption can return an already-running VM and proxy rules do not survive a VMD restart.

## Reviewer execution
| --- | --- | --- |

The only non-blocking testing note is that the new unit test proves the `had_secret_bindings=false` fast path itself, while the cross-request advisory-lock race is primarily established by the existing transaction/locking implementation rather than a new end-to-end concurrency test in this PR. I do not consider that sufficient to block this change because the invariant is enforced directly by the database transaction boundaries and the current production wiring.

PR #476, opened 2026-09-04
vm: wall-clock manifest, pause intent and a host floor for images that owe a wake

meAmitPatil:
**What it solves.** A restored sandbox's clock jumps forward by the snapshot's age and the guest works through the timer backlog before it is ready. Freezing the clock fixes that, but only safely if the supervisor knows whether the guest corrects its clock and whether its workload was frozen. This PR records those facts beside each image and adds the safety rails. Builds on #471.

**Why we need it.** The supervisor changes that freeze and wake workloads are the risky part. Landing the artifact format and the deployment safety first, dormant, gives them a fixed contract and guarantees no host ever runs a supervisor that cannot handle the images it holds.

**What changes.** The empty marker beside a memory image becomes a versioned manifest (guest corrects clock, workload frozen, release token); a pause-intent file blocks restores of an image a crash left mid-pause (this PR lands the reader and the refusal; the pause that writes one, and the vmd that carries the wake capability, arrive with #478); a host that is about to hold a frozen image gets a flag file first, and a systemd guard refuses to start a vmd without the wake capability while that flag exists; the guard reads only that file, never the snapshot trees, and the deploy script runs the same guard against a new binary before installing it; the template builder can freeze and mark an image behind a switch nothing sets yet; a build gets a unique id when its caller supplies none and never lands on a published template, so a template directory is immutable once published. Also carries the flaky backfill-test fix.

**Worth knowing.** Nothing in production produces a manifest, intent or evidence, so create, pause and resume take exactly today's paths; the only new costs are one small manifest open per create, beside the sidecar read that path already does (not-found for every template today), and one file check at vmd start, measured at 0.6 ms including the shell; a host whose flag is set also pays one pass over the binary, about 17 ms. This vmd deliberately lacks the wake capability, so the guard would refuse it if a frozen image ever appeared.

**Risks.** An old empty marker, if any exists, now reads as "restore the old way", the safe direction. The guard trusts the flag alone, so activation must set it on every host before the feature is turned on and must make every path that copies images between hosts refuse a frozen one onto an unflagged host. The floor is one-way: once flagged, a host takes only wake-capable builds. None of this can trigger until the supervisor changes land.

### DIFF
--- a/cmd/template-builder/main.go
+++ b/cmd/template-builder/main.go
@@ -387,7 +387,13 @@ func runBuild(ctx context.Context, cfg buildConfig) error {
 	}
 	emitInternal("system", "artifacts fsynced")
 
-	writeBuildMeta(snapDir, snapPath, memPath, basePath, deltaPath, correctsWallClock, br)
+	// Published last, and whole: the metadata is what tells a supervisor the
+	// directory holds a template, so it appears by rename once everything
+	// else is durable, and a build whose metadata cannot land has not been
+	// published.
+	if err := writeBuildMeta(snapDir, snapPath, memPath, basePath, deltaPath, correctsWallClock, br); err != nil {
+		return fmt.Errorf("publish build metadata: %w", err)
+	}
 
 	return nil
 }
@@ -1082,7 +1088,7 @@ func createBuildOverlay(runDir, vmID, basePath string) (string, error) {
 // Build metadata
 // ---------------------------------------------------------------------------
 
-func writeBuildMeta(dir, snapPath, memPath, basePath, deltaPath string, correctsWallClock bool, br builder.BuildRootfsResult) {
+func writeBuildMeta(dir, snapPath, memPath, basePath, deltaPath string, correctsWallClock bool, br builder.BuildRootfsResult) error {
 	// rootfs_path stays in the schema for backwards compat with existing
 	// readers; supervisors that understand overlay use base_path/delta_path.
 	meta := struct {
@@ -1111,9 +1117,38 @@ func writeBuildMeta(dir, snapPath, memPath, basePath, deltaPath string, corrects
 	}
 	data, err := json.MarshalIndent(meta, "", "  ")
 	if err != nil {
-		return
+		return err
 	}
-	_ = os.WriteFile(filepath.Join(dir, "build.meta.json"), data, 0o644)
+	final := filepath.Join(dir, "build.meta.json")
+	tmp := final + ".tmp"
+	f, err := os.OpenFile(tmp, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o644)
+	if err != nil {
+		return err
+	}
+	if _, err := f.Write(data); err != nil {
+		f.Close()
+		os.Remove(tmp)
+		return err
+	}
+	if err := f.Sync(); err != nil {
+		f.Close()
+		os.Remove(tmp)
+		return err
+	}
+	if err := f.Close(); err != nil {
+		os.Remove(tmp)
+		return err
+	}
+	if err := os.Rename(tmp, final); err != nil {
+		os.Remove(tmp)
+		return err
+	}
+	d, err := os.Open(dir)
+	if err != nil {
+		return err
+	}
+	defer d.Close()
+	return d.Sync()
 }
 
 // ---------------------------------------------------------------------------
--- a/cmd/template-builder/main.go
+++ b/cmd/template-builder/main.go
@@ -190,7 +190,13 @@ type buildConfig struct {
 }
 
 func runBuild(ctx context.Context, cfg buildConfig) error {
-	buildVMID := "build-" + cfg.templateID
+	// The build id names the runtime: the socket, overlay and network of
+	// this build alone, so two builds of one template can run side by side.
+	// The template-derived name remains for a caller that supplies no id.
+	buildVMID := cfg.buildID
+	if buildVMID == "" {
+		buildVMID = "build-" + cfg.templateID
+	}
 	// Per-build subdir — rebuilds don't disturb existing sandboxes.
 	rootfsDir := filepath.Join(cfg.runDir, vm.TemplatesDirName, cfg.templateID, cfg.buildID)
 	basePath := filepath.Join(rootfsDir, "base.ext4")
--- a/cmd/template-builder/main.go
+++ b/cmd/template-builder/main.go
@@ -701,7 +701,10 @@ func boxdFreezeWorkload(ctx context.Context, vmIP, token string) error {
 var errNoSuchFreeze = errors.New("guest holds no freeze under this token")
 
 // boxdThawWorkload releases a freeze whose outcome is unknown. 200 released;
-// 409 errNoSuchFreeze; anything else leaves the state unknown.
+// 409 with the guest reporting a running workload errNoSuchFreeze; anything
+// else leaves the state unknown. A 409 alone is not proof: the guest answers
+// it both when it holds no freeze under the token and when it holds one under
+// another, and only its own report tells those apart.
 func boxdThawWorkload(ctx context.Context, vmIP, token string) error {
 	url := fmt.Sprintf("http://%s:%d/thaw", vmIP, boxdPort)
 	payload, _ := json.Marshal(struct {
@@ -722,10 +725,38 @@ func boxdThawWorkload(ctx context.Context, vmIP, token string) error {
 	case http.StatusOK:
 		return nil
 	case http.StatusConflict:
+		if err := boxdWorkloadRunning(ctx, vmIP); err != nil {
+			return fmt.Errorf("POST /thaw: %s; %w", strings.TrimSpace(string(body)), err)
+		}
 		return errNoSuchFreeze
 	default:
 		return fmt.Errorf("POST /thaw: status %d: %s", resp.StatusCode, strings.TrimSpace(string(body)))
 	}
+}
+
+// boxdWorkloadRunning asks the guest whether its workload runs. Anything but
+// a plain ok is not proof that it does.
+func boxdWorkloadRunning(ctx context.Context, vmIP string) error {
+	url := fmt.Sprintf("http://%s:%d/health", vmIP, boxdPort)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return err
+	}
+	resp, err := (&http.Client{Timeout: 5 * time.Second}).Do(req)
+	if err != nil {
+		return fmt.Errorf("GET /health: %w", err)
+	}
+	defer resp.Body.Close()
+	var health struct {
+		Status string `json:"status"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&health); err != nil {
+		return fmt.Errorf("decode /health: %w", err)
+	}
+	if resp.StatusCode != http.StatusOK || health.Status != "ok" {
+		return fmt.Errorf("guest reports %q (status %d), not a running workload", health.Status, resp.StatusCode)
+	}
+	return nil
 }
 
 // userEnv filters the build-time env map down to the keys set via `env`
--- a/cmd/template-builder/main.go
+++ b/cmd/template-builder/main.go
@@ -690,8 +690,8 @@ func boxdFreezeWorkload(ctx context.Context, vmIP, token string) error {
 	if err := json.Unmarshal(body, &echo); err != nil {
 		return fmt.Errorf("POST /freeze: reply: %w", err)
 	}
-	if echo.Version != vm.WallClockManifestVersion || echo.Token != token {
-		return fmt.Errorf("POST /freeze: guest speaks protocol %d and holds token %q, asked %d %q", echo.Version, echo.Token, vm.WallClockManifestVersion, token)
+	if echo.Version != vm.WakeProtocolVersion || echo.Token != token {
+		return fmt.Errorf("POST /freeze: guest speaks protocol %d and holds token %q, asked %d %q", echo.Version, echo.Token, vm.WakeProtocolVersion, token)
 	}
 	return nil
 }
--- a/cmd/template-builder/main.go
+++ b/cmd/template-builder/main.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -181,6 +182,11 @@ type buildConfig struct {
 	slotIndex   int
 	launcherNS  string
 	netlinkOps  bool
+	// freezeWorkload builds the image with its workload freezer and, once the
+	// guest has proven it corrects its clock, snapshots it frozen. Set by
+	// nothing yet: a frozen image is only safe under a supervisor that can wake
+	// it, so the switch arrives with that supervisor.
+	freezeWorkload bool
 }
 
 func runBuild(ctx context.Context, cfg buildConfig) error {
@@ -215,6 +221,7 @@ func runBuild(ctx context.Context, cfg buildConfig) error {
 		BoxdBinaryPath:           cfg.boxdBin,
 		MaxUncompressedSizeBytes: int64(cfg.diskMiB) * 1024 * 1024,
 		ProxyCACertPath:          proxyCA,
+		FreezeWorkload:           cfg.freezeWorkload,
 	})
 	if err != nil {
 		return fmt.Errorf("create builder: %w", err)
@@ -311,6 +318,43 @@ func runBuild(ctx context.Context, cfg buildConfig) error {
 		return fmt.Errorf("guest sync: %w", err)
 	}
 
+	// Ask the guest, in the state the snapshot captures, whether it can correct
+	// its wall clock and freeze its workload; only then may the image be marked.
+	correctsWallClock, why := false, "template workload freezing not enabled"
+	if cfg.freezeWorkload {
+		correctsWallClock, why = boxdWallClockProven(ctx, netInfo.HostIP)
+	}
+	// The token this image's freeze carries; every wake of a sandbox created
+	// from it must present it. Minted here, kept by the guest in the snapshot.
+	freezeToken := vm.NewFreezeToken()
+	if correctsWallClock {
+		// The rollback floor rises before the first frozen artifact can exist
+		// on this host: a supervisor without the wake protocol must never be
+		// installed over one, so a floor that cannot be made durable fails
+		// the build rather than freeze anything.
+		if err := vm.RaiseWakeProtocolFloor(); err != nil {
+			return fmt.Errorf("raise wake-protocol floor: %w", err)
+		}
+	}
+	if correctsWallClock {
+		// A freeze whose answer was lost may still have frozen the guest, and
+		// an image with a stopped workload and no manifest would never be
+		// woken. So a failed freeze is followed by a thaw with the same
+		// token: released, or never frozen, means the image is unfrozen and
+		// the build goes on the older way; anything else fails the build.
+		if err := boxdFreezeWorkload(ctx, netInfo.HostIP, freezeToken); err != nil {
+			if terr := boxdThawWorkload(ctx, netInfo.HostIP, freezeToken); terr != nil && !errors.Is(terr, errNoSuchFreeze) {
+				return fmt.Errorf("workload freeze failed (%v) and the guest could not be released (%v); refusing to snapshot a workload in an unknown state", err, terr)
+			}
+			correctsWallClock, why = false, "workload freeze: "+err.Error()
+		}
+	}
+	if correctsWallClock {
+		emitInternal("system", "guest corrects its wall clock and froze its workload: marking template")
+	} else {
+		emitInternal("system", "template stays on the unfrozen restore path (%s)", why)
+	}
+
 	// block_delta_dir → emits rootfs.delta so sandboxes can be created
 	// from this template's snapshot.
 	emitUser("system", "Saving template")
@@ -322,14 +366,28 @@ func runBuild(ctx context.Context, cfg buildConfig) error {
 	}
 	emitInternal("system", "snapshot captured")
 
+	// The manifest tells a supervisor this image's workload is frozen, that its
+	// guest corrects its clock, and which token the wake must present.
+	markerPath := ""
+	if correctsWallClock {
+		markerPath = vm.WallClockMarkerPath(memPath)
+		man := vm.WallClockManifest{
+			Version: vm.WallClockManifestVersion, ArtifactID: vm.NewArtifactID(),
+			WorkloadFrozen: true, GuestCorrectsClock: true, FreezeToken: freezeToken,
+		}
+		if err := vm.WriteWallClockManifest(memPath, man); err != nil {
+			return fmt.Errorf("write wall-clock manifest: %w", err)
+		}
+	}
+
 	// Flush host page cache for each artifact so a sandbox cp'ing them
 	// later doesn't see sparse holes from still-dirty pages.
-	if err := fsyncBuildArtifacts(snapDir, snapPath, memPath, basePath, deltaPath); err != nil {
+	if err := fsyncBuildArtifacts(snapDir, snapPath, memPath, basePath, deltaPath, markerPath); err != nil {
 		return fmt.Errorf("fsync build artifacts: %w", err)
 	}
 	emitInternal("system", "artifacts fsynced")
 
-	writeBuildMeta(snapDir, snapPath, memPath, basePath, deltaPath, br)
+	writeBuildMeta(snapDir, snapPath, memPath, basePath, deltaPath, correctsWallClock, br)
 
 	return nil
 }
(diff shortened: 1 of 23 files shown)


meAmitPatil:
@codex review

internal/vm/manager.go
```diff
@@ -1981,6 +1993,26 @@ func readLayeredBase(memPath string) (string, bool) {
...
+func (m *Manager) imageManifestCached(memPath string) (*WallClockManifest, error) {
+	if !m.isTemplateMemPath(memPath) {
+		return imageManifest(memPath)
+	}
+	key := filepath.Clean(memPath)
+	if v, ok := m.templateManifests.Load(key); ok {
+		return v.(*WallClockManifest), nil
+	}
+	man, err := imageManifest(memPath)
+	if err != nil {
```
awille-ai: **High — don't permanently cache manifest facts by a mutable template path.** `imageManifestCached` keeps the first answer for `memPath` forever, but template artifacts at that path are replaceable: `BuildTemplate` defaults to `build-<templateID>`, writes to `templates/<templateID>/<buildVMID>/...`, and `registerBuild` allows that same build ID again after the prior build is terminal. So a same-template rebuild can replace this `mem.snap`/manifest in place. If this daemon previously cached `nil` for the legacy artifact, it will keep treating a later frozen artifact as legacy; stacked PR #478 still uses this cache to decide whether to take the wake path, so after activation that can restore a frozen workload without waking it. The reverse transition can also preserve a stale positive. Key by an immutable artifact/build generation or invalidate/update the cache when a build replaces the path, and change `TestTemplateManifestIsReadOncePerDaemon` to exercise the supported rebuild case instead of asserting stale data.

awille-ai (commented):
## Verdict
COMMENT

The skill-based review found one High correctness/rollout issue that the initial manual pass missed: `imageManifestCached` assumes a template memory path is immutable for the lifetime of a vmd process, but the repository allows the same template/build path to be rebuilt in place.

`BuildTemplate` defaults `buildVMID` to `build-<templateID>`, `buildTemplateSync` writes to `templates/<templateID>/<buildVMID>/...`, and `registerBuild` permits the same build ID to be registered again after the previous record is terminal. So a rebuild of the same template can replace `mem.snap` and its `.wallclock` manifest at the exact path already cached by vmd.

That makes the permanent path-keyed cache unsafe. A daemon that first cached `nil` for a legacy template will continue returning legacy after that path is rebuilt with a frozen-workload manifest. This PR's own `TestTemplateManifestIsReadOncePerDaemon` explicitly asserts that stale behavior. The stacked wake-capable PR #478 still consumes `imageManifestCached` to decide whether an image is frozen and needs `/wake`, and does not invalidate this cache, so after activation the stale-negative case can restore a newly frozen template without taking the wake path and leave its workload stopped. The reverse rebuild can leave a stale positive result and unnecessarily refuse/use the wrong policy for a now-legacy image.

The cache needs an immutable identity/generation in its key, or explicit invalidation/update when a template build replaces a path. The test should model the supported same-template rebuild rather than asserting that a changed manifest is ignored forever.

## Requirement coverage — SS-460
- R1 — versioned wall-clock manifest with legacy compatibility and fail-closed handling of malformed/unknown manifests: **satisfied**.
- R2 — durable pause-intent artifact that distinguishes completed versus interrupted image rewrites and blocks unsafe restore/resume: **satisfied as the groundwork in this stacked PR**.
- R3 — durable wake-protocol rollback floor, deploy-time guard, and host-resident systemd guard that survive rollback to an older deploy script: **satisfied**.
- R4 — template builder can prove PTP/settime + freezer behavior, freeze conservatively, raise the floor first, and emit a frozen manifest; activation remains off: **satisfied**.
- R5 — staged rollout must remain correct when legacy templates are rebuilt/recreated into wake-capable templates: **not satisfied** because the template-manifest cache can retain the old artifact's facts across an in-place rebuild.
- R6 — existing/legacy records and empty pre-manifest `.wallclock` markers continue on the legacy clock path: **satisfied**.
- R7 — carried backfill-pool test fix should cover the real retry window without weakening the assertion: **satisfied**.

## Reviewer execution
| --- | --- | --- |
| Requirements | sequential fallback | FINDINGS |
| Security/resilience | sequential fallback | FINDINGS |
| Test quality | sequential fallback | FINDINGS |
| General correctness | sequential fallback | FINDINGS |
| Data correctness | sequential fallback | FINDINGS |
| Infrastructure/deployment safety | sequential fallback | PASS |

## Validation
- Reviewed current head `514af24296a165671f768b906d78383418a3b2ff`, full diff, surrounding template-build/registry code, deployment guards, state conversion, manifest/intent tests, and the stacked #478 consumer.
- Confirmed same-ID rebuild is supported: terminal build records do not prevent re-registration, and the default build ID/path is deterministic from the template ID.
- Confirmed the new cache intentionally retains a negative answer after a manifest appears at the same template path.
- Confirmed #478 still uses this cache to decide frozen-workload/wake behavior and does not evict it.
- Rechecked crash/order behavior for manifest publication, floor creation, pause-intent fsync/clear, failed freeze + thaw, malformed/future manifests, old empty markers, and rollback to a pre-wake binary.
- CI and Terraform Checks are green on this head. The deployment changes do not alter Terraform-managed resources; the deployment scripts/drop-ins themselves were reviewed directly.

deploy/vmd-wake-floor-guard
```diff
@@ -0,0 +1,38 @@
...
+SNAPSHOTS="${SNAPSHOTS:-/var/lib/sandbox/snapshots}"
+wake_evidence=""
+[ -e "$WAKE_EVIDENCE" ] && wake_evidence="$WAKE_EVIDENCE"
+if [ -z "$wake_evidence" ] && grep -lqs '"workload_frozen":true' "$SNAPSHOTS"/templates/*/*/*.wallclock "$SNAPSHOTS"/templates/*/*.wallclock "$SNAPSHOTS"/*/*.wallclock 2>/dev/null; then
+    wake_evidence="frozen image manifest under ${SNAPSHOTS}"
+fi
+if [ -z "$wake_evidence" ]; then
+    exit 0
+fi
+if grep -qa wake-protocol-1 "$BIN" 2>/dev/null; then
```
pavitrabhalla, line 38: **Blocking.** This greps the vmd binary for the literal string `wake-protocol-1`, but that string doesn't appear anywhere in the Go source, and LDFLAGS is just `-s -w` (no version stamping). The moment `freezeWorkload` gets turned on anywhere and a build succeeds, `RaiseWakeProtocolFloor` writes the evidence file, and every vmd start on that host fails this check forever, since no build can ever contain the string. Same issue in deploy-vmd.py:480. Where's this string supposed to come from — an ldflags stamp on `cmd/vmd` that hasn't landed yet?

internal/vm/pause_intent.go
```diff
@@ -0,0 +1,105 @@
...
+
+type pauseIntent struct {
+	VMID        string `json:"vm_id"`
+	FreezeToken string `json:"freeze_token,omitempty"`
+	ArtifactID  string `json:"artifact_id"`
+}
+
+func pauseIntentPath(dir string) string { return filepath.Join(dir, pauseIntentName) }
+
+func writePauseIntent(dir string, in pauseIntent) error {
```
pavitrabhalla, line 29: **Blocking.** `writePauseIntent` has no production call sites — only pause_intent_test.go calls it. `PauseVM` clears the old wall-clock marker and later writes the new manifest without ever writing a pause.intent around it, even though `pauseIntentBlocks` is already live in `resumeVMLocked`/`restoreVMSnapshot`. Is wiring this into `PauseVM` deliberately deferred to a follow-up, or a miss? As shipped, the crash-safety story this file describes doesn't actually apply to a real pause.

cmd/template-builder/main.go
```diff
@@ -558,6 +616,149 @@ func postBoxdInit(ctx context.Context, vmIP string, envVars map[string]string, d
...
+	// The reply names the protocol the guest speaks and the token it holds;
+	// an image is only marked when both are what this builder asked for.
+	var echo struct {
+		Version int    `json:"version"`
+		Token   string `json:"token"`
+	}
+	if err := json.Unmarshal(body, &echo); err != nil {
+		return fmt.Errorf("POST /freeze: reply: %w", err)
+	}
+	if echo.Version != vm.WallClockManifestVersion || echo.Token != token {
```
pavitrabhalla: **Non-blocking.** `echo.Version` here is boxd's wire-protocol constant (`protocolVersion` in cmd/boxd/freezer.go), compared against `vm.WallClockManifestVersion`, the on-disk manifest schema version. Unrelated version spaces that both happen to be `1` today. Worth its own constant so a schema bump doesn't silently break this check, or vice versa?

internal/vm/manager.go
```diff
@@ -542,6 +545,9 @@ type Manager struct {
 	// restoreSem bounds concurrent RestoreVMSnapshot operations. Buffered
 	// channel; capacity = effective MaxConcurrentRestores.
 	restoreSem chan struct{}
+	// templateManifests keeps the manifest beside each template image this
+	// daemon has restored from; see imageManifestCached.
+	templateManifests sync.Map
```
pavitrabhalla: **Non-blocking.** `templateManifests` is only ever Store'd/Load'd via `imageManifestCached` — `DeleteTemplateArtifacts`/`DeleteBuildArtifacts` don't evict from it. On a host that builds and retires many templates over a long uptime this grows unbounded. Should deletion clear the corresponding key?

internal/vm/clock_policy.go
```diff
@@ -98,66 +91,63 @@ func (m *Manager) WatchFirecrackerCapability(ctx context.Context, log zerolog.Lo
...
-// look, because the record then describes a different artifact.
-func resumeWallClockProperty(memPath, basePath, pausedMemPath string, recorded *bool) bool {
+// pause wrote the property to the record and the manifest together — so the
+// record is already the answer and the resume path need not go to disk for it.
+// Only an explicit override, supplying an image this VM was never paused into,
+// or a record that never carried the answer (a rollback to a binary without the
+// field drops it on rewrite, and its silence must not be read as "no") has to
+// look. That read also says whether the image holds a frozen workload, which
+// the caller must refuse: a manifest this binary cannot trust is an error.
+func resumeWallClockProperty(memPath, pausedMemPath string, recorded *bool) (correctsWallClock, workloadFrozen bool, err error) {
```
pavitrabhalla: **Non-blocking.** This dropped the base-manifest fallback that the equivalent pause-time check (`guestCorrectsWallClock` in `PauseVM`) still has for an override-resume with no manifest of its own. Currently masked since `clockPolicyFor` is hardcoded `false` at both call sites, but `inst.CorrectsWallClock` can already be persisted incorrectly for that case. Intentional given the "overlay never inherits base manifest" rule on `imageManifest`, or worth aligning with the pause-time check?

internal/vm/wallclock_manifest.go
```diff
@@ -0,0 +1,235 @@
...
+const WallClockManifestVersion = 1
+
+// wakeProtocolEvidencePath records that this host has held an image that
+// owes a wake, so the host-resident guard refuses a vmd without the wake
+// protocol from then on. Raised by the template builder before it publishes
+// a frozen image, and by a supervisor before it acts on one. A supervisor
+// that only starts creates none, or every host would raise the floor with
+// both switches off and block a rollback of itself. A host without the
+// directory is not a fleet host.
+var wakeProtocolEvidencePath = "/var/lib/sandbox/wake-protocol-evidence"
```
pavitrabhalla, line 52: **Non-blocking.** This path, plus the SNAPSHOT_DIR-parsing sed one-liner, are hardcoded independently in Go, deploy-vmd.py, and vmd-wake-floor-guard with no shared source of truth (deploy-vmd.py's comment even calls out "same check as the host-resident start guard"). Same disconnect underlies the dead `wake-protocol-1` grep above — worth a single source these three read from?

internal/vm/manager.go
meAmitPatil: Fixed in f5c72a1e. The cache is gone: every restore reads the manifest from disk, one small open beside the sidecar read that path already does, and the test now covers a template replaced in place in both directions. Builds also get a unique id when the caller supplies none and never land on a published template directory, so a template directory is immutable once published.

deploy/vmd-wake-floor-guard
meAmitPatil: Deliberate: this vmd cannot wake a frozen image, so the guard has to refuse it once a host holds one. The string arrives with the wake-capable vmd in #478 (`vm.WakeProtocolCapability`, printed by `vmd capabilities`); I built that binary and ran this guard against it. The scenario you describe needs the builder's freeze switch on before #478 is deployed, and that switch is set by nothing until activation. As of c45c4fe6 the deploy script runs this same guard against the new binary instead of carrying a copy, and there is a shell harness for it (`wakefloorguard_test.go`).

internal/vm/clock_policy.go
meAmitPatil: Agreed, aligned in c45c4fe6: an override resume of an overlay with no manifest of its own now takes the guest's capability from the base beneath it, as the pause-time check does, and never the frozen fact.

internal/vm/wallclock_manifest.go
meAmitPatil: Agreed, reduced in c45c4fe6: the daemon records its resolved snapshot directory in a breadcrumb at start (the pattern the rollback guard already uses for the state path), the guard reads that instead of parsing the env file, and the deploy script runs the bundled guard against the new binary rather than carrying a copy of its logic. The evidence path is the one remaining contract, written by the daemon and documented in the guard header.

internal/vm/pause_intent.go
meAmitPatil: Deliberate split, not a miss: this PR lands the reader and the refusal, and #478's pause writes the intent before it freezes, after raising the floor. Only a pause that freezes needs one. I have made the description say so plainly.

cmd/template-builder/main.go
meAmitPatil: Agreed, fixed in c45c4fe6: the echo is now checked against `vm.WakeProtocolVersion`, its own version space, and the manifest schema version is no longer involved.

internal/vm/manager.go
meAmitPatil: Gone as of f5c72a1e: the cache was removed altogether and every restore reads the manifest from disk, so there is nothing to evict.

meAmitPatil:
@codex review

internal/vm/build.go
```diff
@@ -81,7 +83,14 @@ func (m *Manager) BuildTemplate(ctx context.Context, req BuildTemplateRequest) (
...
 	buildVMID := req.BuildVMID
 	if buildVMID == "" {
-		buildVMID = "build-" + req.TemplateID
+		buildVMID = defaultBuildVMID(req.TemplateID)
+	}
+	// A published template is immutable: its directory is what every sandbox
+	// created from it, and every paused overlay layered on it, refers to. A
+	// build never lands on one; a rebuild is a new build id and a new
+	// directory, which is how the control plane names its builds anyway.
+	if err := m.prepareBuildDir(req.TemplateID, buildVMID); err != nil {
```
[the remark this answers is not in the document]
meAmitPatil: Fixed in dcf78812: the directory is prepared inside the registry's critical section, after the in-flight check and before the record is inserted, so a retry of an id still in flight is refused before it could touch the running build's files. Test added for exactly that retry.

internal/vm/clock_policy.go
```diff
@@ -98,66 +91,74 @@ func (m *Manager) WatchFirecrackerCapability(ctx context.Context, log zerolog.Lo
...
 	if memPath == pausedMemPath && recorded != nil {
-		return *recorded
+		return *recorded, false, nil
```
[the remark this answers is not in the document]
meAmitPatil: Fixed in dcf78812: the record now carries the image fact (`snapshot_workload_frozen`, set by every pause and every restore), and the same-image shortcut keys on that, not on the guest's capability. A record that lacks the fact reads the manifest; only a guest that cannot correct its clock skips the read, since such a guest is never frozen. Test added for the case you describe.

internal/vm/manager.go
```diff
@@ -1749,14 +1752,20 @@ func (m *Manager) PauseVM(ctx context.Context, vmID, snapshotDir, pauseToken str
...
+	// drops back to legacy. Only the write lands here: the unsafe direction was
+	// already handled above, before the image existed. Best-effort by design — a
+	// manifest that fails to write costs a slower resume, never a wrong clock.
 	if correctsWallClock {
-		if merr := os.WriteFile(clockFreezeMarkerPath(memPath), nil, 0o644); merr != nil {
-			log.Warn().Err(merr).Str("path", clockFreezeMarkerPath(memPath)).
-				Msg("pause: wall-clock marker write failed; resume falls back to legacy clock behaviour")
+		artifact := NewArtifactID()
+		man := WallClockManifest{Version: WallClockManifestVersion, ArtifactID: artifact, GuestCorrectsClock: true}
+		if merr := WriteWallClockManifest(memPath, man); merr != nil {
```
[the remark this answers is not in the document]
meAmitPatil: Fixed in dcf78812: an unfrozen manifest, the only kind this PR's pause writes, is published atomically but with no fsync or directory sync, since losing it costs a slower resume and nothing else. The durable form stays for a frozen manifest, which must outlive a crash.

meAmitPatil:
@codex review

deploy/vmd-wake-floor-guard
```diff
@@ -0,0 +1,42 @@
...
+# writes both witnesses' locations: the evidence file when it first acts on
+# such an image, and its resolved snapshot directory at every start, read
+# here rather than parsed out of env files, as the rollback guard reads the
+# state path. No breadcrumb means the daemon has never run here.
+WAKE_EVIDENCE=/var/lib/sandbox/wake-protocol-evidence
+SNAPSHOTS=$(head -n 1 /var/lib/sandbox/vmd-snapshot-dir 2>/dev/null)
+SNAPSHOTS="${SNAPSHOTS:-/var/lib/sandbox/snapshots}"
+wake_evidence=""
+[ -e "$WAKE_EVIDENCE" ] && wake_evidence="$WAKE_EVIDENCE"
+if [ -z "$wake_evidence" ] && grep -lqs '"workload_frozen":true' "$SNAPSHOTS"/templates/*/*/*.wallclock "$SNAPSHOTS"/templates/*/*.wallclock "$SNAPSHOTS"/*/*.wallclock 2>/dev/null; then
```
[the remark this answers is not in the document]
meAmitPatil: Fixed in 9e7dbf19. The guard now admits a binary that carries the capability before it looks at anything else, so an ordinary start never walks the trees. The walk runs only for an incapable binary, is batched by find so no argument list can overflow, folds grep's no-match into success so any other failure reaches the exit code, and a walk that cannot complete refuses to start rather than reading as nothing found. Harness case added with three thousand images and one frozen among them.

internal/vm/build.go
```diff
@@ -99,6 +109,31 @@ func (m *Manager) BuildTemplate(ctx context.Context, req BuildTemplateRequest) (
...
+
+// prepareBuildDir refuses a build into a directory that holds a published
+// template, and clears one a crashed build left unpublished: its files —
+// a snapshot, a wall-clock manifest — must not survive beside the new
+// build's. Published means the builder wrote its metadata, which it does
+// last.
+func (m *Manager) prepareBuildDir(templateID, buildVMID string) error {
+	dir := filepath.Join(m.cfg.SnapshotDir, TemplatesDirName, templateID, buildVMID)
+	if _, err := os.Stat(filepath.Join(dir, buildMetaFilename)); err == nil {
+		return status.Errorf(codes.AlreadyExists, "template %s build %s is already published at %s; a rebuild needs a new build id", templateID, buildVMID, dir)
```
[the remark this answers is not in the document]
meAmitPatil: Fixed in 9e7dbf19. Published now means the metadata reads whole: an empty or torn file counts as the leftovers of an interrupted build and is cleared, any other read failure blocks the build, and the builder writes its metadata atomically (temp, fsync, rename, directory sync) and fails the build if that cannot land.

meAmitPatil:
All three fixed in 9e7dbf19.

1. Build ids must each name a single directory: no separators, no dot or dot-dot, and the resolved path must sit exactly two levels under the templates root; anything else is refused before anything is looked at. Metadata read errors other than not-found block the build. Test covers eight traversing ids against a published template that must survive.
2. The guard admits a binary carrying the capability first, so an ordinary start never walks the trees. For an incapable binary the walk is batched by find, folds grep's no-match into success so every other failure reaches the exit code, and refuses to start when it cannot complete. Harness case with three thousand images.
3. Metadata is published atomically and the build fails if it cannot land; an empty or torn file reads as leftovers and is cleared on retry.

deploy/vmd-wake-floor-guard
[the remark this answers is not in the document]
meAmitPatil: Superseded in 82212c19: the walk is gone entirely. The guard reads one file and nothing else; a host holds far too many images to walk at any start, rollback included. Every producer of a frozen image raises the flag before the image exists (the builder here, the daemon in #478), and activation will set it on every host up front and gate the paths that copy images between hosts. Measured on Linux with a 61 MB binary: 0.6 ms per start with no flag, which is every host today; 17 ms with the flag set, for the one pass over the binary. A guard that cannot read the flag's directory refuses to start rather than reading it as absent.

meAmitPatil:
Update on the restart guard, 82212c19: the tree walk is gone entirely, rollback case included. The guard reads one flag file; if absent it exits, if present it checks the binary for the wake capability and refuses one without it; if it cannot read the flag's directory it refuses rather than guessing. Measured per vmd start on Linux: 0.6 ms with no flag (every host today), 17 ms with the flag set. Nothing on create, pause or resume, and nothing that scales with the number of sandboxes or templates.

The trade is that the flag is the only witness, so activation must set it on every host before the feature is turned on, and every path that copies images between hosts must refuse a frozen one onto an unflagged host. Both are the activation change's job; nothing can produce a frozen image until then.

meAmitPatil:
Fixed in f0abf091. The guard no longer asks whether the marker exists; it looks it up with stat, following symlinks, and admits any vmd only when the lookup says no such file. Every other failure, an ancestor that cannot be traversed, a symlink loop, anything else, counts as presence: the binary is checked and an incapable one is refused. Harness case added with a symlink loop at the marker path, and the marker-absent cost is unchanged at about 0.6 ms per start.

meAmitPatil:
@codex review

internal/vm/build.go
```diff
@@ -99,6 +112,59 @@ func (m *Manager) BuildTemplate(ctx context.Context, req BuildTemplateRequest) (
 	return buildVMID, nil
 }
 
+// defaultBuildVMID names a build whose caller supplied no id: unique, so it
+// can never land where a published template already is.
+func defaultBuildVMID(templateID string) string {
+	return "build-" + templateID + "-" + randomHex()[:8]
```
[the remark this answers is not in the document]
meAmitPatil: Fixed in 3cd51f8d. The builder now names its socket, overlay and network by the build id vmd passed (falling back to the template name only when no id is given), and vmd removes that same directory when the build ends. Two builds of one template no longer share a runtime directory.

internal/vm/manager.go
```diff
@@ -3056,6 +3087,37 @@ func (m *Manager) restoreVMSnapshot(ctx context.Context, vmID, snapshotPath, mem
...
 		}
 	}
+	// An in-place restore of a VM this daemon knows may meet the leftover of a
+	// pause that completed; any other intent means an interrupted rewrite. A
+	// template holds no intent, and no host whose floor is down holds one, so
+	// a create looks for nothing.
+	m.mu.RLock()
+	prev := m.vms[vmID]
+	m.mu.RUnlock()
+	if prev != nil && wakeProtocolFloorRaised() {
```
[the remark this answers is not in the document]
meAmitPatil: Fixed in 3cd51f8d. The intent check now runs on every restore once the floor is up; with no prior record it checks with an empty known artifact, so an intent beside the image can never be cleared as completed. Test covers a restore under a new id: refused before launch, intent kept.

internal/vm/build_registry.go
```diff
@@ -119,6 +119,13 @@ func (m *Manager) registerBuild(buildVMID, templateID string, vcpu, memoryMiB ui
 	if existing, ok := m.builds[buildVMID]; ok && !existing.Status.IsTerminal() {
 		return nil, fmt.Errorf("build %s already in flight", buildVMID)
 	}
+	// Under the lock, so nothing can claim the id between the check above
+	// and whatever prepare removes.
+	if prepare != nil {
+		if err := prepare(); err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
meAmitPatil: Fixed in 3cd51f8d. A build record now carries a done channel the worker closes as its last act, after the subprocess has exited. A retry against a cancelled id is refused with a retry-shortly error until then; the two generation tests stage the worker exit before re-registering.

meAmitPatil:
@codex review

pavitrabhalla (approved):

awille-ai (approved):
Re-review of current head `3cd51f8d`: APPROVE.

- Requirements / SS-460: PASS for this PR's dormant artifact/deployment-safety slice
- Hostile/adversarial: PASS
- Security/resilience: PASS
- Test quality: PASS
- General correctness: PASS
- Compatibility/rollout: PASS
- Data correctness/state transitions: PASS
- Infrastructure/deployment safety: PASS (no Terraform resource plan changes; deployment-script/systemd guard inspected directly)

The previous High about permanently caching manifest facts by a mutable template path is fixed: the cache is removed, restores read the manifest from disk, default builds receive unique IDs, published build directories are immutable, duplicate/cancelled build IDs stay fenced until the worker exits, and tests cover replacement/concurrency cases. I also checked the later fixes for fresh-restore pause intents, same-template concurrent runtime isolation, durable-vs-lazy manifest publication, and the bounded host floor guard. CI and Terraform Checks are green.

Non-blocking rollout dependency remains explicit: activation must wait for the wake-capable supervisor in #478, and any future cross-host image import path must establish the wake-protocol evidence floor before a frozen image lands.

PR #483, opened 2026-09-08
boxd: create the parent only when absent and surface filesystem errors with their mount

meAmitPatil:
os.MkdirAll treats any stat error as "missing" and then fails with EEXIST on the existing entry, hiding the filesystem's real error. boxd now creates a parent only when it is genuinely absent and lets any other lookup error surface, so it reaches the ESTALE retry instead of failing outright.

Failures raised by the filesystem itself (ESTALE, EIO, an unreachable mount) now return `error.code = sandbox_filesystem_error` with the errno and the mount that served the path, alongside the existing `sandbox_storage_full` shape. Status codes are unchanged.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -4357,7 +4357,10 @@ components:
           properties:
             error:
               type: object
+              required: [code, errno]
               properties:
+                code:
+                  const: sandbox_filesystem_error
                 errno:
                   type: string
                 path:
@@ -4433,7 +4436,7 @@ components:
       content:
         application/json:
           schema:
-            oneOf:
+            anyOf:
               - $ref: "#/components/schemas/Error"
               - $ref: "#/components/schemas/FilesystemError"
     ServiceUnavailable:
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1904,7 +1904,7 @@ paths:
         "404":
           $ref: "#/components/responses/NotFound"
         "500":
-          $ref: "#/components/responses/InternalError"
+          $ref: "#/components/responses/FilesInternalError"
 
     post:
       operationId: writeFile
@@ -1934,7 +1934,7 @@ paths:
         "507":
           $ref: "#/components/responses/StorageFull"
         "500":
-          $ref: "#/components/responses/InternalError"
+          $ref: "#/components/responses/FilesInternalError"
 
   /templates:
     get:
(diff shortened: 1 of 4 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
## Verdict
APPROVE

I did not find any remaining Critical or High issue in the current head (`4085190a`). The change fixes the original `MkdirAll` masking problem without changing success semantics, preserves the bounded ESTALE retry behavior, and consistently surfaces qualifying filesystem-originated failures through the new structured `sandbox_filesystem_error` envelope.

## Review scope
Generic review (no explicit ticket/spec is linked from the PR). Sequential passes covered hostile/adversarial behavior, security/resilience, test quality, general correctness, performance/resource efficiency, observability, and compatibility/rollout.

## Validation
- Rechecked all existing review threads against the current code rather than assuming resolved means fixed. The current head avoids re-entering an already-known mount during error formatting, decodes mountinfo octal escapes, classifies filesystem failures during symlink resolution, and documents the response in OpenAPI using `anyOf`.
- Traced GET/download, JSON directory listing, archive, and upload error branches to confirm filesystem errnos reach `writeFSError` while request/validation errors retain their prior 4xx behavior.
- Checked `ensureParentDir`: existing directories and symlink parents are left alone; only genuine ENOENT causes `MkdirAll`, so ESTALE/EIO/etc. survive to the classifier instead of being replaced by EEXIST.
- Checked ESTALE upload retry: both parent lookup and open/write errors remain unwrap-compatible through `fmt.Errorf(... %w ...)`, so the existing stale-handle retry still recognizes them.
- Checked mount attribution: `/proc/self/mountinfo` parsing uses generic octal decoding and longest mountpoint matching; symlink expansion stops before inspecting components once the textual path enters a non-root mount, avoiding another filesystem timeout on the error-reporting path.
- The extra mountinfo parsing/Lstat work occurs only after an operation has already failed and does not add work to the normal file-transfer path.
- The public HTTP status remains 500, so existing callers that only branch on status are compatible; clients that understand the new code can consume errno/path/mount metadata.

## Reviewer execution
| --- | --- | --- |

One semantic nuance is worth keeping in mind but is not a blocker: the classifier keys on errno, so an EIO/ESTALE from the sandbox root filesystem is also labeled `sandbox_filesystem_error`, not only failures from user-added FUSE/network mounts. The response includes mountpoint/fstype, so operators/callers can distinguish that case; the behavior is still materially more diagnostic than the previous opaque 500.

PR #485, opened 2026-09-09
resume: one daemon call carrying the preview policy, rules acked on adoption, and no secrets re-injection when the guest already holds the current environment

meAmitPatil:
Resume makes one daemon call: the preview policy the claim read rides the resume request, the daemon stamps it before the guest runs and attests it, so the post-boot policy re-read and push remain only for the stateless fallback and for daemons from before the echo. The daemon also applies the request's egress rules when a retried resume adopts an earlier attempt's VM and acks it, so the control plane replays them only when unacked.
A resume with secret bindings skips the JWT re-mint and guest round trip while the recorded binding set, guest IP, JWT life, and snapshot age all say the guest still holds the current environment; any doubt injects. Either deploy order is safe; the migration adds four nullable columns to the sandbox row.

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -1119,15 +1119,23 @@ func (h *Handlers) resumePausedSandbox(c *gin.Context, sandbox *db.Sandbox, team
 			return "", false
 		}
 	case attested.PreviewProtocol == preview.HostCapabilityPorts:
-		// Stamped before the guest ran. A record that already held a newer
-		// policy kept it, and the reply must report that one.
-		if attested.PreviewPolicyRevision != resumePolicy.Revision {
-			currentPolicy, policyErr := h.loadPreviewPolicy(postCtx, sandboxID, teamID)
-			if policyErr != nil {
-				failPost(policyErr, "reload preview policy after resume failed")
+		// Stamped before the guest ran, or a newer policy the record already
+		// held and kept. The database may be newer still: a mutation that
+		// committed after the claim and whose own push failed. One revision
+		// read proves what is current and names the policy the reply
+		// reports; only a daemon behind it gets the full read and push.
+		currentPolicy, policyErr := h.loadPreviewPolicy(postCtx, sandboxID, teamID)
+		if policyErr != nil {
+			failPost(policyErr, "reload preview policy after resume failed")
+			return "", false
+		}
+		effectivePolicy = currentPolicy
+		if currentPolicy.Revision > attested.PreviewPolicyRevision {
+			l.Warn().Int64("db_revision", currentPolicy.Revision).Int64("vmd_revision", attested.PreviewPolicyRevision).
+				Msg("daemon preview policy behind the database after resume; pushing")
+			if !reapplyPolicy() {
 				return "", false
 			}
-			effectivePolicy = currentPolicy
 		}
 	case attested.PreviewProtocol == "":
 		// A daemon from before the resume request carried the policy. It may
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -1130,10 +1130,22 @@ func (h *Handlers) resumePausedSandbox(c *gin.Context, sandbox *db.Sandbox, team
 			return "", false
 		}
 		effectivePolicy = currentPolicy
-		if currentPolicy.Revision > attested.PreviewPolicyRevision {
+		switch {
+		case currentPolicy.Revision > attested.PreviewPolicyRevision:
 			l.Warn().Int64("db_revision", currentPolicy.Revision).Int64("vmd_revision", attested.PreviewPolicyRevision).
 				Msg("daemon preview policy behind the database after resume; pushing")
 			if !reapplyPolicy() {
+				return "", false
+			}
+		case resumePolicy.requiresBrowserCapability():
+			// The policy is the claim's. The host's browser heartbeat can
+			// lapse during the boot, so re-check it before activation the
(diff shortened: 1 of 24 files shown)


meAmitPatil:
@codex review

pavitrabhalla (commented):
Three spots I want to flag before this lands.

internal/api/reaper.go
```diff
@@ -443,8 +443,11 @@ func (h *Handlers) rollbackPausedVM(ctx context.Context, sbx db.ClaimExpiredSand
 	// retry as the user-facing boots — the background ctx never reads dead,
 	// which is right: a rollback landing late still beats marking the
 	// sandbox failed.
+	// No policy carried: the record keeps the one it has, which this
+	// rollback never changed.
 	ipAddr, _, _, _, err := retryTransientBoot(ctx, sbx.ID.String(), sbx.HostID, func(rctx context.Context) (string, uint32, uint32, error) {
-		return vmd.ResumeInstance(rctx, sbx.ID.String(), snapshotPath, memPath, sbx.NetworkConfig)
+		ip, vcpu, memMiB, _, rerr := vmd.ResumeInstance(rctx, sbx.ID.String(), snapshotPath, memPath, sbx.NetworkConfig, "", nil, 0)
```
pavitrabhalla: This drops the new attestation return value. The resume handler in handlers.go checks `NetworkRulesApplied` and calls `reapplyNetworkConfig` when it's false, since a retried resume that adopts an already-running VM can land with only the firewall rules applied (`applyAdoptedNetworkRules`). Rollback doesn't do that check. Is a rollback landing on that adoption case expected to leave the sandbox without egress-proxy rules, or should this also inspect the attestation?

internal/vm/grpc_adapter.go
```diff
@@ -95,7 +95,27 @@ func (a *GRPCAdapter) ResumeVM(ctx context.Context, req *vmdpb.ResumeVMRequest)
...
+			return nil, status.Errorf(codes.InvalidArgument, "preview_access must be empty, %q, %q or %q, got %q", preview.AccessLegacyPublic, preview.AccessPublic, preview.AccessPrivate, previewAccess)
+		}
+		previewPorts, perr := previewPortsFromProto(req.GetPreviewPorts())
+		if perr != nil {
+			return nil, perr
+		}
+		if previewPortsContainTokenizedAccess(previewPorts) && req.GetPreviewPolicyRevision() <= 0 {
+			return nil, status.Error(codes.InvalidArgument, "tokenized preview policy requires a positive preview_policy_revision")
+		}
+		if err := a.mgr.UpdateSandboxPreviewPolicy(req.GetVmId(), previewAccess, previewPorts, req.GetPreviewPolicyRevision()); err != nil {
```
pavitrabhalla, line 113: This runs `UpdateSandboxPreviewPolicy`'s exact-revision-match check on every resume that carries a policy, unlike the existing stateless-restore path (`previewPolicyForRestore`), which is lenient at equal revision. A mismatch here returns `FailedPrecondition`, not `NotFound`, so `isVMDNotFound` won't catch it and the stateless fallback in handlers.go never triggers. Is a hard resume failure the intended behavior for a same-revision content mismatch, or should this fall back the way restore does?

internal/api/handlers.go
```diff
@@ -1044,68 +1053,118 @@ func (h *Handlers) resumePausedSandbox(c *gin.Context, sandbox *db.Sandbox, team
...
+		// committed after the claim and whose own push failed. One revision
+		// read proves what is current and names the policy the reply
+		// reports; only a daemon behind it gets the full read and push.
+		currentPolicy, policyErr := h.loadPreviewPolicy(postCtx, sandboxID, teamID)
+		if policyErr != nil {
+			failPost(policyErr, "reload preview policy after resume failed")
+			return "", false
+		}
+		effectivePolicy = currentPolicy
+		if currentPolicy.Revision > attested.PreviewPolicyRevision {
```
pavitrabhalla: This revision comparison gates whether `reapplyPolicy()` runs, and `reapplyPolicy()` is also what re-validates the host's browser-proxy heartbeat via `validateHostPreviewBrowserCapabilities`. When the attested revision matches the DB (the common case), this branch is skipped, so a host that loses browser-auth capability during the resume RPC (which can run up to 2×vmdBootTimeout) never gets re-checked before the sandbox goes active. The deleted comment above this block described exactly this scenario. Should the browser-capability check run independent of whether the revision moved?

internal/api/reaper.go
meAmitPatil: Good catch, fixed in 9b928e7b. The rollback now keeps the attestation and pushes the rules itself when the daemon did not ack them, failing the rollback like its other steps if the push fails. Test covers both the unacked and acked cases.

internal/vm/grpc_adapter.go
meAmitPatil: Agreed, a hard failure would repeat on every retry. Fixed in 9b928e7b: a same-revision mismatch keeps the record's policy at that revision, logs it, and still attests, matching the restore path. Test added.

internal/api/handlers.go
meAmitPatil: Right, the re-check was lost in the common case. Fixed in 9b928e7b: when the revisions agree and the policy needs browser capability, the host is re-validated against the database before activation, as the reapply did. Test covers a host that loses the capability during the boot.

pavitrabhalla (approved):

PR #487, opened 2026-09-09
infra: add a second staging vmd host with its own data disk

meAmitPatil:
Adds `superserve-vmd-staging-2` alongside the existing staging host — same image and shape, own 500 GB data disk, labeled `component=vmd` so every label-discovering deploy reaches both. It self-registers as `provisioning` and stays out of placement until an operator activates it.

The startup script pins this host's own `HOST_ID` from instance metadata, formats and mounts the data disk, and sets the vmd runtime flags both hosts run with. Staging only; production environments are untouched.

### DIFF
--- a/infra/envs/staging/us-central1/main.tf
+++ b/infra/envs/staging/us-central1/main.tf
@@ -399,31 +399,54 @@ module "sandbox_host_b" {
   can_ip_forward = true
 
   metadata = {
+    # cloud-init runs bootcmd on every boot before any service starts, so
+    # the identity is already this host's own by the time vmd can launch.
+    # A daemon that came up under another host's HOST_ID would heartbeat
+    # as that host and reconcile its sandboxes; stopping it afterward is
+    # too late.
+    user-data = <<-EOT
+      #cloud-config
+      bootcmd:
+        - |
+          NAME=$(curl -sf -H 'Metadata-Flavor: Google' http://metadata.google.internal/computeMetadata/v1/instance/name) || exit 0
+          for f in /etc/sandbox/vmd.env /etc/superserve/vmd.env; do
+            [ -f "$f" ] || continue
+            if grep -q '^HOST_ID=' "$f"; then
+              sed -i "s/^HOST_ID=.*/HOST_ID=$${NAME}/" "$f"
+            else
+              echo "HOST_ID=$${NAME}" >> "$f"
+            fi
+          done
+          mkdir -p /run/sandbox && touch /run/sandbox/host-identity-pinned
+    EOT
+
     startup-script = <<-EOT
       #!/bin/bash
-      # Runs on every boot. Everything here is idempotent.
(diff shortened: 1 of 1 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
## Verdict
APPROVE

I did not find a remaining Critical or High defect in the current head (`e4b1818e`). The earlier first-boot identity and disk-attachment findings are addressed in the current implementation, and the Terraform plan is non-destructive for the host change.

## Review scope
Generic review: the PR does not link an explicit Linear ticket/spec. Sequential passes covered hostile/adversarial behavior, security/resilience, test quality, correctness/maintainability, compatibility/rollout, observability, and infrastructure-plan safety.

## Validation
- Rechecked every existing review thread against the current head rather than trusting resolution state. The bootstrap now updates both legacy/current env locations, pins the instance identity in early cloud-init, avoids disabling the deployed service on later boots, and makes the data-disk mount device-triggered instead of relying on a one-shot attachment wait.
- Verified the existing host-registration path self-registers a previously unknown host as `provisioning`, so the new VM does not enter placement merely by starting VMD; operator activation remains a separate step.
- Verified `component=vmd` is present at instance creation, matching the current label-discovering deploy path, and staging deploy config already supplies the dedicated-disk `BACKUP_JOURNAL_PATH` / `BACKUP_STAGING_DIR` settings. The deploy script fails before host mutation if those paths are still on the root filesystem.
- Reviewed first boot, delayed disk attachment, reboot after deployment, cloud-init failure fallback, deploy-before/after disk appearance, and host activation sequencing.
- Terraform Checks, Terraform Plans, and CI are green on the head commit.
- Inspected the actual staging Terraform plan: the intended host work is three creates (`sandbox_host_b`, its 500 GB data disk, and the attachment), with **0 destroys**. The plan also contains five in-place monitoring-dashboard diffs unrelated to this PR; they appear to be existing/provider-normalization drift rather than host resources, so I would inspect/accept that drift when applying rather than treating it as part of this change.

## Reviewer execution
| --- | --- | --- |
| Test quality | sequential fallback | PASS with rollout-validation caveat |
| Infrastructure plan safety | sequential fallback | PASS with unrelated-plan-drift caveat |

## Activation caveat
Before changing the new host from `provisioning` to `active`, verify its first heartbeat is under `superserve-vmd-staging-2`, `/mnt/sandbox-data` is the attached PD (not the root filesystem), the deploy has installed the current VMD units/config on that host, and the expected capability/health signals are present. In particular, cloud-init `bootcmd` is very early but is not by itself a universal systemd ordering barrier for arbitrary third-party services; the current image/bootstrap shape makes the identity pin reasonable, but first-heartbeat verification is the appropriate fail-safe before activation.

PR #488, opened 2026-09-09
api: keep asking the host until a pause is decided instead of reverting on a timeout

meAmitPatil:
A pause that outran its 30s attempt and one 30s retry was reverted to active even though the host went on to complete it, leaving the row and the VM disagreeing. Pause now gets a 60s attempt and keeps repeating it, parking behind the pause in flight, until the host answers definitively or a 10 minute budget is spent; only then is it reverted.


meAmitPatil:
@codex review

PR #490, opened 2026-09-09
deploy-vmd: prime the runner ssh key before the parallel host fan-out

meAmitPatil:
With two labeled hosts in a cell, the parallel per-host `gcloud compute scp` calls race to create the runner's SSH key on first use and the loser fails with `google_compute_engine already exists`. One sequential connection before the fan-out creates the key once; single-host cells are unaffected beyond one extra round trip.

### DIFF
--- a/.github/workflows/scripts/deploy-vmd.py
+++ b/.github/workflows/scripts/deploy-vmd.py
@@ -333,6 +333,23 @@ def main() -> int:
 
     print(f"Deploying VMD to {len(instances)} instance(s) in {where}")
 
+    # Prime the runner's gcloud SSH key with one sequential connection before
+    # the parallel fan-out. On a fresh runner the key does not exist, and
+    # every `gcloud compute scp`/`ssh` that finds it missing runs ssh-keygen
+    # to create it; two hosts starting together both do, and the loser fails
+    # with "google_compute_engine already exists" before its bundle is even
+    # uploaded. A single-host cell never raced this. Any host will do — the
+    # key is per runner, not per host, and OS Login registers it account-wide.
+    first = instances[0]
+    run_or_die(
+        [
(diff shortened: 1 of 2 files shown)


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
## Verdict
APPROVE

I did not find any remaining Critical or High issues on the current head (`106a87a1`). The fix preserves per-host failure isolation while eliminating the shared runner SSH-key creation race before the parallel SCP fan-out.

## Review scope
Generic review of the deploy-script concurrency change and its regression coverage. I inspected the changed deploy path, surrounding per-host fan-out/error aggregation behavior, the existing socket/service ordering guards, prior review threads, and current CI.

## Validation
- The SSH key is now generated locally before `ThreadPoolExecutor` starts, so concurrent `gcloud compute scp` workers cannot race first-use key creation.
- The priming step no longer depends on any particular VM being reachable; one unhealthy host therefore cannot prevent healthy hosts from being attempted.
- On a fresh runner, `ssh-keygen -f ~/.ssh/google_compute_engine` creates the private/public pair synchronously before workers start.
- A local key-generation failure aborts before any host deployment begins, which is preferable to partially mutating the cell with an invalid shared SSH setup.
- Existing-key behavior remains a no-op; the deploy keeps the previous parallel per-host execution and aggregated failure semantics.
- Rechecked the prior P2 about losing `test_no_bare_service_stop`: the current test file preserves that regression guard and adds the SSH-key-before-fan-out assertion separately.
- Current CI and Terraform Checks are green.

## Reviewer execution
| --- | --- | --- |

No infrastructure-plan specialist was triggered because this PR changes deployment tooling only; it does not modify Terraform/resource definitions.

PR #491, opened 2026-09-10
deploy-proxy, deploy-otel-collector: prime the runner ssh key before the parallel host fan-out

meAmitPatil:
Same fix as the vmd deploy, for the proxy and collector deploys: create the runner's gcloud SSH key locally before the per-host pool starts, so parallel first-use ssh-keygen runs can't collide. Adds an ordering test for each script.


meAmitPatil:
@codex review

pavitrabhalla (approved):

PR #493, opened 2026-09-10
api: durable pause operations, and a background reconciler for pauses the caller gave up on

meAmitPatil:
**Problem.** When the host did not answer a pause in time, the API reverted the row to `active` without knowing what the host had done. A VM that had in fact paused was then billed and reported as active, and a second pause could be dispatched while the first was still running on the host. Callers also had to hold the connection for the whole pause.

**Fix.** Each pause is a durable operation on the row (id, start time, lease, trigger, actor). Finalize, revert, and mark-failed are fenced on that lease, so only the worker holding it can change the row. Once dispatched, a pause is never reverted: the row stays `pausing` with its lease released, a background reconciler re-asks the host every 30s, fails the row only on NotFound or FailedPrecondition, and flags anything pending past 30 minutes. Billing stops as soon as the row is `pausing`. `Prefer: respond-async` returns `202 {"status":"pausing"}` at once; without it the request holds until the pause is decided, and answers 202 if the host has not replied by then. Only a row known to be back at `active` is reported as failed.

Stacked on #510, which ships the columns and the reconciler first; until it merges this diff includes it. Rolling this PR back lands on a release that still reconciles any pause left in flight.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (commented):
High: the pre-dispatch revert still does not guarantee the documented failure contract. `revertPause` retries `RevertPauseToActive` three times, but if all writes fail it returns while leaving the row `pausing`; `dispatchPause` then reports `pauseUnresolved`, and the synchronous HTTP path returns 500. Once the lease expires, the reconciler can claim that same operation and actually dispatch `PauseInstance`. That means a caller can receive a definitive failure for a pause that is later executed anyway. The code comment currently states both that the row “must be back to active before the caller hears failed” and that, after retries fail, the reconciler will complete the pause—those are incompatible contracts. Make the pre-dispatch outcome durable (cancelled/not-dispatched), keep retrying until the fenced revert is known to have landed, or return an accepted/indeterminate outcome rather than 500 when the revert cannot be confirmed.

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (commented):
## Verdict
COMMENT — 2 High findings

I reviewed the current head (`c36e92d5`) twice: first manually without the review skill, then independently with the PR-review skill. The structured pass confirmed the manual correctness issue and added a rollout/version-skew issue. CI and Terraform Checks are green, but I would fix both before landing.

## High — Do not mark an undispatched automatic pause failed when the revert write fails

`pauseClaimed` only calls `revertToActiveOrFail` when `vmdForHost` failed, so **no pause RPC was dispatched and the VM is known to still be running**. `revertToActiveOrFail` correctly tries the fenced `RevertPauseToActive`, but if that DB write fails it immediately calls `failPause` under the same lease.

That creates a reachable split-brain state during a partial DB outage:

1. timeout/billing worker claims an active sandbox → DB becomes `pausing`, billing interval closes;
2. host resolution fails before any VMD call;
3. `RevertPauseToActive` hits a transient DB error;
4. the DB recovers enough for `failPause` to succeed;
5. row becomes `failed` and secret bindings are cleared, while the VM was never contacted and is still running.

This loses lifecycle/billing truth and can leave untracked compute running. The user-facing pause path now handles this case more safely: if its pre-dispatch revert cannot be written, it leaves the durable operation `pausing`, returns/retains the accepted state, and lets reconciliation converge. The automatic timeout/billing path should use the same invariant rather than converting a known-running VM to `failed` merely because compensation storage failed.

There is also no test for the transient-DB-error branch of `revertToActiveOrFail`; the new cancellation test only proves the revert succeeds when given a detached context.

## High — Define a rollback-safe handoff for durable `pausing` rows

This change introduces durable pause state whose recovery depends on the new background reconciler. The new binary calls `StartPauseReconciler`, but the current base release does not. The migration is additive, so rolling the application binary back is mechanically possible — but operationally unsafe once a new-style pause has been left `pausing` after an undecided host RPC or failed finalize.

A concrete rollback sequence is:

1. new API records a pause operation and returns `202` (or the synchronous request ends undecided);
2. row remains `pausing` for the reconciler;
3. deployment is rolled back before reconciliation finishes;
4. old API starts successfully because the schema is backward-compatible, but it has no `StartPauseReconciler` and its timeout/billing workers only claim `active` rows;
5. the sandbox can remain `pausing` indefinitely, with billing already stopped and resume returning conflict.

Because the PR explicitly specifies `migration first, then API`, rollback/version skew is part of the delivery contract. Please either make reconciliation available in a rollback-compatible precursor, or add an enforced rollback procedure/gate that drains or resolves every durable pause before the old binary can be restored. A prose warning alone is weaker than ideal because a normal emergency rollback is exactly when this invariant is easiest to miss.

## Other review results

- Fencing on pause operation ID + lease version is otherwise coherent across finalize, release, revert, and mark-failed.
- The earlier pre-dispatch user-facing race is fixed: a failed revert now returns `pausing` instead of reporting failure and pausing later unexpectedly.
- Undecided RPCs remain durable rather than being incorrectly reverted to `active`.
- Lost finalize replies are re-read via `pauseLanded`, avoiding duplicate/missing completion side effects in the reviewed paths.
- Backup-report gating now tracks recently held pause leases and stays retryable for long-running reconciliations.
- Timeout/billing candidate listing + point claiming avoids the earlier repeated fleet-wide scan issue; bounded refill and worker concurrency look reasonable.
- `Prefer: respond-async` now removes host resolution/RPC work from the request path as documented.
- Existing telemetry/logging is adequate to diagnose retries, attention-threshold pauses, terminal host responses, and reconciliation outcomes.

## Reviewer execution

| --- | --- | --- |
| Security/resilience | sequential fallback | FINDINGS |
| Test quality | sequential fallback | FINDINGS (coverage of the first defect) |
| General correctness | sequential fallback | FINDINGS |
| Data correctness | sequential fallback | FINDINGS |

The two findings are independent: fixing the automatic compensation path does not make rollback safe, and adding rollback gating does not fix the live-VM/failed-row split.

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

internal/api/backups.go
```diff
@@ -29,11 +29,16 @@ var errFinalizeInFlight = errors.New("pause finalization in flight")
...
+// A transition without a pause operation is bounded so a stranded one cannot
+// head-of-line-block the host's outbox. One carrying an operation stays in
+// flight until the operation is resolved, however old its lease: the
+// reconciler finishes it, and only then can a report be matched to the
+// snapshot it describes.
+func finalizeInFlight(row db.LockSandboxRowRow) bool {
+	if row.Status != db.SandboxStatusPausing && row.Status != db.SandboxStatusResuming {
+		return false
+	}
+	return row.PauseOpID.Valid || time.Since(row.UpdatedAt) < 10*time.Minute
```
pavitrabhalla, line 41: `64a6eb47` makes this unbounded by design for any row carrying an operation. The consequence is larger than one sandbox deferring its report.

`flushNotifications` (`internal/backup/uploader.go:486`) stops the entire batch on the first delivery failure, and `PendingNotifications` walks bolt from `c.First()` in key order on every flush. A single sandbox parked in `pausing` is therefore re-hit at the head of the outbox on every pass, and every other generation on that host queues behind it, retrying every 30s with no dead-letter.

Nothing time-bounds that state. The reconciler's only terminal outcomes are finalize-success and `failPause` on NotFound/FailedPrecondition; every other error is `releasePauseLease` and retry. So a host that keeps returning a retryable error for one VM, or one `vmdForHost` cannot resolve while the host itself still posts reports, stalls that host's whole backup-coverage pipeline indefinitely.

The correctness argument in the comment is per-sandbox: a report cannot be matched to the snapshot it describes until the operation resolves. Does that require a 503 that stops the batch, rather than skipping this sandbox's entry and continuing the flush?

Minor: `pause_op_lease_until` and `pause_op_attention_at` are both still selected by `LockSandboxRow` and now unread here. `pause_op_attention_at` is the 30-minute stranded marker if you want a bound back on the API side.

internal/api/pause_dispatch.go
```diff
@@ -0,0 +1,154 @@
...
+// respondPause answers a dispatch the caller waited for. Only a row known to
+// be 'active' again is reported as failed; one still 'pausing' is accepted,
+// since the reconciler will pause it.
+func respondPause(c *gin.Context, o pauseOutcome) {
+	switch o {
+	case pauseDone:
+		c.Status(http.StatusNoContent)
+	case pauseGone:
+		respondError(c, ErrSandboxGone)
+	case pauseUndecided:
```
pavitrabhalla, line 53: `pauseUndecided` answers 202, and `lifecycleResult` (`internal/api/telemetry.go:186`) scores any 2xx as `ResultSuccess`. During a host pause outage `sandbox_transition{operation="pause",result="error"}` now reads zero where it previously showed 500s.

Paging still works: `infra/modules/observability/sandbox-lifecycle-alerts.tf:117` matches `result=~"error|timeout"` across all operations, and the reconciler emits `reconcile_pause`/`error`. But the signal moves to a different `operation` label and lags by at least one reconcile cycle.

Is anything pinned to `operation="pause"` in the dashboards or the pause runbook that should move with it?

pavitrabhalla (commented):
Reviewed at `64a6eb47`, the incremental diff on top of merged #510.

Verified as correct: the lease fencing on `FinalizePause` / `RevertPauseToActive` / `MarkSandboxFailed` / `ReleasePauseLease` closes the double-dispatch window, since `attemptDeadline` reserves `asyncTimeout + pauseLeaseSkew` so a caller's attempts always end before `ClaimPendingPause`'s `min_age` gate opens; `isVMDFailedPrecondition` is genuinely terminal (both `FailedPrecondition` sites in `Manager.PauseVM` are unrecoverable); `FOR UPDATE OF s SKIP LOCKED` is scoped so the CTE doesn't trip Postgres; and the inlined actor-inheritance subquery in `RevertPauseToActive` is a faithful port of `GetMostRecentClosedSandboxIntervalActor`.

One question on the `finalizeInFlight` change in `64a6eb47` plus two notes, inline.

Separately: `GetMostRecentClosedSandboxIntervalActor` (`db/queries/sandbox_active_intervals.sql:49`) lost its last caller when `openSandboxIntervalInheritActor` was deleted, and is now dead.

internal/api/billing_reconciliation.go
```diff
@@ -109,35 +113,36 @@ func (h *Handlers) reconcileActivatedSandbox(ctx context.Context, teamID uuid.UU
...
-				log.Warn().Str("team_id", teamID.String()).Msg("billing: reconciliation batch limit reached")
-			}
-			return
+	// Claims stop with ctx; a pause already in flight finishes on its own
+	// detached budget.
+	cleanupCtx := context.WithoutCancel(ctx)
+	claimed, err := claimBatch(ctx, billingPauseWorkers, billingPauseCap, func(ctx context.Context, limit int32) ([]uuid.UUID, error) {
+		return h.DB.ListBillingIneligibleSandboxes(ctx, db.ListBillingIneligibleSandboxesParams{TeamID: teamID, Limit: limit})
+	}, func(ctx context.Context, id uuid.UUID) (db.ClaimBillingIneligibleSandboxRow, error) {
+		row, err := h.DB.ClaimBillingIneligibleSandbox(ctx, db.ClaimBillingIneligibleSandboxParams{ID: id, TeamID: teamID, LeaseSeconds: pauseLeaseSeconds})
```
pavitrabhalla, line 122: `ListBillingIneligibleSandboxes` and `ClaimBillingIneligibleSandbox` are only exercised through `mockDBTX`, unlike the reaper's equivalent list+claim pair which has integration coverage. A runtime SQL failure in either would first surface as a Stripe cancellation webhook that pauses nothing and retries every 30s.

Any reason not to add the integration test alongside the reaper's?

internal/api/backups.go
meAmitPatil: Agreed the batch stall was the real problem, and it was host-side: in d8b1aad4 the reporter surfaces `finalize_in_flight` as a per-entry deferral and the flush skips that entry and carries on, so one unresolved pause no longer holds the outbox. The gate stays per-sandbox until the operation resolves, and the two unread columns are gone.

internal/api/pause_dispatch.go
meAmitPatil: Nothing in infra/ or the runbook pins `operation="pause"`; the alert matches `result=~"error|timeout"` across operations. Still, d8b1aad4 records an undecided 202 as a `timeout` on the pause label, so the signal neither moves nor lags a cycle.

internal/api/billing_reconciliation.go
meAmitPatil: No reason. Added in d8b1aad4 next to the reaper's: the ineligible team's sandbox is listed and claimed with trigger `billing_ineligible`, the eligible team's is left alone, and a second claim finds nothing.

meAmitPatil:
Dead `GetMostRecentClosedSandboxIntervalActor` removed in d8b1aad4.

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
Re-reviewed the current head (`9dbfd07d`) with the full PR-review workflow. **APPROVE** — no remaining Critical or High findings.

The High from my previous pass is fixed: an automatic pause whose host cannot be resolved no longer marks a known-running VM `failed` when the fenced revert write fails. It leaves the durable operation `pausing`, hands back the lease, and reconciliation decides from the host; there is now a regression test for the unwritable-revert path.

The earlier rollback/version-skew concern is also resolved structurally now that #510 has landed in the base: the reconciler exists before this PR starts creating durable pause operations, so rolling this change back leaves a release capable of completing in-flight operations.

I also rechecked the later review rounds rather than relying on resolved threads. Current code safely handles ambiguous BeginPause replies without stealing a reclaimed lease, makes explicitly released operations claimable on the next reconciler tick, records async pause telemetry from the eventual detached dispatch rather than HTTP acceptance, preserves the backup report until the durable pause resolves without head-of-line blocking the host outbox, rotates deferred notification pages with bounded work, and has real integration coverage for both timeout and billing-ineligible list+claim paths.

- Hostile/adversarial: PASS
- Security/resilience: PASS
- Test quality: PASS
- Correctness/maintainability: PASS
- Observability: PASS
- Compatibility/rollout: PASS
- Data correctness: PASS
- Documentation/OpenAPI hygiene: PASS

CI and Terraform Checks are green.

PR #494, opened 2026-09-10
api: Prefer: respond-async lets a pause return 202 pausing instead of holding the connection

meAmitPatil:
With `Prefer: respond-async`, a pause the host has not finished within 20s returns `202 {"status":"pausing"}` with `Retry-After: 1`, and the caller polls the sandbox until it is `paused`; the host call keeps running on its own deadline and records its answer exactly as before. Without the header nothing changes: the request holds until the pause is decided.

Based on the pause-operation branch; retarget to main once that merges.


meAmitPatil:
Folded into the pause-operation PR.

PR #496, opened 2026-09-10
resume and create: one fewer database round trip in the lookup stage

meAmitPatil:
The explicit resume endpoint claims the paused row as its first statement instead of reading it first; the row is read only when the claim finds nothing to claim, to tell a pause still settling from a real conflict or a missing sandbox, and a settled pause is claimed again. The settle wait still counts as lookup.
Create resolves secret references while the template is read, since neither needs the other, and joins the result once the template is in hand.


meAmitPatil:
@codex review

internal/api/handlers.go
```diff
@@ -2378,12 +2410,23 @@ func (h *Handlers) CreateSandbox(c *gin.Context) {
...
+		meta     []SecretBindingMeta
+		err      *AppError
 	}
+	// The context is taken here, not in the goroutine: gin recycles c once
+	// the handler returns, and a template failure can return before the
+	// goroutine has even started.
+	secretsCtx := c.Request.Context()
+	secretsCh := make(chan resolvedSecrets, 1)
+	go func() {
+		bindings, meta, appErr := h.resolveSecretBindingsForCreate(secretsCtx, teamID, req.Secrets)
```
pavitrabhalla, line 2427: If the template lookup fails while this query is still in flight, the request context is canceled when the handler returns. The resolver then logs "DB GetSecretsByNames during sandbox create" at error level for a cancellation, and pgx v5 discards the pooled connection on cancel. Only reachable when the create names secrets and the template is bad. Worth skipping the error log in the resolver when ctx.Err() is already set?

pavitrabhalla (approved):
Claim-first is sound: ClaimResume already filters on paused and destroyed_at and returns the embedded row, so the snapshot check moving after the claim is fine. One non-blocking nit inline.

PR #501, opened 2026-09-11
fix(infra): enforce TLS 1.2 on HTTPS load balancers

awille-ai:
Production HTTPS frontends currently have no explicit SSL policy. Add one shared `TLS_1_2` / `MODERN` policy and attach it to all five existing Target HTTPS Proxies. Adopt the four proxies missing from Terraform into the production us-east4 state, preserving their identities, certificates, URL maps, and forwarding behavior. Both reusable HTTPS LB modules now require an explicit policy input.

Validation:
- Terraform formatting and all four supported environment validations pass; the unused reusable HTTPS LB module also validates.
- Read-only GCP inventory and production state checks confirm the frontend mappings and existing ownership.
- Saved plan comparison confirms four imports, one policy creation, and five in-place proxy updates changing only `ssl_policy`, with no replacements or deletions.
- The TLS-scoped plan contains only those changes. The full production plan also contains an unchanged, pre-existing update to `module.cloud_ids.google_compute_packet_mirroring.this`; review that separately before applying.

No infrastructure has been applied. Terraform import blocks perform the four proxy adoptions during the normal apply.

Post-deployment verification:
- Confirm all five proxies reference the shared policy with `TLS_1_2` and `MODERN`.
- Test each frontend IP using a real certificate-covered hostname as SNI: TLS 1.0/1.1 must be rejected by the server, while TLS 1.2/1.3 succeed. Use a client capable of attempting the older protocols; client-side protocol errors do not prove server rejection.
- Exercise normal HTTPS requests with certificate verification enabled using `curl --resolve HOST:443:IP https://HOST/`, then re-run the compliance scan.


awille-ai:
@codex review

meAmitPatil (approved):

PR #502, opened 2026-09-11
infra: east standby host with first-boot storage, artifact, dns, and env bootstrap

meAmitPatil:
Adds a second sandbox host to the east cell as a labeled-out standby under a new identity, with its own runtime service account, and moves the host preparation that was manual into first-boot metadata: package install, ip_forward, local-SSD RAID with XFS reflink bind-mounted under the canonical paths, the background-data disk (blank-only format, otherwise mount or refuse), kernel and base rootfs fetched from the cell's hostprep prefix and verified against pinned digests, the fleet's published Firecracker release verified against the fleet-approved digest, unbound, and a vmd.env with every non-secret key. Secrets, the egress policy, and the secretsproxy CA remain operator steps. Existing hosts are unaffected: the module's external-address flag defaults off and metadata is ignored after creation.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

pavitrabhalla:
The bootstrap comment in `deploy/host-bootstrap/sandbox-host-bootstrap.sh.tftpl` says a host prepared before a later fleet rollout is "brought current by running that deploy against this host alone before it joins discovery." I don't see a way to do that: `deploy-vmd.py` and `deploy-firecracker.py` discover by `VMD_LABEL` (`component=vmd`) plus region only, with no per-host override. With `component=vmd-use4-standby` the host is correctly outside every fan-out, but the only way to give it its first vmd deploy (CONTROL_PLANE_URL / DATABASE_URL / INTERNAL_API_TOKEN upserts, units, boxd inject) is to relabel it `component=vmd`, which enrolls it in the vmd, proxy, firecracker, and otel rollouts alongside the serving host in one step.

Which is intended: an explicit instance override in the deploy scripts, or the relabel as the documented step? If the latter, the comment should say that. Not blocking; the plan and the rest of the bootstrap check out against the live host table and hostprep objects.

pavitrabhalla (approved):
Plan is clean (6 add, 1 pre-existing drift, 0 destroy), the standby is correctly outside the component=vmd fan-outs, and HOST_ID, capacity, and the hostprep artifacts line up with the live host table and bucket. The first-deploy question in my comment above is not blocking.

awille-ai (commented):
Given what we just learned from the staging Host 2 MWI rollout, I think this PR needs one architectural change before we use it to create the east standby.

The new east standby should be created with Managed Workload Identity / workload identity certificates enabled in the VM create request, rather than creating the instance first and retrofitting MWI later. In staging, an existing VM showed `identityCertificateEnabled=true` in the control plane and “Enabled” in the console, but the guest metadata server still returned `Workload certificate feature not enabled` even after a full stop/start and explicit `gcloud compute instances update --identity --identity-certificate`. We’re now treating fresh-create-with-MWI as the safe path for new replacement hosts.

For #502 specifically, that means extending the shared `sandbox-host` module so the new east standby is born with the peer managed identity + identity certificates enabled, and making sure the east peer trust-domain / attestation / CA setup is in place before first boot. Since this is a brand-new host, east is the clean case where we can avoid the retrofit problem entirely.

A couple of related points from the same rehearsal:

- The standby label in this PR is correct: `component=vmd-use4-standby` matches the new manual deploy selector contract exactly.
- The expected standby hostname should be `superserve-vmd-use4` once this lands (`VMD_STANDBY_HOST_USE4=superserve-vmd-use4`).
- I’d also make the replacement reservation explicit rather than leaving `primary_reservation_name = null` in the production rollout path. Staging showed that stop/start without protected capacity can strand a host on zonal stockout. For this Z3 metal replacement, I’d want capacity secured before we depend on it.
- The new host should use the dedicated VMD runtime identity rather than `api_runner` if we’re aligning it with the multi-host peer identity model.

The rest of the standby/bootstrap shape here looks directionally right; I don’t think this needs a redesign. The main thing is to make this new east host the first production host that is created correctly with MWI from birth, while keeping `sandbox_status=provisioning` / `component=vmd-use4-standby` until it is fully verified and admitted.

awille-ai (commented):
Update to my earlier review: the MWI recommendation above is superseded. We subsequently reproduced the same `Workload certificate feature not enabled` failure on a completely fresh staging VM created with MWI enabled from birth, with the workload identity pool, managed identity, and exact-instance attestation all correct. We are no longer treating Compute MWI as a reliable prerequisite for the multi-host rollout; peer credentials are being provisioned through the Superserve/CAS provider while preserving the existing SS-470 TLS/SPIFFE contract.

I’d hold this PR until the current multi-host/credential-provider stack lands, then rebase it onto `main`. The staging rehearsal has also changed several fresh-host provisioning contracts that overlap this PR, so rebasing afterward should let east consume the tested path rather than duplicate assumptions that are moving underneath it.

The standby label/hostname and capacity-reservation comments from my previous review still stand.

awille-ai (changes requested):
I think this is the right PR to continue with for east, but I would not land the current branch as-is after the multi-host/identity work that has landed since it was opened.

The biggest blocker is durable host identity reuse. This PR creates the new host as `superserve-vmd-use4` at `10.2.0.2` (the old primary slot). With the current multi-host model, `host_id` is a durable ownership identity, not a reusable slot name. SS-199 (retire the old host row/services) is also still outstanding, so old sandbox ownership records can still refer to that identity. Please give the new machine a genuinely new stable identity/IP (for example a new `use4-N` host) rather than reusing the retired slot.

A few other changes should come along with the rebase onto current `main`:

1. Use the dedicated VMD runtime identity model introduced by #500 rather than attaching the shared `api_runner` service account to the new host. This is much easier to get right at VM creation than to retrofit later.
2. Make background-data initialization blank-only/fail-closed. The current bootstrap formats whenever the detected filesystem is not XFS, which can destroy a preserved disk containing an unexpected filesystem/signature. Blank -> format; existing expected XFS -> mount; anything else -> refuse.
3. Pin/verify the kernel and base-rootfs artifacts. The bootstrap currently fetches `vmlinux-4.14-fuse` and `base.ext4` and only checks that the files are non-empty. We have already had east snapshot-restore compatibility problems tied to host/kernel lineage; use immutable/versioned objects and/or expected SHA-256s.
4. Rebase on current `main` and normalize against the now-current peer identity / standby deployment contracts from #500/#492. For this rollout we intend to use the Superserve/manual peer-cert provider, not retrofit Google MWI onto a serving/legacy bare-metal host.
5. After the rebase, produce a fresh production `us-east4` Terraform plan and verify the existing serving `superserve-vmd-use4-2` has zero changes. This PR should only add the new standby and its supporting resources.

The overall first-boot approach is good and worth keeping: local-SSD RAID/XFS setup, canonical bind mounts, the separate `/mnt/sandbox-data` mount, artifact staging, Firecracker installation, and labeled-out standby provisioning are exactly the automation we wanted after doing west manually. I would update this PR rather than replace it.

meAmitPatil:
@codex review

meAmitPatil:
On the access-config note: that lifecycle entry predates this PR (it is on main today) and is what lets operators add a bootstrap address by hand on existing hosts without Terraform reconciling it away. This PR only adds the creation-time flag. Managing that address declaratively is tracked separately; leaving the existing behavior unchanged here.

The backup-bucket grant for the runtime identity now matches the cell contract: create and read within the cell, no delete or overwrite, same shape as the west identity.

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
APPROVE — reviewed in two independent passes (manual first, then the PR-review workflow) and consolidated here.

No Critical or High findings remain on the current head. I rechecked the destructive/bootstrap surfaces directly rather than relying on resolved threads: Local SSD vs persistent-disk classification, RAID reassembly, blank-only background-disk formatting, startup/IAM ordering, hostprep artifact integrity, pinned Firecracker digest, bucket restore access, external-IP/guest-DNS egress, dedicated runtime IAM, deployment-discovery labels, and the real production-us-east4 Terraform plan.

Infrastructure plan: 11 creates, 1 in-place update, 0 destroys. The single update is the existing Cloud IDS packet-mirroring empty-filter normalization; it does not mutate either serving host.

The earlier review findings are genuinely addressed on this head. CI, Terraform Checks, and Terraform Plans are green.

Low / non-blocking test-quality gap: `deploy/host-bootstrap/sandbox-localssd.sh` is safety-critical destructive shell (its output feeds mdadm/mkfs), but there is no automated fixture test for representative `lsblk` and `/dev/disk/by-id` layouts. Given the device-selection regressions caught earlier in this PR, a small test harness for local SSD aliases, attached Hyperdisk, surviving RAID metadata, and no-local-SSD failure would be valuable, but I would not block this head on it.

Rollout checks: because `sandbox-data.service` is started with `--no-block`, verify the service is active and `/mnt/sandbox-data` is actually mounted before treating the host as operationally prepared; also inspect the packet-mirroring normalization at apply time. The host remains labeled `sandbox_status=provisioning`, so neither check is bypassed by normal deployment discovery.

Reviewer execution (sequential fallback): hostile PASS; security/resilience PASS; test quality FINDINGS (Low only); correctness PASS; observability PASS; compatibility/rollout PASS; data/storage correctness PASS; documentation hygiene PASS; sandbox lifecycle PASS; infrastructure-plan safety PASS.

PR #503, opened 2026-09-12
feat(qm): add qm schema, qm_api role, and tenant queries

pavitrabhalla:
## Summary
Data model for hosted QM instances (one QM stack per customer team), kept in its own Postgres schema `qm` with a dedicated role `qm_api` so the future `qm-api` service can run without reaching the sandbox product's tables. No service uses it yet; this is the foundation the provisioner and `/v1/qm/*` API build on.

## Technical decisions
- `qm.tenants`, `qm.tenant_events` (append-only), `qm.tenant_secrets` (Secret Manager references only, never values). Slug is CHECK-constrained to a DNS label.
- `qm_api` has no BYPASSRLS; policies key on a transaction-local team scope (`set_config('qm.team_id', …, true)`), so an unscoped query sees nothing and an unscoped insert fails instead of leaking. Slug availability is answered by a `SECURITY DEFINER` boolean function so it spans teams without exposing rows.
- Least privilege on `public`: column-level `SELECT` on `team`, `team_memberships` (the canonical membership table), and `api_key`; no `DELETE` on tenants (retirement is a status) and no `UPDATE`/`DELETE` on the event log. Queries treat `deleted` as terminal so a racing status write cannot resurrect a retired tenant.
- A tenant may only reference an API key its own team owns: composite FK on `(sandbox_api_key_id, team_id)` backed by a unique index on `api_key(id, team_id)`, with `ON DELETE SET NULL (sandbox_api_key_id)` so the existing expired-key sweep keeps working. The index follows the repo's hot-table pattern: pre-build it `CONCURRENTLY` by hand before merging (instructions in the migration), and the migration validates the index and runs under bounded lock/statement timeouts.

## Testing
- `sqlc generate` clean; `GOOS=linux go build ./... && go vet ./...`.
- Integration tests (`-tags integration`) against a local Postgres: CRUD as `qm_api`, cross-team invisibility, denied reads of `sandbox`/`team_member`/`credential_store_config`/`key_hash`, denied hard deletes and event mutations, cross-team key rejection, key-deletion releases the reference, deleted stays deleted, invalid slugs, idempotent re-apply; plus the `cmd/migrate-team` suite.
- Local Codex review iterated to a clean pass.

## Before merging
Pre-build `api_key_id_team_unique` concurrently on staging and production (see the migration header); the migration is a no-op for the index once it exists and fails loudly if it is missing or invalid.


pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

supabase/migrations/20260912000001_qm_tenants.sql
```diff
@@ -0,0 +1,315 @@
...
+
+CREATE UNIQUE INDEX IF NOT EXISTS idx_qm_tenant_events_tenant_seq ON qm.tenant_events(tenant_id, seq);
+
+-- secret_ref is a Secret Manager resource name, never a secret value; the
+-- value only ever lives in Secret Manager and is resolved by the tenant's
+-- own runtime identity.
+CREATE TABLE IF NOT EXISTS qm.tenant_secrets (
+    tenant_id  uuid NOT NULL REFERENCES qm.tenants(id) ON DELETE CASCADE,
+    name       text NOT NULL,
+    secret_ref text NOT NULL,
```
awille-ai, line 148: **High — enforce the “Secret Manager reference only” invariant at the DB boundary.** SS-516 explicitly requires that no secret values are stored anywhere in `qm.*`, and this table documents `secret_ref` as a Secret Manager resource name, never a value. But `secret_ref text NOT NULL` plus `SetQMTenantSecretRef` accepts any string, so a provisioning bug could persist a raw model key, `DATABASE_URL`, Superserve API key, or other credential in plaintext Postgres. SS-521 makes those exact sensitive values inputs to this layer. Please constrain this column to the accepted Secret Manager resource-reference shape (or otherwise make raw values impossible to persist here) and add a negative integration test that a credential-like/raw value is rejected.

awille-ai (commented):
Reviewed twice: first independently in the context of the hosted-QM architecture/POC, then with the PR-review workflow against SS-516 and the related provisioning/deprovisioning contracts.

The current head is substantially stronger than the earlier Codex-reviewed revisions. The migration/cutover races, one-active-tenant invariant, event ordering, lock ordering, and child-write-vs-retirement cases now have explicit handling and strong integration coverage. The RLS/least-privilege shape also looks good, and both CI and Terraform Checks are green on this head.

**Verdict: COMMENT** — one High finding remains (inline): the schema says `tenant_secrets.secret_ref` can contain only Secret Manager references, and SS-516 explicitly requires that no secret values ever be stored anywhere in `qm.*`, but the column/query currently accept arbitrary text. Given SS-521 will pass model keys, DB URLs, Superserve API keys, etc. through this provisioning layer, I think that invariant belongs at the DB boundary rather than relying solely on caller discipline.

Reviewer execution: Requirements — FINDINGS; Hostile/adversarial — FINDINGS; Security/resilience — FINDINGS; Test quality — FINDINGS; Correctness — FINDINGS; Performance — PASS; Compatibility/rollout — PASS with caveat; Data correctness — FINDINGS (same root cause); Infrastructure-plan review — configuration inspected, environment state UNVERIFIED.

Non-blocking rollout caveat: this migration intentionally requires `api_key_id_team_unique` to be pre-built `CONCURRENTLY` in staging and production before merge/deploy. I can verify the migration fails boundedly if that step is missed, but I cannot verify the actual staging/prod index state from the PR evidence. Please treat that pre-build as a merge/deploy prerequisite.

pavitrabhalla:
Pushed a fix for the CHECK-constraint finding: `qm.tenant_secrets.secret_ref` now rejects anything that isn't shaped like a Secret Manager resource name, so a provisioning bug can't persist a raw model key, DATABASE_URL, or sandbox API key in plaintext Postgres. Verified the negative direction too — reverted the constraint locally and confirmed the new test fails without it, so it's actually exercising the guard rather than trivially passing.

The two multi-cell findings (slug uniqueness and admission-during-migration both being cell-local) are real — verified by reading `qm.slug_available()`, the tenants UNIQUE constraint, `QMTeamHomedHere`, and `runCopy`'s copy-before-detach ordering. Both need a design decision (a global slug registry or cell-qualified names; an authoritative cutover marker instead of inferring homing from a row migration deliberately duplicates) rather than a quick patch, and there's no existing cross-cell coordination primitive in the repo to reuse. Filed as SS-548 with the technical detail rather than guessing at a fix under time pressure.

@codex review

PR #504, opened 2026-09-13
feat(infra): add shared Terraform module for hosted QM

pavitrabhalla:
## Summary
Adds `infra/modules/qm`, the shared, tenant-independent infrastructure for hosted QM, and wires it into the staging root behind `enable_qm` (default off, so the staging plan stays a no-op). Per-tenant resources are created at runtime by the provisioner job, not by Terraform; this module owns everything a tenant service depends on.

## What the module creates
- Cloud SQL Postgres 16 (private IP on the environment VPC, PITR, deletion protection) plus its admin role and Secret Manager secrets for the admin password and the control-plane `DATABASE_URL`.
- Wildcard edge: DNS authorization, managed certificate for the apex and wildcard, certificate map, global HTTPS load balancer with the redirect service as its default route, and the `:80` to `:443` redirect. `host_rule` and `path_matcher` drift is ignored so applies never strip provisioner-added tenant routes.
- The qm-api Cloud Run service, the provisioner Cloud Run job, the redirect service, their service accounts, and an Artifact Registry repository for tenant images.

## Technical decisions
- Secret Manager bindings are project-level with `resource.name.startsWith` conditions on the `qm-` prefix, minus every environment's platform secrets (`qm-sql-admin-*`, `qm-api-database-url-*`) — both identities reach their own through explicit secret-level bindings, so excluding the whole family costs nothing and keeps two environments in one project from reading each other's credentials. Permissions that cannot be conditioned (secret create) live in tiny custom roles.
- The provisioner manages tenant databases through the SQL Admin API, so it gets a custom role limited to database and user CRUD rather than a broad Cloud SQL role. It also needs `run.jobsExecutorWithOverrides` and subnet-scoped `compute.networkUser` for Direct VPC egress on tenant services.
- The shared instance carries three independent destroy guards: `deletion_protection`, `deletion_protection_enabled`, and `prevent_destroy`. The third is what catches a *replacement* — a changed suffix, region, or private network plans destroy-then-create, which the first two do not make obvious at plan time. Teardown is deliberately a staged edit; the README has the order.
- Private Service Access is created by default and can be disabled where the VPC already has a servicenetworking peering, because updating an existing peering from Terraform would replace its range list. The reserved range is abandoned along with the peering, since GCP refuses to delete a range an active peering still holds.
- Terraform owns the redirect service's image (nothing in deploy tooling updates it) but ignores image drift on qm-api and the provisioner job, which the qm-api deploy workflow rolls together.
- Env names and tenant naming follow `internal/qm` and the provisioner steps so Terraform serves the code as written. The module exports the derived tenant slug length budget rather than leaving the Go side to recompute it. The README lists the reserved slugs, the one-environment-per-project constraint, and the first-apply ordering (images, secret versions, certificate DNS).

## Testing
- `terraform fmt -check -recursive infra/` clean; `scripts/terraform-validate.sh` passes for all four roots.
- Real read-only plan of the staging root against live state: zero `module.qm` resources, confirming `enable_qm = false` gates the module completely.
- The module planned standalone against the staging project with every input set: 62 to add, no errors, every attribute accepted by the provider.
- Local Codex review iterated to two consecutive passes with no findings.

## Note on CI
`terraform-plan (staging-us-central1)` is red for a reason that predates this branch and is not caused by it — the same failure reproduces on other open infra PRs. Staging state was written by a newer google provider than the lockfile pins, and the root also carries drift from out-of-band applies (a sandbox host that now plans as a replacement, plus CA and VMD resources in state with no config). Reproduced locally: bumping the provider clears the version error and then fails on the host replacement, so it needs its own change rather than being folded in here.

## Tenant runtime configuration (added later on this branch)

The module now also owns everything a provisioned tenant is configured with, so a deploy supplies none of it out of band and there is no second owner to fight the next apply (this module does not ignore environment drift on the service or job):

- `QM_RESEND_SECRET` — an empty `qm-resend-<suffix>` secret, the platform's one Resend key for the whole fleet. The provisioner holds a three-permission role on it (`secrets.get`, `getIamPolicy`, `setIamPolicy`): enough to grant and revoke tenant access, not enough to read the key or delete it. Both identities are excluded from it by exact name *and* from its `/versions/` subtree — a version's access check is evaluated against the version's own resource name, so an inequality on the secret alone is true for exactly the operation that reads the key. By name rather than prefix because a `qm-resend-` prefix would also swallow the secrets of a tenant slugged `resend`. No override: every tenant mounts this one by name, so a second supported location means an apply that moves the name can delete a secret the fleet is still reading. Rotation is a new version.
- `QM_EMAIL_FROM` — default `QM <no-reply@mail.qm.superserve.ai>`. Not derived from `var.domain`: the verified sending domain belongs to the Resend account, not to an environment, so staging and production share it.
- `QM_SANDBOX_API_URL` — no default; the staging root supplies `https://api-staging.superserve.ai`. The only one that genuinely differs per environment, because the sandbox key the provisioner issues a tenant is bound to the cell that issued it.
- `QM_SANDBOX_TEMPLATE` — default `qm-agent-0.1.0`.
- `QM_SANDBOX_KEY_REGION` — optional; null leaves issued keys untagged, which is today's behaviour. Added because it was the one qm-api setting reachable only by hand-editing the Cloud Run service, which this module would then silently revert.

Plus the storage HMAC-key role the provisioner needs: QM reaches its bucket through an S3 client, so every tenant gets an interoperability key. Without it, every real provision 403s at the bucket step *after* its secrets, identity and database already exist.

The apply order was also corrected: the unqualified apply used to come before the `DATABASE_URL` version existed, and both workloads mount it at `latest`, so that apply could not bring up a revision. The secret bootstrap is now its own step ahead of it.

### CI position — this PR is not green

`terraform-plan (staging-us-central1)` is **red**, and was already red on `ff7dc89a` before these commits. The plan is byte-identical on both — `Plan: 0 to add, 5 to change, 18 to destroy` — and the destroys are unrelated vmd peer / host2 / peer-CA drift; `module.qm` appears nowhere in either failure log. That identity is also the confirmation that `enable_qm = false` still plans empty.

`Terraform Fmt` and `Terraform Validate` pass. Variable validations were exercised against a real `terraform plan` (11 accept/reject cases). Local `codex exec review` clean over the whole change.

A fresh environment still needs a human for: a version on `qm-resend-<suffix>` and on `qm-api-database-url-<suffix>`, the certificate DNS authorization CNAME and the apex/wildcard A records, image tags that exist, and a service-account quota increase before scale.

### DIFF
--- a/infra/envs/staging/us-central1/.terraform.lock.hcl
+++ b/infra/envs/staging/us-central1/.terraform.lock.hcl
@@ -21,3 +21,26 @@ provider "registry.terraform.io/hashicorp/google" {
     "zh:ff18322b786ad6216ee3aebb02e6374223eaa16b5f181109089786128718afc3",
   ]
 }
+
+provider "registry.terraform.io/hashicorp/random" {
+  version     = "3.9.1"
+  constraints = ">= 3.5.0"
+  hashes = [
+    "h1:PYbnOqRuGn4c0/Ae1f7yOS/0zvmXNHFJRZjuF4KECnM=",
+    "h1:PlW+UZ4EElQF3NQwf41KQwavFujab3Czc51zu9dyVM8=",
+    "h1:g40qr7yDmIpaur4SsK5BcOda3HSo1RJ6zHVMqN4EJ+0=",
+    "zh:05f4734c1f0be840b711b3eff259ebc5fca436784c728955b1678078466f48d7",
+    "zh:0b91bf19371d012434eba1deeb6aab77158def9b39601dcbd94450b3974a2a26",
+    "zh:0ee6eacd47ec00183d55d726a4b6c4ce951a199f944bf22f1aa58392ebdfa7a2",
+    "zh:19388a4074b76a89a43a6c8328d7ae8ee2e7de3d346af51e80d3e6d3d12925f1",
+    "zh:23e74d48c5e2ac2e823fd527f49fee9db37d32a1990c9e3bf126ead697b843eb",
+    "zh:3cabf7fbd096c520064aae3aba61aba670af83ab91291a71fa1b1332929c2b7f",
+    "zh:5c0a3b8af0be60be4eca12ddee385cfa8babc1ec8e98cdf9de2f2274c73eabfa",
+    "zh:60b4f8a8ef18f52bf8e19215229dae408bee732825964092db7c989fd2de4097",
+    "zh:7359015acfedcbd6366f2329c854cf8d3c8ca5cd0faa89d2d37db358d6eba6c5",
+    "zh:78d5eefdd9e494defcb3c68d282b8f96630502cac21d1ea161f53cfe9bb483b3",
+    "zh:7b38758402f0e13a1071162da28994023cd2ac676e54af350c9ffd8dfa73fa7b",
+    "zh:7c7fbb8895eb75bb4de1f933e98553bd99c8d048c89a925ddba490aa5a67f7dc",
+    "zh:8c2b8c6a7ccdec16b73e2fb9f3700ea097f58c592571e4c5de60c93d2301732c",
+  ]
+}
--- a/infra/envs/staging/us-central1/main.tf
+++ b/infra/envs/staging/us-central1/main.tf
@@ -25,6 +25,15 @@ locals {
   region          = var.region
   zone            = var.zone
   resource_suffix = coalesce(var.resource_suffix, var.environment)
+  # Used by the qm module only; the existing staging accounts predate the
+  # suffix convention and keep their imported names.
+  service_account_suffix = coalesce(var.service_account_suffix, local.resource_suffix)
+
+  # deploy-qm-api.yml pushes the qm-api image (service and provisioner job
+  # share it) to the existing superserve repository; the placeholder tags
+  # below are overridden by qm_*_image in terraform.tfvars before the first
+  # apply. Tenant images live in the qm-<suffix> repository the module creates.
+  qm_image_prefix = "${local.region}-docker.pkg.dev/${local.project_id}/superserve"
 
   common_labels = {
     environment = local.environment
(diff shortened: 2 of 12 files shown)


pavitrabhalla:
Plan checks are red for a reason unrelated to this change: the staging state was applied from the in-flight host2 migration branch with a newer google provider than main's lockfile pins, so every plan against main's config currently fails with "Resource instance managed by newer provider version" and proposes destroying the peer CA / peer identity / VMD runtime resources that only exist on that branch. The same failure shows on the east-standby-host PR since 2026-09-12 18:29 UTC. This module adds nothing to the staging plan while `enable_qm=false`; checks should recover once the host2 migration lands and the provider pin catches up.

pavitrabhalla:
@codex review

pavitrabhalla:
Pushed the tenant runtime configuration the provisioner needs, so nothing is left for a deploy to supply out of band:

- **`QM_RESEND_SECRET`** — the module creates an empty `qm-resend-<suffix>` secret and hands the provisioner a three-permission role on it (`secrets.get`, `getIamPolicy`, `setIamPolicy`): enough to grant and revoke tenant access, not enough to read the key or delete it. Both identities are excluded from it by exact name, and from its `/versions/` subtree — a version's access check is evaluated against the version's own resource name, so an inequality on the secret alone is true for exactly the operation that reads the key. By name rather than by prefix because a `qm-resend-` prefix would also swallow the secrets of a tenant slugged `resend`. There is deliberately no way to point at a secret elsewhere: every tenant mounts this one by name, so a second supported location means an apply that moves the name can delete a secret the fleet is still reading. Rotation is a new version.
- **`QM_EMAIL_FROM`** — module variable, default `QM <no-reply@mail.qm.superserve.ai>`. Not derived from `var.domain`: the verified sending domain is a property of the Resend account, not of an environment, so staging and production share it.
- **`QM_SANDBOX_API_URL`** — module variable with no default; the staging root defaults to `https://api-staging.superserve.ai`. The one of the four that genuinely differs per environment, because the sandbox key the provisioner issues a tenant is bound to the cell that issued it.
- **`QM_SANDBOX_TEMPLATE`** — module variable, default `qm-agent-0.1.0`.

Also adds the storage HMAC-key role the provisioner needs: QM reaches its bucket through an S3 client, so each tenant gets an interoperability key. Without it every real provision 403s at the bucket step, after its secrets, identity and database already exist.

All of it is inside the module, so a plan with `enable_qm = false` is unchanged — zero resource, data or module blocks added under `infra/envs/`, and the tfvars diff is comments only. `terraform fmt -check -recursive infra/` and `scripts/terraform-validate.sh` pass on all four roots, and the variable validations were exercised against a real `terraform plan` (11 accept/reject cases). Local `codex exec review` clean over the whole change.

What a fresh environment still needs a human for: a version on `qm-resend-<suffix>` and on `qm-api-database-url-<suffix>`, the certificate DNS authorization, and the service-account quota. Apply order is in the module README.

@codex review

pavitrabhalla:
On the `terraform-plan (staging-us-central1)` failure: it is pre-existing, not from this change.

The same job failed identically on `ff7dc89a`, the commit before this push, with the same plan — `Plan: 0 to add, 5 to change, 18 to destroy` — and the destroys are all vmd peer / host2 / peer-CA resources. `module.qm` appears nowhere in either failure log.

That is also the confirmation I wanted that `enable_qm = false` still plans empty: the plan on this branch is byte-for-byte the plan on its base, so the module's new resources contribute nothing while the flag is off. Structurally the same thing — every new resource is inside `infra/modules/qm`, which the staging root instantiates with `count = var.enable_qm ? 1 : 0`, and zero resource/data/module blocks were added under `infra/envs/`.

pavitrabhalla:
Pushed two more:

- **Apply order fixed** — the unqualified apply used to come before the `DATABASE_URL` version existed, and both workloads mount it at `latest`, so that apply could not bring up a revision. The secret bootstrap (`-target` both secrets, then add versions) is now its own step ahead of it, and a step was added for tenants left behind by an environment that ran with `QM_PROVISIONER_STUB=1`.
- **`QM_SANDBOX_KEY_REGION`** is now an optional module variable. It was the one qm-api setting reachable only by hand-editing the Cloud Run service — which this module would silently revert on the next apply, since it does not ignore environment drift. Null leaves issued keys untagged, which is today's behaviour.

`terraform fmt -check -recursive infra/` and `scripts/terraform-validate.sh` still pass on all four roots. Local `codex exec review` clean over the whole change.

@codex review

PR #505, opened 2026-09-13
feat(qm): add the qm-api service and tenant provisioner

pavitrabhalla:
## Summary

Adds the hosted-QM control plane: a `qm-api` Cloud Run service serving `/v1/qm`, and a tenant provisioner that runs as a Cloud Run Job from the same image. The console calls `qm-api` to create, inspect, retry and delete a team's QM instance and to mint portal admin sign-in links; each create writes the model key straight to Secret Manager and queues a provision run that builds the tenant's stack.

This is the second half of the control plane, stacked on #503 (the `qm` schema, the `qm_api` role and the tenant queries). It brings the engine, the HTTP surface, the auth path and the deploy; the cloud steps themselves (service account, database, bucket, sandbox key, Cloud Run service, load-balancer route, probes) are declared here with their client interfaces but not yet implemented, so the binary refuses to start unless `QM_PROVISIONER_STUB=1`. The infra shape is one Cloud Run service, database, bucket, service account and subdomain per tenant — multi-tenant infra, single-tenant app.

- **Auth without widening grants.** `qm_api` has no read on `api_key.key_hash` and cannot see the RBAC tables, so authentication and authorization go through three definer functions (`qm.resolve_api_key`, `qm.actor_can_team`, `qm.touch_api_key`). Presenting a hash already proves possession, so the resolver exposes nothing the caller doesn't hold, and a key's holder gets exactly what their team role grants. `actor_can_team` mirrors `internal/authz`'s query and has to be kept in step with it — the cost of not granting the role those tables.
- **Provisioning is a plan of idempotent steps.** Deprovision is the same list reversed, each step running its `Rollback`, so teardown is written once per step. A failed provision is left in place for a retry to converge rather than rolled back; rolling a half-built stack back on a transient error turns one retry into a full rebuild.
- **One run per tenant, enforced in Postgres.** A run holds a 64-bit advisory lock in an open transaction for its whole life, on a separate small pool so run locks can never starve the status and event writes. A queued run that arrives after the intent changed finds a status it wasn't queued for and exits.
- **Ambiguous trigger failures stay in flight.** A 4xx from Cloud Run means no execution exists and the tenant is marked failed; anything else may have started one, so the tenant stays in flight and a stale-run reclaim (no events for 30 minutes, lock free) makes it retryable. Marking it failed immediately would let a retry race a run that is about to start.
- **The model key never travels through the job.** Cloud Run Job overrides are visible in execution metadata, so the key goes to Secret Manager from the request and only identifiers reach the trigger. Event detail is scrubbed before it is stored, since a cloud error can echo a connection string back.
- **Teardown revokes the tenant's sandbox key rather than just forgetting it.** That key is bound to this cell, and the cutover guard in #503 refuses a region migration while a tenant still points at a live one. `qm_api` has no `UPDATE` on `api_key` and should not get one, so the revoke goes through a definer function that can only reach the key the caller's own tenant references. Issuing the key belongs to the not-yet-implemented half; revoking it is real in every mode, so the guarantee is in place before the step that creates the obligation.
- **A delete is never left marked `failed` without its intent event.** The mode lives only in that event, so a tenant marked failed without one reads as a failed provision and a retry would rebuild the stack the caller asked to tear down. Those failures put the tenant back where the delete found it and ask the caller to issue it again.
- **Deploy updates the job before the service**, so a run the new service triggers can never land on an older provisioner, and it sets the provisioner's region on both resources from the region it is targeting rather than relying on a default.

## Testing

- Unit tests for the handlers (auth, permissions, validation, status races, stale reclaim, trigger-failure paths), the runner, the steps, the scrubber, the config and the admin-link codec — the last verified against a port of the portal's own verifier, so a minted link is provably one the tenant will accept.
- Integration tests run against Postgres as the `qm_api` role, so the grants, row policies and definer functions are exercised, not bypassed: the key resolver, the store's rules, a stub run end to end, and the HTTP API end to end in in-process mode.
- `go build`, `go vet`, `sqlc generate` and the deploy-trigger-coverage check are clean; the full unit and integration suites pass locally.


pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
@codex review

pavitrabhalla:
One note for whoever picks this up: the automated reviewer has re-posted "Declare 403 responses for read-protected routes" and "Declare 503 responses for all QM operations" on several successive commits. Both were addressed a while back — every operation under `/v1/qm` now declares 401, 403, 500 and 503 (plus 404/409/413/502 where they apply), and the `QMUnavailable` response component exists. The comments are anchored to lines that still exist rather than to missing content.

pavitrabhalla:
Same for "Retry the current attempt after a stale holder releases the lock" — the job entrypoint has waited out a lock holder since an earlier commit (`lockedRunRetries` / `lockedRunBackoff` in `cmd/qm-api/main.go`), on the reasoning the comment describes: a superseded execution releases within a query or two, so the replacement should wait rather than exit as though the work were done.

Two things I left alone on purpose:

- The migration gate in `deploy-qm-api.yml` is a deliberate copy of `deploy-api.yml`'s, limits included — it inspects the pushed range, so it does not re-check a migration an earlier push left failing. Both services deploy against the same database and a gate that differs between them is worse than one that is uniformly conservative; changing it is its own PR, touching both.
- CI does not run on this PR: `ci.yml` triggers on `pull_request: branches: [main]`, and this is based on #503. Everything it would run was run locally against a real Postgres — unit and integration with `-race`, vet, the binary builds, `sqlc generate`, the deploy-script tests, and the deploy-trigger-coverage check.

pavitrabhalla:
@codex review

PR #507, opened 2026-09-13
feat(qm): provision hosted-QM tenants against real GCP, with per-tenant resource ownership

pavitrabhalla:
## Known blocker — do not issue tenant keys in production yet

The tenant's `SUPERSERVE_API_KEY` is issued with `created_by` set to the tenant's creator, because that is how the control plane resolves what a key may do. The creator is usually a team owner, and `internal/api` aliases sandbox permissions onto `settings:read` / `settings:write` (`rbac_phase3.go`), so a tenant container whose key leaks can reach the team's secrets, templates and management endpoints — not only the sandboxes it needs. The empty `scopes` array does not help: ordinary keys are authorized from `created_by`, not from scopes. A secondary consequence: a tenant whose creator leaves the team loses the ability to create sandboxes.

The same root cause bites from the other direction: because authority resolves through `created_by`, the key also dies when its creator leaves the team. A stack that has run for months breaks over an offboarding weeks earlier, with nothing in the tenant to explain it. One fix answers both — a team-owned service principal rather than a member's.

Closing it means giving sandbox endpoints a permission of their own and enforcing it for the `__qm_tenant__` key name, which is a change to the shared auth path every control-plane request takes.

**The blocking part is not the enforcement, it is knowing what to enforce.** The set of control-plane endpoints a tenant's QM actually calls has to be established by *watching a live tenant*, not inferred by reading either codebase. QM reaches sandboxes through several paths that are not obvious from the handlers here — create, connect, exec, pause and resume, the object store, the cron and webhook surfaces — and the two repositories move independently. An allowlist read off the code is wrong in one of two ways: too narrow, and every tenant breaks the first time a QM feature calls something nobody anticipated, surfacing as a tenant outage long after the change went in; too broad, and it is the appearance of a control without the control. Neither is better than this gap with a hard gate in front of it.

So it is deliberately **not** in this PR, and it is a blocker on production issuance rather than on merging. Anyone picking it up: do not start by guessing the endpoint list. The constraint and the reasoning are recorded at the top of `supabase/migrations/20260913000001_qm_issue_tenant_api_key.sql`.

## Read this first: the ownership model

Every per-tenant resource this provisioner creates is named after the tenant's **slug**, and the slug is chosen by whoever creates the tenant. So "a resource with the expected name exists" is not evidence that it is this tenant's, and treating it as interrupted-run state is a privilege-escalation path.

Concretely: the QM Terraform module names its own service accounts `qm-<word>-<suffix>`. A tenant created with the slug `provisioner-stg-usc1` derives the account `qm-provisioner-stg-usc1` — the identity the provisioner itself runs as. Adopting it would have deployed tenant-controlled code under that identity, and deleted it on teardown. The same shape reaches a Cloud Run service, a backend service, a bucket, a database and a secret.

**The invariant:** every per-tenant resource carries a marker keyed on the tenant's **id**, never its slug. No step will reconcile, grant on, or delete a resource that does not carry its own. An unmarked resource counts as *not ours*.

| Resource | Where the marker lives |
|---|---|
| Cloud Run service, bucket, Secret Manager secret | label `qm-tenant-id` |
| service account, serverless NEG, backend service | `description` (IAM and Compute take no labels) |
| database, Postgres role | shared-object comment (`shobj_description`) |

Two corollaries, both of which were live bugs before they were closed:

- **The skip path had no check.** The secrets step skips a secret whose reference is already recorded, so a recorded secret deleted and recreated under another owner was never re-verified where it is written. The grant in `cloud_run` is where that is now caught — granting on it would have mounted another workload's credential inside the tenant's container. `secrets.Store` gained `Owner` for exactly this.
- **Teardown skips rather than refuses.** Provision refuses a foreign resource; teardown must *skip* it. A sweep that failed on a name the tenant never created would have left that tenant permanently undeletable — there is nothing of ours there to remove, so there is nothing to fail over.

Slug validation also reserves the module's `qm-<word>` names, including as a first label. That is defence in depth, not the guarantee: enumerating the names to reserve is precisely the approach that fails, which is why the marker carries the weight.

One deliberate trade: an unmarked database or role is refused, not adopted. `CREATE DATABASE` cannot join a transaction, so a run that dies in the two statements between it and its `COMMENT` leaves an empty, connection-less database that an operator must drop. Refusing to reset a stranger's password and drop their data is worth that.

## Summary

Implements the GCP-calling half of the hosted-QM provisioner. Before this, a tenant could be created through the API and the run would execute, but every cloud step was a stub and no tenant was actually provisioned. Every step in the plan is now real, and the binary no longer needs `QM_PROVISIONER_STUB=1` to start.

The step interfaces were already declared, so the steps keep their shape: ordering, idempotence and rollback stay in `internal/qm/provisioner/steps`, and the mapping onto Google's APIs is a new `internal/qm/provisioner/gcp` package behind them. Stub mode is unchanged and still exercises the engine without a project.

Real now: `service_account`, `database`, `bucket`, `sandbox_key`, `cloud_run`, `load_balancer`, `health_check`, `smoke`. `secrets` and `admin_link` already were.

**A provisioned tenant can be signed into.** The reference tenant shipped with an auth broker that had no email transport: it answered its health check and 503'd the moment anybody hit `/idp/authorize`, and nobody noticed because the signed admin link bypasses the broker. So the Resend key is mounted from one platform-level secret every tenant shares (not a generated per-tenant one), `AUTH_EMBEDDED` / `AUTH_EMAIL_TRANSPORT` / `AUTH_EMAIL_FROM` are rendered per tenant, a missing transport stops the binary at startup, and the smoke step fails the provision on anything but the broker's own answer to a well-formed authorization request.

**Depends on #504.** Every value a tenant is provisioned with is the QM Terraform module's, and this branch removes the four the deploy workflow used to carry. Safe in either merge order — the workflow does nothing until `QM_API_SERVICE` is set, which only happens once the module has been applied — but #504 should land first. #504 also carries IAM this needs: the storage HMAC-key permissions, and the narrowed grant on the shared Resend secret.

## Other technical decisions

- **`QM_SECRETS_BACKEND=memory` now requires `QM_PROVISIONER_STUB=1`.** In-process mode alone used to allow it, and the real cloud steps would then build a service account, database, bucket and HMAC key around secrets that do not outlive the process. A restart made it worse rather than better: the tenant's recorded secret references survive, so the next run skips regenerating the values that are now gone. This is a trap that is easy to hit locally and easy to misdiagnose, so it is refused at config load with a message saying why.
- **Generated secrets widened to what the tenant image actually requires.** QM refuses to boot in production without its five signing secrets, and its broker refuses without `AUTH_SIGNING_JWK` (a P-256 private JWK), `AUTH_TOKEN_SECRET` and `AUTH_CLIENT_SECRET`. `DATABASE_URL`, the bucket's HMAC credentials and the sandbox key are derived by the steps that create those resources.
- **Databases over a direct connection, not the SQL Admin API.** That API can create a database and a user but cannot set an owner or take `CONNECT` from `PUBLIC` — and on a shared instance those two *are* the isolation. The create-locked/revoke/chown/open ordering is covered against a real Postgres in `internal/integration/qm_sql_test.go`, which also caught that a Postgres 16 `CREATEROLE` admin cannot hand a database to a role it created without granting itself membership first, and that a tolerated duplicate-role error aborts the surrounding transaction.
- **Sign-in opens to the admin's address only.** The brief asked for `AUTH_ALLOWED_EMAIL_DOMAIN` defaulted from the admin email; I deliberately did not. Nothing in an address distinguishes a company domain from a mailbox provider, a denylist of providers is incomplete by construction, and one miss opens a tenant holding the team's model key and sandbox credentials to everyone at that provider.
- **Slack sign-in is rejected at create.** The schema keeps the value; the provisioner renders one flow, and a tenant reported as Slack while serving magic links is worse than refusing it.
- **The model key is checked with its provider before the tenant row exists.** Slugs are globally unique across deleted tenants, so a rejected key would otherwise burn the name for good. Only a 401 counts — a restricted key, a rate limit or an unreachable provider all pass. The smoke step asks again before calling a tenant ready.
- **Tenant keys are issued through a definer function.** `qm_api` has no `INSERT` on `api_key` and should not get one. The function inserts the key and points the tenant row at it in one statement, so a live credential can never end up unreferenced, and it fixes the key's name and scopes itself rather than trusting its caller.
- **Tenant services accept unauthenticated requests.** The shared load balancer's serverless NEG carries no identity, so without it every public request including the health probe is a 403. Authentication is the tenant's own.
- **The startup gate narrowed rather than disappeared.** It no longer reports unimplemented steps — there are none — but it still refuses a binary whose plan could not run: a missing client where the plan executes, or shared infrastructure the deploy never configured. The API service in its default mode holds no cloud clients (it only queues the job) but validates the same configuration it hands that job.

## Testing

- `go test -short -race ./internal/qm/...` — steps covered with fakes for each client: the whole stack built, idempotent re-runs, rollback after a partial failure at seven different calls, rollback when `Run` never ran, a retried teardown, **refusing to adopt each resource kind when it carries another owner and skipping it on teardown**, a failure path recording a usable event, and an assertion that nothing secret-shaped reaches the event log. Teardown tests end on "no identity, database, bucket, HMAC credential, service, route or secret remains".
- `go test -tags integration -race ./internal/integration/ ./cmd/migrate-team/` against a real Postgres — the definer function (scoping, idempotence, fixed name and scopes, refusal for a deleted or creator-less tenant), the direct `INSERT` staying refused, and the provisioner's DDL against a role with exactly Cloud SQL's admin privileges, including that another tenant's role cannot connect, that an unmarked or foreign-marked object is refused, and that an interrupted run's damage is repaired.
- `go vet`, `gofmt`, `sqlc generate` clean.
- Local `codex exec review` run to a clean pass over the whole change.

Nothing was run against live GCP.

**CI position: this PR shows no checks because it targets a non-main branch, where CI does not run. That is not green.** The local runs above are the only signal. `/tmp/qm-live/m2-exit.ts` encodes what done means end to end and needs a deployed qm-api; this change is written to satisfy it, but it has not been run.


pavitrabhalla:
@codex review

pavitrabhalla:
Pushed: the four tenant-runtime variables are now the Terraform module's (#504), so this branch no longer sets them from the deploy workflow. Also closes an ownership hole the review surfaced — every per-tenant resource now carries a tenant marker and no step will reconcile, grant on or delete one that does not carry its own.

@codex review

pavitrabhalla:
Pushed. Addresses the review:

- **Tenant secrets now carry an ownership marker**, the last resource class without one. `secrets.Store.Put` and `Delete` take the tenant's id, write it as a `qm-tenant-id` label on create, and refuse a secret that exists under a different owner or none — so a slug chosen to collide can no longer add a version to somebody else's secret, be granted read on it, or delete it on teardown. qm-api's model-key write goes through the same check, before the tenant row exists.
- **Dropped the point-in-time justification** from the deploy workflow comment; the actionable migration instruction stays.
- The whitespace lifecycle finding was already fixed in `6fe0329` — it also now rejects `{}`, `null`, a misspelled `rules`, unknown fields and trailing data, through one parser both the readiness check and the bucket client use.

Also closed a config hole the local loop found: `QM_SECRETS_BACKEND=memory` now requires `QM_PROVISIONER_STUB=1`. In-process mode alone let the real cloud steps build a service account, database, bucket and HMAC key around secrets that do not outlive the process — and a restart made it worse, since the recorded references survive and the next run skips regenerating the values that are gone.

@codex review

pavitrabhalla:
Pushed both findings:

- **Secret ownership is checked before the grant, not only before the write.** The secrets step skips a secret whose reference is already recorded, so a recorded secret deleted and recreated under another owner was never re-checked on the write path — and the grant loop would then mount another workload's credential in the tenant's container. `secrets.Store` gained `Owner`, and `cloud_run` verifies every per-tenant secret it is about to grant on, exempting the intentionally shared Resend secret.
- **Teardown skips a secret it does not own instead of failing on it.** With `Delete` now refusing a foreign secret, an unconditional sweep would have left a tenant undeletable forever over a name it never created. One `deleteTenantSecret` helper carries that rule for all four rollbacks.

@codex review

pavitrabhalla:
Pushed the last round of review findings:

- **A tenant secret is verified before it is consumed, not only before it is written.** The secrets step skips a secret whose reference is recorded, so a recorded `DATABASE_PASSWORD` replaced by something else was read and applied — setting the tenant's role to a password that workload knows — before any later check ran. One `checkSecretOwner` rule now covers the read, grant and write paths.
- **Teardown never stops on a resource that is not this tenant's.** The runner stops a deprovision at its first failing step, so refusing over a name the tenant never created would strand its own database, bucket, identity, secrets and shared-secret binding with no way to retry past it. One `steps.ErrNotOwned` sentinel across the layer; provision fails on it, teardown treats it as absent and carries on — including deleting this tenant's own NEG and backend service when only the *route* is foreign, which an earlier version of the skip leaked.
- **Route ownership is judged inside the compare-and-swap**, against the map each attempt just read, rather than once before the loop.
- **The rolling-deploy window is handled.** qm-api and the job deploy separately, job first, so a tenant created by the previous API revision has an unlabelled model key; that is adopted and stamped rather than refused, which would have consumed the slug and then failed the run partway through. A secret labelled for *another* tenant is still refused.

**One finding is deliberately not fixed** — see the "Known blocker" section now at the top of the description. The tenant key carries its creator's full team role. Closing it needs sandbox endpoints to get a permission of their own and enforcement for the `__qm_tenant__` key name, which is a change to the shared auth path every control-plane request takes and needs the exact endpoint set a tenant's QM calls — not establishable without running one against a live API, and an enforcement that is too narrow breaks every tenant. It is a blocker on production issuance, recorded at the top of the migration, not on merging this.

@codex review

pavitrabhalla:
Pushed one commit addressing the two open findings against 027db961:

- **A legacy-unlabelled secret is claimed the moment it is accepted, not just read past.** The model key is written once by qm-api and after that the provisioner only ever reads or grants it, so the tolerance for an unlabelled secret from the job-first deploy window was, in practice, permanent rather than one-time: `checkSecretOwner` never stamped it. `secrets.Store` gained `Claim`, called on the unlabelled path; `GCP.adopt` conditions its patch on the fetched etag and reads the secret back afterward either way, since a conflict there only proves something changed the secret, not that a different owner won it, and this API surface may not enforce that precondition at all.
- **A bucket name can collide with one in a different GCP project**, which Cloud Storage reports as a 403 — the same response it gives when this project's own access to its own bucket has regressed, since it deliberately does not distinguish the two. `Buckets.Get` now returns the shared `ErrNotOwned` sentinel for it, so provisioning still fails on it (unchanged) but teardown treats it as absent instead of stopping there and stranding the tenant's database, service account and secrets on every retry.

Local Codex review flagged the bucket classification again after this, suggesting it only count as a collision when the tenant never recorded owning that bucket name. I didn't take that: the scenario the original finding described is exactly a *recorded* bucket that was deleted and then reclaimed by another project, so narrowing it that way would refuse that same case again. GCS gives nothing else in the response to tell "reclaimed elsewhere" apart from "this project's own access regressed," an access regression broad enough to hit this path would show up the same way across every tenant's bucket operations rather than as one quiet case, and the skip lands on the tenant's own event log rather than silently. Documented the tradeoff in `Buckets.Get`.

Reproduced both with new tests (`TestProvisionClaimsAnUnlabelledModelKeyItAccepts`, `TestProvisionRefusesAModelKeyClaimThatLosesItsRace`, `TestTeardownContinuesPastACrossProjectBucketCollision`) and confirmed they fail against the prior code. `go test -race ./internal/qm/...` is green.

@codex review

PR #510, opened 2026-09-14
api: pause operation columns and a background reconciler

meAmitPatil:
Adds the pause operation columns (id, start time, lease, trigger, actor) and a background reconciler that finishes any sandbox left `pausing` with an operation: it re-asks the host every 30s, fails the row only on NotFound or a failed precondition, and flags anything pending past 30 minutes. Finalize and mark-failed accept an optional operation fence; callers without one keep today's behaviour.

Nothing creates pause operations yet, so the reconciler idles. Shipping it first means a later rollback from the operation-based pause still has a reconciler to finish in-flight pauses.


meAmitPatil:
@codex review

awille-ai (approved):
## Verdict
APPROVE

I reviewed this as the new rollout foundation split out of #493. I do not see a remaining Critical or High defect in the current head.

The split materially improves rollout safety: this PR ships the schema, fenced finalize/fail primitives, and the background reconciler **before** any production path creates a durable pause operation. Existing pause code leaves every new operation column NULL, so the reconciler is dormant until the stacked #493 lands. If #493 is later rolled back, the rollback target is this release, which still knows how to finish any operation-based pauses already in flight.

## Review scope
- Migration compatibility with the current control plane and existing sandbox rows.
- `ListPendingPauses` / `ClaimPendingPause` lease ownership, fencing, retry/backoff, and multi-replica claiming.
- Terminal NotFound/FailedPrecondition handling vs undecided host errors.
- Finalize-reply-loss recovery (`pauseLanded`) and completion side effects.
- Original trigger and actor preservation through reconciliation.
- Cancellation/shutdown behavior and detached lease release/failure writes.
- Batch/refill behavior and bounded concurrency.
- Forward/backward compatibility with the stacked #493 rollout.

## #493 finding verification
One of the two High findings from my previous #493 review is fixed by this split:

- **Rollback/version skew — addressed.** `StartPauseReconciler` now lands in this prerequisite release. #493 is explicitly based on this branch, so rolling #493 back leaves a reconciler running that can finish durable pauses created by #493.

The other High is **not a #510 defect and remains in current #493**: the automatic timeout/billing pre-dispatch path still calls `failPause` if `RevertPauseToActive` itself returns a DB error after host resolution failed. In that state no pause RPC was dispatched, so the VM is known to still be running; marking the row failed can stop billing/state tracking while leaving the workload alive. That needs to be fixed in #493 before the stack lands.

## Reviewer execution
| --- | --- | --- |
| Infrastructure/migration safety | sequential fallback | PASS |

PR #511, opened 2026-09-14
fix: allow absent staging standby for manual production deploys

awille-ai:
Manual production standby deployments currently fail in staging when no staging standby is provisioned, preventing deployment to the selected production standby. Add a staging-only presence check that skips the deployment step successfully when the standby label has no matches in staging's region.

Manual serving deployments still require staging to succeed first. Existing staging standby hosts proceed through normal deployment validation, and discovery errors remain fatal. Production zero-match failures and push behavior are unchanged; `deploy-vmd.py` is unchanged.

Validation:
- All 17 selector and staging workflow tests pass, including optional absence, region scoping, discovery failures, and deployment failure propagation.
- `git diff --check` passes.
- The broader deployment suite ran 84 tests with 11 failures and 6 errors locally; unchanged main reproduces the same failures and errors across its 81 tests, including macOS/Linux command incompatibilities.


awille-ai:
@codex review

meAmitPatil (approved):

PR #515, opened 2026-09-14
Add explicit standby targeting to proxy deployments

awille-ai:
## Summary

Proxy deployments currently discover only serving hosts, leaving standby hosts untouched. Add manual `target` (`serving` by default or `standby`) and `production_cell` (`usw2` or `use4`) inputs. Standby deployments reuse VMD's shared selector: west targets `superserve-vmd-usw2-2`; east uses `VMD_STANDBY_HOST_USE4`.

Require exactly one running standby with the expected identity in the configured region. Missing production standbys, unconfigured identities, discovery errors, and empty regions fail closed. An absent staging standby is a successful skip only for manual production standby deployments; a present but stopped or mismatched standby fails.

Serving/default/push rollout guards and regional configuration stay intact. Standby hosts use the same binary, config, unit, and Superserve-managed `/etc/superserve/peer/` credential-loading implementation. No labels, admission state, or TLS material are changed.

## Validation

- 12 focused proxy tests pass, including standby selection, region/identity/status failures, staging skip, default/push guards, and identical generated installation/peer configuration.
- 12 shared VMD selector tests pass.
- Workflow YAML parses; `git diff --check` passes.
- Full deploy-script suite was attempted locally: the same 11 failures and 6 errors reproduce on unchanged main due to GNU/Linux shell assumptions on macOS. Linux CI will run the complete suite.

## Manual rollout

Dispatch **Deploy Proxy** with `environment=production`, `target=standby`, and `production_cell=usw2`. The production step targets only `superserve-vmd-usw2-2`, retaining its standby label. For east, configure the existing `VMD_STANDBY_HOST_USE4` variable and select `use4`.


awille-ai:
@codex review

meAmitPatil (approved):

PR #516, opened 2026-09-15
fix(observability): preserve valid backup identity filters

awille-ai:
## Summary

Fix the Terraform rollout failure introduced by #514: Cloud Monitoring rejects `result/priority AND (host_id OR collector_host_id)` because metric-label restrictions cannot mix AND and OR.

For upload failures and oldest pending age, generate two threshold conditions combined with `OR`. Each retains the exact result/priority restriction, one exact host identity, and the original threshold, alignment, trigger, and duration. Keep the other backup and launch-path filters as single host-label unions, preserving percentile aggregation and host grouping. Legacy callers without `collector_host_id` still get one condition.

Audited all five backup and three launch-path policy variants and their staging, production west, and production east callers. The shared fix covers all six affected policy instances. No host identity, heartbeat, collector deployment, or admission changes.

## Validation

- `terraform validate`: observability module and staging environment passed.
- `terraform test`: all three observability runs passed. The new tests fail against unchanged main with the original mixed-label filter.
- Terraform formatting and whitespace checks passed.
- Read-only Cloud Monitoring `timeSeries.list` accepted all 10 threshold filters from the staging plan, plus the optional backup-disabled and launch-path union shapes. A control query using the original invalid filter returned the exact `AND and OR cannot be mixed` error.
- Full staging plan: **0 add, 7 update, 0 destroy**, with no replacements. Two backup alert policies change conditions; five dashboard updates are identical in a separate plan from unchanged main. Comparing every planned resource value between main and this branch shows this PR changes only the two expected backup policies. The full plan therefore still includes pre-existing dashboard differences; this PR does not modify them.

No Terraform apply or host rollout was performed.


awille-ai:
@codex review

meAmitPatil:
P2 — The two identity conditions overlap. Modern backup metrics contain both host_id and collector_host_id, so the same upload failure or
  backlog violation satisfies both generated conditions in /private/tmp/sandbox-pr516-review/infra/modules/observability/backup-alerts.tf:159.
  Cloud Monitoring policies with multiple conditions can produce separate notifications, creating duplicate pages. Use an explicit legacy/
  migrated identity mode instead of keeping both conditions active permanently. Cloud Monitoring documentation

meAmitPatil (approved):

meAmitPatil:
The split filters overlap for modern metrics: the collector preserves `host_id` and also adds `collector_host_id`, so one failure can satisfy both OR conditions and potentially generate duplicate alerts/pages. Please make identity selection an explicit legacy/migrated mode (or otherwise prevent both conditions from matching the same series).

PR #517, opened 2026-09-15
backup: materialize a shared base once per restore and clone it into each sandbox

meAmitPatil:
Bulk restores wrote a full copy of the template base into every sandbox directory, so a fleet sharing one base cost one base of disk per sandbox. The base cache now unpacks and verifies each base once and reflink-clones it into place; filesystems without reflink get a plain copy. Bytes and per-file verification are unchanged.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

internal/backup/hostrestore.go
```diff
@@ -543,3 +548,145 @@ func (c *CachingBaseReader) NewReader(ctx context.Context, object string) (io.Re
...
+		if _, err := tmp.Seek(0, io.SeekStart); err != nil {
+			tmp.Close()
+			return nil, err
+		}
+		extents, apparent, err := Extents(tmp)
+		if err != nil {
+			tmp.Close()
+			return nil, err
+		}
+		sum, err := hashApparent(ctx, tmp, extents, apparent)
```
awille-ai: **High — extra full-base hash remains on the synchronous restore path.**

The first restore for each distinct shared base runs `hashApparent` across the entire unpacked base inside the singleflight before any waiter can clone it. `RestoreGeneration` then runs the existing `verifyFile` full-content hash on every destination, including the first one. For multi-GiB bases this adds an extra complete read/hash per base and makes all concurrent restores for that base wait behind it.

This conflicts with the repo's protected lifecycle-path guidance (`AGENTS.md`: avoid new hashing/blocking work in restore/startup paths unless unavoidable and measured). Please either preserve the cache-poisoning guarantee without this additional synchronous full pass (e.g. promote an already-verified copy/master after normal verification), or include a concrete production-size latency measurement/justification showing this cost is acceptable.

awille-ai (commented):
Consolidated review (manual pass + pr-review skill): one High finding remains.

The earlier path traversal, cache-poisoning, destination-filesystem fallback, and repeated clone-probe issues have been addressed in the current head. CI is also green.

The remaining concern is restore-path performance: first materialization of a shared base performs a complete apparent-content hash before publishing the master, while RestoreGeneration still performs the normal full verifyFile hash of each cloned destination afterwards. This adds one additional full read/hash of every distinct base on the synchronous restore path and serializes all concurrent restores waiting on the same base behind it. The repository's lifecycle guidance explicitly calls out hashing on restore/startup paths as something that must be kept off the hot path when possible, or otherwise justified and measured.

I would either avoid the extra full hash while preserving the cache-poisoning guarantee (for example by promoting an already-verified restored copy/master at a point where verification has completed), or provide a concrete latency measurement/justification showing that this additional pass is acceptable for production-sized bases.

Review scope: shared-base cache/materialization correctness, hostile inputs and crash/retry behavior, reflink/fallback behavior, concurrency, tests, resource usage, observability, and sandbox restore-path latency.

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

internal/backup/restore.go
```diff
@@ -187,26 +187,39 @@ func RestoreGeneration(ctx context.Context, r BlobReader, sandboxID, generation,
...
+		err := verifyFile(ctx, root, mf)
+		if publish := publishers[mf.Name]; publish != nil {
+			// This verification is the one that blesses the copy the
+			// reader materialized from; a failure here must also stop it
+			// being reused.
+			if perr := publish(err == nil); perr != nil {
+				report("shared base %s not cached: %v", mf.SHA256, perr)
+			}
+		}
+		if err != nil {
```
awille-ai, line 234: **High — early verification failure strands unverified shared-base candidates.**

All files are materialized before this verification loop, so `publishers` can already contain callbacks for shared bases later in manifest order. If this `err != nil` path returns on an earlier file, those later callbacks are never invoked with `false`. Their `.candidate-<sha>-*` files remain in the persistent cache, and `MaterializeBase` later `Glob`s and reuses existing candidates even though nobody ever verified those bytes.

Please invalidate every still-pending candidate on any restore failure/early verification return (without discarding masters already promoted by successful verification), and add a regression test with a failing earlier file followed by a materialized shared base.

awille-ai (commented):
Second consolidated review (fresh manual pass + full pr-review rubric): one High finding remains.

The prior restore-path performance concern is resolved: the extra pre-publication full-base hash is gone, and candidate promotion now reuses the normal destination verification. The earlier path traversal, reflink fallback/probing, and direct cache-poisoning issues also remain addressed. CI is green.

However, the candidate lifecycle is incomplete on an early verification failure. RestoreGeneration materializes every file first and stores publish callbacks, then verifies files in manifest order. If file N fails verification, the function returns immediately; callbacks for materialized shared bases appearing after N are never called with false. Their unverified `.candidate-<sha>-*` files remain in the persistent cache. MaterializeBase explicitly glob-reuses such candidates on subsequent restores, so bytes that were never verified can become the source for a later restore. A malformed/corrupt generation can therefore contaminate retries/other generations until that candidate eventually gets selected, fails, and is discarded.

The failure path should invalidate every outstanding unpublished candidate before returning (while preserving already-published verified masters), and there should be a regression test where an earlier non-base file fails verification before a later shared base reaches its callback.

Review scope: correctness, adversarial/corrupt manifests, concurrency/singleflight, crash/retry/cache lifecycle, security/resilience, tests, performance/resource behavior, observability, and sandbox restore-path latency.

awille-ai (approved):
Re-reviewed current head at the same bar (manual pass + full pr-review checklist). The previous High finding is fixed: outstanding candidate callbacks are now invalidated on every early restore failure, and per-file callbacks are removed before delivery so they cannot be invoked twice.

I re-checked the candidate/master lifecycle, concurrent restores, failed verification, direct-unpack fallback, reflink probing/cache behavior, path validation, crash/retry semantics, cleanup, and restore-path latency. I do not see any remaining Critical or High findings in the current head.

The earlier extra-hash latency regression remains fixed, the candidate is only promoted after verification of a clone derived from that exact candidate, and failure paths no longer leave this restore's unverified candidate eligible for later reuse.

CI for this exact head is still in progress at the time of review, so this approval is code-review approval rather than a statement that the checks have completed.

PR #519, opened 2026-09-15
backup-restore: migrate paused sandboxes onto this host from a host restore

meAmitPatil:
Adds a `migrate` subcommand to the operator tool: for paused sandboxes pinned to another host whose filesystems a host restore materialized here, it re-pins the row, cold-boots the restored disk through ReviveVM, and lets the timeout reaper take it back to paused through the normal pause path, keeping the number in flight inside the reconciler's grace. Rows that fail to boot are reverted to the source host and recorded so reruns skip them; rows that already had a timeout keep it.
The reaper's per-tick bounds (50 / 10) become overridable with REAPER_BATCH_SIZE and REAPER_PARALLELISM so a planned wave can be paused at a useful rate; unset, nothing changes.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

cmd/backup-restore/migrate.go
```diff
@@ -0,0 +1,778 @@
...
+						mu.Lock()
+						recordFailure(id, err.Error())
+						if jerr := journalDone(journal, id); jerr != nil {
+							fmt.Printf("WARN %s: journal not updated: %v\n", id, jerr)
+						}
+						mu.Unlock()
+						return
+					}
+					// Boot, then activate with the fresh address so nothing
+					// routes to the old slot. A failed boot hands the row back
```
awille-ai: **High — the migration claim still leaves the row resumable during `ReviveVM`.**

This claim changes only `host_id`; the sandbox remains `status='paused'`. The ordinary resume path's `BeginResume` claims any paused row, so after this statement commits an owner can resume the sandbox on `toHost` while this worker is still about to/in the middle of `ReviveVM` for the same sandbox ID.

Moving the claim before boot fixes the old cross-host dual-run race, but it does not establish exclusive control-plane ownership of the lifecycle transition. The VMD lifecycle lock can serialize host operations, yet the normal resume may still run before or after the revive and act on state created by the migration, while the migration later tries to activate based only on `status='paused'`.

Please make the pre-boot claim non-resumable (dedicated transitional state/operation ownership, or an equivalent fence that `BeginResume` excludes) before calling `ReviveVM`, then activate only after the revive succeeds.

awille-ai (commented):
Consolidated review (manual pass + pr-review skill): one High finding remains.

Most of the earlier review issues are addressed in the current head: activation happens after ReviveVM, egress policy is supplied, standalone restores are supported, timeout state is journaled/recovered, snapshot+timeout fences were added, fresh IP is persisted, and the reaper overrides are bounded. CI and Terraform checks are green.

The remaining issue is the pre-boot claim state. The tool moves `host_id` to the destination but deliberately leaves the sandbox `paused` while ReviveVM runs. Ordinary owner resume claims any `paused` row via BeginResume and does not distinguish a migration-owned paused claim. That means, during the operator boot window, a user resume can claim the same row on the destination and enter the normal resume path while ReviveVM is concurrently cold-booting the restored disk for the same sandbox ID. The host-side lifecycle lock may serialize the two VMD operations, but whichever runs second can still act on state created by the first, and the control-plane state machine has no migration-specific ownership/fence to make one side fail before touching the VM.

I would make the claim non-resumable before invoking ReviveVM (e.g. a dedicated migration/transitional status, or an equivalent ownership token that BeginResume excludes), then transition to active only after ReviveVM succeeds. The existing snapshot/timeout predicates are good stale-read fences but they do not prevent owner resume after the claim because the row remains paused.

Review scope: migration state machine, owner resume/pause races, ReviveVM interaction, timeout/reaper behavior, billing side effects, journal crash recovery, restored-generation freshness, egress/network policy, concurrency/resource bounds, rollout compatibility, and tests.

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

supabase/migrations/20260915000003_sandbox_host_live_index.sql
```diff
@@ -0,0 +1,16 @@
...
+-- On a populated database this index must be pre-built CONCURRENTLY, by
+-- hand: the migration runner wraps every file in a transaction, where
+-- CONCURRENTLY cannot run, and a plain build blocks sandbox writes for the
+-- duration of the table scan. Run before merging:
+--
+--   CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_sandbox_host_live
+--     ON sandbox(host_id)
+--     WHERE status IN ('active', 'starting', 'migrating') AND destroyed_at IS NULL;
+CREATE INDEX IF NOT EXISTS idx_sandbox_host_live
+    ON sandbox(host_id)
```
awille-ai: **High — this automatic migration can block sandbox writes unless an unenforced manual pre-step happened everywhere.**

The comment correctly notes that a populated `sandbox` table needs this index built with `CREATE INDEX CONCURRENTLY`, but the actual migration is a plain `CREATE INDEX IF NOT EXISTS`. Our CD path runs `supabase db push` automatically in staging and then production DBs and does not verify that the concurrent pre-build already exists. If even one production DB/cell misses the manual command, this migration takes the normal table-locking index-build path during rollout.

Please make the prerequisite machine-enforced/checkable in the rollout, or pre-create and verify the exact index concurrently in every target DB before merge. A comment saying “run before merging” is too easy to miss for a write-blocking production migration.

awille-ai (commented):
Consolidated re-review (manual pass + pr-review rubric): one High rollout finding remains.

The previous lifecycle/concurrency finding is fixed: `migrating` now provides a non-resumable pre-boot claim, the timeout reaper understands that state, failed pause compensation preserves it rather than exposing/billing the recovery boot, timeout mutations are rejected while migration owns the row, and the new state is carried through quota, scheduler load, drain accounting, telemetry, team migration blocking, and the public API schema.

The remaining concern is the new `idx_sandbox_host_live` migration. The migration itself explicitly says the index must be pre-built with `CREATE INDEX CONCURRENTLY` before merge because the normal migration runner cannot use `CONCURRENTLY`; however, the committed migration still executes a plain `CREATE INDEX IF NOT EXISTS`. The normal CD workflow just runs `supabase db push` in staging and then the production databases and has no prerequisite check that the concurrent index already exists. If any populated production DB/cell misses the manual pre-step, merge can take a write-blocking index build on `sandbox` during automatic migration rollout.

Before merging, either make the prerequisite machine-enforced/verifiable in the rollout path, or confirm/pre-create the exact index concurrently in every database the migration workflow will touch. I would not rely on the SQL comment alone for this production-safety requirement.

Review scope: migration state-machine correctness and races; crash/retry/idempotency; reaper/pause interactions; billing/quota semantics; owner-visible lifecycle/API behavior; host scheduling/capacity/drain accounting; security/network/preview policy restoration; persistence/journaling; migration and rollout safety; performance and observability.

CI/Terraform checks for this exact head were still running at review time.

meAmitPatil:
@codex review

awille-ai (approved):
Re-review of the current head at the same bar: APPROVE.

The previous High rollout finding is fixed. The `idx_sandbox_host_live` migration has been removed and `ListActiveHostsByLoad` keeps its existing indexed predicate, so this PR no longer introduces a potentially write-blocking index build on `sandbox` or a new synchronous-create-path query shape.

I re-checked the current migration state machine and the fixes accumulated across the review cycle: non-resumable `migrating` ownership before boot, snapshot-generation fencing, timeout journaling/recovery, ambiguous DB/RPC outcomes, forced cleanup before releasing claims, migration-aware reaper compensation, billing/quota behavior, timeout PATCH exclusion, preview/egress restoration, proxy exposure, VMD/host binding, concurrent-run locking, team-migration/drain behavior, public API compatibility, and capacity telemetry. I do not see a remaining Critical or High correctness/security/resilience issue.

One deliberate operational tradeoff remains: `migrating` rows are not included in placement load ranking. That means an active destination can temporarily look lighter to the scheduler during a recovery wave. Given this is an operator-controlled recovery path, the concurrency/in-flight bound is explicit, the condition remains visible in host-capacity telemetry and drain accounting, and preserving the existing placement predicate avoids putting a broader DB scan/index rollout onto the create hot path, I do not consider that a blocking defect for this PR.

Review scope: hostile/concurrency failure modes; security/resilience; test coverage; general correctness; DB/data-state transitions; crash/retry/idempotency; performance and sandbox lifecycle latency; observability; compatibility/rollout; schema/migration safety.

CI for this exact head is still running. Terraform Checks was cancelled, so this approval is for the reviewed code and migration logic rather than a claim that all checks have completed successfully.

PR #526, opened 2026-09-16
api: reclaim a deleted sandbox's host state after the response, recorded until it completes

meAmitPatil:
DELETE /sandboxes marks the row deleted and then reclaims the VM and its snapshots on the host before answering; when the host is unreachable the caller waits through the RPC timeout and its retry. The same statement that deletes the row now records the reclaim in sandbox_teardown, owned by the request for one inline attempt bounded as a whole (reclaim and settle) by TEARDOWN_INLINE_BUDGET, 5 s by default, so a reachable host is fully reclaimed before the response and an unreachable one costs only the budget. A sweeper on every replica claims what nobody owns from the database, taking a per-host lease in the same statement so one attempt per host holds fleet-wide, with attempt-fenced settles and releases, per-host backoff, permanent-failure marking for unregistered hosts, and backlog metrics.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

awille-ai (approved):
## Consolidated review — APPROVE

I reviewed the current head twice: first manually without the review skill, then independently with the PR-review workflow. Both passes converge on **no remaining Critical or High findings**.

### Review scope
DELETE/auto-delete teardown durability, migration/rollback behavior, teardown leases and attempt fencing, per-host concurrency, retry/backoff behavior, team-purge interaction, snapshot/artifact cleanup, request latency/resource bounds, telemetry, and the migration itself.

I specifically rechecked the earlier review findings against the current code rather than relying on resolved threads:
- pending teardowns now block team purge, including a second check under the purge locks;
- auto-delete teardowns are born at attempt 0, so the sweeper's first real attempt is attempt 1;
- `ClaimNextTeardown` selects at most one candidate per host before row locking, and integration coverage exercises both same-host serialization and spreading to another host while one claim is in flight;
- the sweeper starts the settlement timeout after reclaiming, so a slow-but-successful reclaim can still clear its durable row;
- snapshot-row deletion failures propagate and leave the teardown retryable;
- backlog `retrying` is derived from recorded failure state rather than requiring a second failed attempt.

The resulting design looks sound: the soft-delete and teardown obligation are atomic, the user request gets one bounded inline attempt, auto-delete never blocks on a host, all later work is fenced/idempotent, one attempt per host is enforced fleet-wide, and the sweeper has bounded workers/backoff plus useful backlog/attempt telemetry. The repository's migration/deploy gating also preserves the required schema-before-code ordering.

| --- | --- | --- |
| Infrastructure/migration safety | sequential fallback | PASS |

PR #530, opened 2026-09-17
fix(infra): exclude draining staging host from routine deploys

awille-ai:
Staging Host 1 is draining but still matches the `component=vmd` selector used by routine VMD and proxy deployments. Those deployments would continue targeting its legacy installation.

Change its component label to `vmd-staging-draining`, matching the operator-applied exclusion. Keep `sandbox_role=vmd` so role-based monitoring still includes it while existing workloads drain. Host 2 remains the routine deployment target.

Validation: `terraform fmt -check`, `git diff --check`, and live readback confirmed Host 1 is RUNNING with the intended component and role labels. No Terraform apply was run for this PR. Merging triggers the existing Terraform CD workflow.


awille-ai:
@codex review

pavitrabhalla (approved):

PR #535, opened 2026-09-17
vmd: activation of frozen templates behind two switches, both off

meAmitPatil:
The switches that turn the guest-clock work on, and what turning them on needs. VMD_TEMPLATE_FREEZE_WORKLOAD has the daemon pass --freeze-workload to the builds it forks, so a template's workload is frozen for its snapshot once the guest proves it corrects its clock; VMD_GUEST_CLOCK_FREEZE, already present, makes restores and pauses freeze and wake. Both default off, and nothing sets them. The deploy script reconciles both into the host's vmd.env from the workflow, the same way as the backup backfill switch, so unsetting one in the workflow turns it off on the next deploy.

Also here: queued wakes are served during the startup scan rather than after it, so a recovering sandbox no longer waits behind every record on the host; the template scan replaces the per-directory tree watch, bounded by the number of templates; a raise-wake-floor operator command raises the rollback floor on a host ahead of the first frozen image, so no vmd without the wake protocol can start on it afterwards; and a frozen image's restore records its latency phases under a frozen mode, so the two kinds of restore can be compared on the dashboard.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

pavitrabhalla (approved):
Reviewed by QM on behalf of pavitrabhalla at ebfdfe6a84003118ee15237f6628994ac2b09b31. No blocking findings.

Reviewed both opt-in switches and deploy-time removal, builder flag propagation and freeze/manifest publication ordering, the durable rollback-floor command and guard, concurrent startup wake recovery and lifecycle-lock handoff, periodic template scanning, and success/failure restore-phase labels. The earlier setup-failure telemetry finding is addressed in this revision.

Validation:
- All 12 deploy-vmd ordering tests passed.
- Full `go test ./internal/vm ./cmd/template-builder ./cmd/vmd -count=1` passed after dropping this review environment's elevated Linux capabilities; the initial permission-denial fixture failures reproduced on the base commit and disappeared without those capabilities.
- Race-enabled targeted wake/frozen/freeze/template/dormant/wall-clock tests passed across those three packages.
- The new startup-scan wake test passed 500 repetitions.
- All 11 reported GitHub checks are successful; diff whitespace check passed.

No live Firecracker/fleet rollout was exercised here. Both activation switches remain off by default; enabling them still requires the documented wake-capable deployment and rollback-floor preparation. No merge performed.

PR #539, opened 2026-09-18
vmd: a same-id restore after a cleaned-up failure starts over instead of restarting in place

meAmitPatil:
When a restore fails after boot and its cleanup removes the rundir, the instance stays tracked in the error state. A same-id retry then took the in-place path, which drops the delta dir and requires the overlay that cleanup deleted, so the retry always failed with a misleading overlay-missing error. A tracked instance with no rundir now starts over as a fresh restore.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

pavitrabhalla (approved):
Reviewed e9542e810b0140d4191f01954c625cacc711253b. No blocking findings.

The separation of prior-life evidence (inPlace) from resource reuse (inPlace && priorRunDir) fixes the cleaned-up same-ID retry without dropping the existing supervisor stop or launch safety gates. I checked overlay/delta planning, File/UFFD selection, network-slot rotation, and the failure cleanup call sites. The earlier review concerns about prior-life evidence and fresh-slot cleanup are addressed in this revision.

Validation:
- Targeted restore/planning/failed-restore tests pass locally with the race detector.
- The new FailedAttemptWithoutRunDir regression test fails against the base commit with the old overlay-missing error and passes on this head.
- All 11 reported GitHub checks pass, including Unit Tests, Integration Tests, Build, and Vet.
- The full local VM package run had three permission-sensitive failures (rollback unreadable-state guard, unremovable frozen-manifest pause, and unreadable-manifest durability); I reproduced all three on the unchanged base commit as well. This is not a claim that the full local suite passed.

Non-blocking follow-up: add a hook-based same-ID/no-rundir retry test that reaches launch and a tap-busy failure, asserting hadPriorLife=true/freshUnit=false and slot teardown/rotation. The new regression test currently stops at disk preparation, so it does not directly pin those important downstream uses of reuse.

Review performed by QM on Pavitra's behalf; no merge action taken.

PR #540, opened 2026-09-18
vmd: resume a paused sandbox from its bucket backup when the pause artifacts are gone from the host

meAmitPatil:
A resume whose snapshot, memory image or disk is no longer on the host used to fail outright. With the backup bucket configured and the opt-in switch set, the host now reports that the pause artifacts are gone, the control plane looks up the backup generation recorded as covering exactly that pause and retries naming it, and vmd restores that generation and revives the sandbox from its disk under its own record, so the caller sees an ordinary resume. Pause backups carry the disk and not the memory, so the sandbox boots cold with its files intact.

The ordinary resume path is untouched: the generation lookup runs only after the host has refused, over a covering index that must be pre-built concurrently before the migration lands. The host restores the recorded generation or fails closed, never a neighbouring pause. The fetch runs on its own budget and is shared by every attempt for the same generation, so a caller whose deadline expires gets a retryable error and its retry joins the download; a completed restore in staging is reused, a retry after the boot adopts the live VM, a failed boot and an unclaimed fetch drop their staging, and a restart sweeps whatever a previous process left. A template base the host still holds is used in place instead of downloaded; one materialized elsewhere is kept alive for the revived VM and swept once no VM reads it. Fetches share one base cache that is pruned to a size cap, spools included, and are bounded in concurrency and bandwidth; the identities that write backups gain read access to their bucket. Off by default.


internal/vm/resume_backup.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	if err := os.MkdirAll(m.backupRestoreRoot, 0o700); err != nil {
+		return nil, status.Errorf(codes.Internal, "restore root: %v", err)
+	}
+	defer os.RemoveAll(dest)
+
+	log := m.log.With().Str("vm_id", vmID).Logger()
+	start := time.Now()
+	reader := &backup.CachingBaseReader{Inner: m.backupReader, Dir: filepath.Join(m.backupRestoreRoot, ".base-cache")}
+	r, err := backup.FetchNewest(ctx, reader, m.backupLister, vmID, dest, func(format string, args ...any) {
```
pavitrabhalla: [P1] Require a backup matching the pause before replacing the sandbox

`FetchNewest` chooses the last completed manifest by object creation time, not the generation belonging to this pause. For example, pause A is uploaded, the guest resumes and writes new files, pause B is still uploading, then only B's memory file disappears. This branch downloads A and `reviveVMLocked` destroys the existing VM/run directory, including B's still-good disk, before booting A. The RPC can return success after silently discarding the newer files. The migration command already has a `restored.current(recorded)` digest check for precisely this distinction. Require evidence that the restored generation matches the current pause before teardown; preserve a usable current local disk, and fail closed (or require explicit rollback consent) when only an older backup is available. Add a two-pause test with the newer upload incomplete and only memory missing.

internal/vm/resume_backup.go
```diff
@@ -0,0 +1,95 @@
...
+	inst.mu.RUnlock()
+
+	dest := filepath.Join(m.backupRestoreRoot, vmID)
+	if err := os.RemoveAll(dest); err != nil {
+		return nil, status.Errorf(codes.Internal, "clear restore dir: %v", err)
+	}
+	if err := os.MkdirAll(m.backupRestoreRoot, 0o700); err != nil {
+		return nil, status.Errorf(codes.Internal, "restore root: %v", err)
+	}
+	defer os.RemoveAll(dest)
```
pavitrabhalla: [P1] Keep the restored overlay base alive after returning

`RestoredDisk` returns `r.Base` inside `dest`. Revival resolves that path and `coldBootFromRootfs` copies only the writable disk; it retains this exact base path in `Config.BasePath` and uses it in Firecracker setup. `CachingBaseReader` materializes a regular destination file (possibly a reflink), not a symlink to a durable cache pathname. This defer therefore deletes the base referenced by the successful running VM. The open file may let the current process continue, but subsequent launch/resume and backup staging cannot reopen that dependency; the missing-base detector will route the next pause back through recovery. Promote/pin the base into lifetime-managed storage and persist that stable path before deleting staging. Cover fallback -> cleanup -> pause/backup -> resume for an overlay.

internal/vm/grpc_adapter.go
```diff
@@ -123,7 +123,14 @@ func (a *GRPCAdapter) ResumeVM(ctx context.Context, req *vmdpb.ResumeVMRequest)
 		}
 	}
 
-	inst, rulesApplied, err := a.mgr.resumeVMLocked(ctx, req.GetVmId(), req.GetSnapshotPath(), req.GetMemFilePath(), resumeNetworkRules)
+	var inst *VMInstance
+	var rulesApplied bool
+	if a.mgr.backupReader != nil && a.mgr.pauseArtifactsMissing(req.GetVmId(), req.GetSnapshotPath(), req.GetMemFilePath()) {
+		inst, err = a.mgr.resumeFromBackupLocked(ctx, req.GetVmId(), resumeNetworkRules)
```
pavitrabhalla: [P1] Make a successful fallback adoptable by a retried ResumeVM

After a cold revival, the new instance is Running and has no SnapshotPath/MemFilePath (the revival seed preserves ownership/policy, not the old pause paths). If the response is lost, an identical ResumeVM retry bypasses this branch because `pauseArtifactsMissing` only accepts Paused, then enters `resumeVMLocked`. Empty request paths now fail required-path validation; explicit original paths cannot match `retriedLaunchTarget` and subsequently fail the missing-artifact checks. A successful recovery thus strands the caller behind errors rather than returning the already-running instance. Persist an identity for the recovery/resume attempt and adopt its live result before ordinary snapshot validation; do not merely copy obsolete paths and risk restoring the old snapshot. Add a lost-response retry test using the original request.

internal/vm/resume_backup.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return false
+	}
+	inst.mu.RLock()
+	defer inst.mu.RUnlock()
+	if inst.Status != StatusPaused {
+		return false
+	}
+	disk := pausedDiskPath(m.cfg.RunDir, vmID, inst.DiskPath, inst.RunDirID, inst.Config.BasePath)
+	return !pauseArtifactsPresent(snapshotPath, memPath, inst.BaseMemPath, disk, inst.Config.BasePath)
```
pavitrabhalla, line 66: [P2] Resolve omitted request paths before testing artifact presence

`resumeVMLocked` supports empty snapshot/memory request paths by defaulting them to the instance fields. This predicate tests the raw request instead, so with backups enabled a healthy paused VM resumed by ID alone is always classified as missing artifacts. It either fails because no backup exists or downloads a backup only for revival to refuse the healthy paused VM; the previously valid warm resume never runs. Apply the same path normalization before routing, and test a Paused instance with recorded existing paths and an empty-path request (including the one-path-omitted variants).

pavitrabhalla (changes requested):
Request changes on 7d98be1. The reuse of the verified restore machinery and the locked revival lifecycle is a reasonable direction for disk-only disaster recovery, but it does not yet safely implement ordinary resume: the generation is not tied to the current pause, the overlay base is deleted on success, and retry/default-path semantics regress. Cold boot also intentionally loses in-memory processes/state, so this is recovery rather than transparent continuation.

Validation: inspected the changed code and surrounding restore/revive/resume, base-cache and migration paths. All 11 reported GitHub check runs are successful, including the eight required checks. I could not execute local Go tests: Go was absent and installing the toolchain exhausted the review environment's disk space. The added tests exercise artifact detection and fetch results, not the fallback lifecycle and failure sequences identified below. No merge performed.

Before shipping, add the lifecycle regressions below and measure the inline download/hash/cold-boot latency against the resume RPC deadline; this path holds the per-VM operation lock throughout.

db/queries/sandboxes.sql
```diff
@@ -566,11 +566,17 @@ FROM (
...
+         COALESCE(bg.generation, '')::text AS backup_generation
   FROM sandbox sb
   LEFT JOIN snapshot s ON s.id = sb.snapshot_id AND s.team_id = sb.team_id
   LEFT JOIN sandbox_preview_policy p ON p.sandbox_id = sb.id
   LEFT JOIN template t ON t.id = sb.template_id
+  LEFT JOIN LATERAL (
+    SELECT generation FROM backup_generation
+    WHERE covered_snapshot_id = s.id
+    ORDER BY completed_at DESC LIMIT 1
+  ) bg ON true
```
awille-ai: Critical: this is still not an exact-pause lookup. `snapshot` is intentionally reused across pauses and its `generation` counter advances; the coverage migration therefore stores both `covered_snapshot_id` *and* `covered_snapshot_generation`. Filtering only on `covered_snapshot_id = s.id` allows a generation from a previous pause on the same snapshot row to match. If the current pause's upload is incomplete and local artifacts are missing, this can pass an older backup generation to VMD and cold-boot stale customer data successfully. Match both identities (at minimum `covered_snapshot_id = s.id AND covered_snapshot_generation = s.generation`) and add the two-pause/incomplete-newer-generation regression at the control-plane claim boundary.

db/queries/sandboxes.sql
```diff
@@ -566,11 +566,17 @@ FROM (
...
   LEFT JOIN LATERAL (
```
awille-ai, line 574: High (scaling): this lookup has no supporting index for its predicate/order. Existing backup indexes are `(sandbox_id, completed_at DESC)` / uniqueness by sandbox+bucket+generation; none start with `covered_snapshot_id` (and after fixing the correctness issue, `covered_snapshot_generation`). `ClaimResume` is the resume hot path, so as `backup_generation` grows this can become a table scan + sort per resume. Add an index aligned with the exact-pause lookup, e.g. on `(covered_snapshot_id, covered_snapshot_generation, completed_at DESC)` (optionally including `generation`), and verify the plan.

infra/modules/backup-storage/main.tf
```diff
@@ -77,6 +77,16 @@ resource "google_storage_bucket_iam_member" "writer_create" {
   member = each.value
 }
 
+# A runtime that writes pause backups reads them back when a resume finds
+# the pause artifacts gone from the host.
+resource "google_storage_bucket_iam_member" "writer_view" {
+  for_each = toset(var.writer_members)
+
+  bucket = google_storage_bucket.backup.name
+  role   = "roles/storage.objectViewer"
```
awille-ai: Critical security boundary regression: `writer_members` is explicitly the legacy/shared write-only identity set; the module README says those identities must not read because they are shared across cells. This grant gives every `writer_member` `objectViewer`, which includes read + list of the whole cell bucket. Current envs still pass shared API/runtime identities here, while dedicated per-cell VMD identities already receive create+viewer via environment-owned `vmd_backup` grants. A compromise of a shared runtime can therefore enumerate/exfiltrate backup contents beyond its intended cell. Keep `writer_members` create-only and grant viewer only to the dedicated per-cell VMD SA.

internal/vm/resume_backup.go
```diff
@@ -0,0 +1,285 @@
...
+		})
+		if f.err == nil {
+			f.err = m.promoteBase(&f.restored, dest)
+		}
+		if m.backupRestore.CacheBytes > 0 {
+			if _, perr := backup.PruneBaseCache(m.backupBaseDir(), m.backupRestore.CacheBytes); perr != nil {
+				log.Warn().Err(perr).Msg("backup base cache prune")
+			}
+		}
+		m.sweepPromotedBases()
```
awille-ai: High (resource exhaustion): fetch concurrency only bounds active downloads. Once a caller has timed out, a successful fetch releases `backupFetchSem` and then retains the full per-VM restored disk in staging for `AbandonAfter` (default 1h) while waiting for a claim. New restores can continue, so a resume storm with clients that do not retry promptly can accumulate an unbounded number of full sandbox disks under `SnapshotDir/.restore`; `CacheBytes` only bounds the base cache. Put a host-wide byte/count budget on completed-unclaimed staging (or reclaim it much sooner / make retention explicitly bounded) so the recovery path cannot fill the snapshot disk while respecting the nominal concurrency limit.

awille-ai (changes requested):
Security/scaling review of the current head. The direction is workable, but I would not ship this version yet. The biggest issue is that the control plane still does not bind the restore to the exact pause generation under the existing reused-snapshot-row model, so an older filesystem can still be restored successfully. Separately, this broadens the legacy/shared writer identities to bucket read/list despite the module's documented isolation contract, and the restore staging lifetime is not bounded by the concurrency/cache controls. CI and Terraform plan workflows are green on this head, but those checks do not make these invariants safe.

meAmitPatil:
@greptileai

meAmitPatil:
@codex review

meAmitPatil:
@greptileai

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

meAmitPatil:
@codex review

internal/vm/resume_backup.go
```diff
@@ -0,0 +1,458 @@
...
+// fail on a file the host no longer has. Omitted request paths resolve
+// from the record, as the resume itself resolves them.
+func (m *Manager) pauseArtifactsMissing(vmID, snapshotPath, memPath string) bool {
+	inst, err := m.getInstance(vmID)
+	if err != nil {
+		return false
+	}
+	inst.mu.RLock()
+	defer inst.mu.RUnlock()
+	if inst.Status != StatusPaused {
```
pavitrabhalla: [P1] Keep a failed backup boot eligible for the next ResumeVM attempt. `coldBootFromRootfs` leaves an in-memory StatusError instance on copy/network/launch failure. `reviveVMLocked` restores the original paused record to the state store, but not that in-memory instance. Consequently this guard rejects backup recovery on the next request, and the adapter falls through to ordinary resume against the same missing snapshot. I reproduced this with a successful mocked fetch followed by a real rootfs-copy failure: durable status=paused, memory status=error; after removing the injected filesystem failure, retry returns FailedPrecondition (`snapshot file missing on host`) and the fetch count remains one. Restore the in-memory retry anchor as well, or safely recognize this specific failed backup-revival state; don't broadly allow arbitrary error VMs. Add a regression that drives a boot failure through ResumeVM and successfully re-enters backup recovery on retry.

internal/vm/resume_backup.go
```diff
@@ -0,0 +1,458 @@
...
+			continue
+		}
+		f.state = flightDropped
+		delete(m.backupFlights, vmID)
+		f.drop()
+		dests = append(dests, m.restoreStagingDir(vmID))
+	}
+	m.backupFlightsMu.Unlock()
+	for _, d := range dests {
+		_ = os.RemoveAll(d)
```
pavitrabhalla: [P2] Do not release a staging path for reuse before deleting its retired flight. After deleting the map entry and releasing backupFlightsMu, a concurrent retry can register a new flight for this VM and start writing to the same `staging/<vmID>` directory before this RemoveAll executes. The old flight's eviction then deletes the new flight's files, potentially even a newly completed/claimed restore. The state lock prevents claiming the old flight but does not give the filesystem directory the same ownership guarantee. Use a unique directory per flight, atomically rename the retired directory before permitting a replacement, or keep a dropping tombstone until cleanup completes; add a test that interleaves eviction with a retry for the same VM.

pavitrabhalla (changes requested):
Re-reviewed 2511fab. The exact-pause generation lookup, default-path resolution, base promotion, bounded shared fetches, cold-boot attestation, and startup-sweep guards address substantial parts of the earlier feedback. Two retry/lifecycle issues remain below.

Validation: all 11 GitHub check runs report success. Independently ran the existing internal/backup, internal/vmdclient, internal/vm, internal/api, and cmd/backup-restore tests successfully (sequential reruns after an initial parallel run exhausted the temporary environment's disk). An additional focused regression test reproduces the failed-boot retry bug: after an injected rootfs-copy failure, the durable record is paused but the in-memory instance is error; after removing the failure, ResumeVM returns `snapshot file missing on host` without fetching again. The eviction interleaving is a code-inspection finding, not a reproduced race. No merge performed.

internal/vm/resume_backup.go
```diff
@@ -0,0 +1,516 @@
...
+	}
+	m.mu.Lock()
+	inst := m.vms[vmID]
+	failed := false
+	if inst != nil {
+		inst.mu.RLock()
+		failed = inst.Status == StatusError
+		inst.mu.RUnlock()
+	}
+	if failed {
```
pavitrabhalla: [P1] Restore the retry anchor when teardown removed the instance, too. This repairs copy/network failures that leave StatusError in m.vms, but a boxd readiness failure takes a different path: reviveVMLocked calls teardownSelf/DestroyVM, which removes m.vms[vmID], then its deferred rollback restores the old paused record with RevivalPending=true. Here inst is nil, so `failed` remains false and the anchor is never restored. On the next ResumeVM, getInstance tries lazy reattachment while ResumeVM already holds the lifecycle lock; reattachRecord refuses the pending-revival record because tryLockVMOp fails, yielding NotFound before any backup fetch. A state-level regression reproduces this with a paused+RevivalPending durable record and an empty VM map: restorePausedAnchor followed by ResumeVM returns NotFound with zero backup reads. Handle the confirmed-teardown/no-instance case as well, preserving the existing destroy-wins and unconfirmed-stop safety rules, and add a regression for readiness timeout followed by retry rather than only rootfs-copy failure.

pavitrabhalla (changes requested):
Re-reviewed fc0c895 against 2511fab. The staging eviction now renames the retired directory under the flights lock, including timer-based cleanup, which addresses the earlier path-reuse race. The copy-failure retry regression also passes. However, the retry-anchor repair misses the other failure outcome: a completed teardown removes the in-memory instance entirely (inline finding below).

Validation: the existing internal/backup, internal/vmdclient, internal/vm, internal/api, and cmd/backup-restore suites pass locally; the latter two required a separate rerun after temporary disk exhaustion during linking. A temporary state-level regression seeded the exact post-teardown rollback state (paused durable record with RevivalPending=true, no in-memory instance), called restorePausedAnchor, then ResumeVM: it returned NotFound without reading the backup. This did not launch a real Firecracker guest.

CI currently reports 10 successful checks and a failing required Integration Tests check. The failing test is TestIntegration_IncrementalWorkerLoad/measurement-cursors, with `query scanned more than 48 rows` at billing_incremental_load_test.go:546 (sequential scans on billing_export_measurement_queue). I have not established that this billing failure is caused by this PR; it still needs resolution or a verified clean rerun. No merge performed.

internal/vm/resume_backup.go
```diff
@@ -0,0 +1,519 @@
...
+		return
+	}
+	if _, destroying := m.destroying.Load(vmID); destroying {
+		return
+	}
+	rec, err := m.state.Get(vmID)
+	if err != nil || rec == nil || rec.Status != StatusPaused {
+		return
+	}
+	m.mu.Lock()
```
pavitrabhalla: [P1] Serialize the anchor read and publication with DestroyVM. The one-time destroying check does not protect the new absent-instance path: this function can read the paused record, an external DestroyVM can then finish deleting both the record and map entry and clear its tombstone, and this helper subsequently inserts toInstance(*rec) back into m.vms and pressureIndex. DestroyVM intentionally bypasses the lifecycle lock held by ResumeVM, so that lock does not prevent the interleaving; m.mu protects the map mutation but not the lifetime of the previously read record. A controlled state-level regression reproduces the stale map entry after the destroy's effects have completed. Hold lockRecordOwner(vmID) across the durable read and in-memory publication (following reviveVMLocked's rollback discipline), or use an equivalent atomic ownership/epoch check, and add a destroy-wins regression for the nil-instance branch.

internal/integration/integration_test.go
```diff
@@ -2020,19 +2020,25 @@ func TestIntegration_GetBillingSummaryUsesCommercialBillingAnchor(t *testing.T)
 	if !ok {
 		t.Fatalf("billing_period not an object: %v", body["billing_period"])
 	}
+	// The summary reports the anniversary period that contains now, which
+	// rolls forward a month at a time from the anchor.
+	wantStart, wantEnd, anchored := billing.AnniversaryPeriod(anchor, time.Now())
```
pavitrabhalla: [P2] Compute the expectation using the router's injected clock. This test now constructs newRouterWithNow at 2026-09-04, but the changed expectation still uses wall-clock time.Now(). The handler therefore correctly returns the Aug 21–Sep 21 anniversary period while the assertion advances to Sep 21–Oct 21 on the current date. The current head's Integration Tests job fails at line 2034 with exactly that mismatch. Bind the fixed test time once and use it for both the router and AnniversaryPeriod, or retain the fixed anchor/anchor.AddDate expectation for this fixture; the assertion should not depend on when CI runs.

pavitrabhalla (changes requested):
Re-reviewed 32c6c48, including the updated base. The previously reported post-teardown NotFound case is fixed, and its added regression passes. Two remaining issues prevent approval: restoring an absent instance now needs synchronization against a concurrent authoritative destroy, and the billing assertion in this PR uses a different clock from the updated test router.

Validation: the existing internal/backup, internal/vmdclient, internal/vm, internal/api, and cmd/backup-restore suites all pass locally on this head. A separate temporary state-level concurrency regression paused restorePausedAnchor after reading the durable record, modeled completion of the destroy's map/index/record deletion and epoch advancement, then allowed publication to proceed: the stale in-memory VM was republished. This models the interleaving; it is not an end-to-end Firecracker teardown test.

CI: 10 checks succeeded; the required Integration Tests check fails TestIntegration_GetBillingSummaryUsesCommercialBillingAnchor with actual start 2026-08-21 versus expected 2026-09-21. This is a different failure from the previous billing query-scan failure and is explained by the inline clock mismatch below. No merge performed.

pavitrabhalla (approved):
Re-reviewed ba72fd6. Both remaining findings are addressed: restorePausedAnchor now holds the record-owner lock across the durable read and map/index publication, preserving destroy-wins ordering, and the billing expectation uses the same fixed clock as the router. The earlier copy-failure, completed-teardown retry, and staging-eviction fixes remain intact. No remaining blocking findings from this review.

Validation on this exact head: internal/backup, internal/vmdclient, internal/vm, internal/api, and cmd/backup-restore all pass locally. Repeated the four focused copy-failure, post-teardown, staging-eviction, and destroy-wins regressions 20 times successfully. All 11 GitHub checks are green, including every required check and Integration Tests. This review did not execute a real Firecracker/GCS disaster-recovery drill locally.

The documented concurrent index pre-build remains a rollout prerequisite for a populated database. Approved; no merge performed.

awille-ai (approved):
Reviewed current head `ba72fd6` in two independent passes: first manually without the review skill, then with the PR-review workflow.

**Verdict: APPROVE.** I don't see a remaining Critical or High issue.

The manual and structured passes both rechecked the failure surfaces that matter most for this change: exact-pause generation selection, missing-artifact routing, lost-response adoption, RPC deadline/retry behavior, failed cold-boot recovery, post-teardown retry, destroy-vs-retry ordering, staging eviction/path reuse, startup residue cleanup, promoted overlay-base lifetime, cache/flight/bandwidth bounds, and cold-boot secret reinjection. The current implementation is coherent across those paths.

| Reviewer | Result |
| --- | --- |
| Hostile/adversarial | PASS |
| Security/resilience | PASS |
| Test quality | PASS |
| Correctness | PASS |
| Performance/resource efficiency | PASS |
| Observability | PASS |
| Compatibility/rollout | PASS |
| Data correctness | PASS |
| Sandbox lifecycle performance | PASS |
| Migration/infrastructure safety | PASS, with rollout prerequisite below |

Two non-blocking caveats:
- On a populated database, `idx_backup_generation_covered_pause` must be pre-built **CONCURRENTLY** before the migration lands, exactly as the migration documents. The checked-in plain `CREATE INDEX` is intentionally bounded by short timeouts and should fail rather than block backup-report writes if that prerequisite is missed.
- I did not execute a real Firecracker + GCS disaster-recovery drill; this review relies on the code paths, regression tests, and current CI for those integration boundaries.

No merge performed.

PR #542, opened 2026-09-18
controlplane: one statement for the resume's post-boot policy read and owner capability check

meAmitPatil:
After the boot, a resume read the policy revision and then ran the owner capability check as its own locked transaction. The two reads are now one statement, sent with the host lock as one pipelined batch, so it still follows a heartbeat that is withdrawing a capability but costs one round trip instead of five.

The rare paths (stateless fallback, version skew, a revision that moved during the boot) keep the full read and validation.


meAmitPatil:
@codex review

meAmitPatil:
@codex review

pavitrabhalla (approved):
Reviewed commit 24ea9beb0597621c0c1c7b30b757e27d75fff165. No blocking findings.

The latest revision restores the lock-then-read ordering: strict policies use a pgx batch containing the host FOR SHARE lock followed by the policy/capability query, with the batch closed before returning. Reviewed the pgx pipeline implementation and the production CacheDescribe configuration; execution preserves the two-statement transaction boundary needed for a fresh READ COMMITTED snapshot after the heartbeat writer. The earlier unlocked-read finding is addressed in this revision.

Checked capability-set selection, active/draining status and heartbeat freshness, tenant/destroyed-row filtering, legacy-policy compatibility, and the full revalidation paths for newer policy revisions, stateless fallback, and version skew.

Validation: `go test -count=1 ./internal/api` and `go test -count=1 ./internal/db` passed locally. All 11 reported GitHub checks pass on this head, including Integration Tests and sqlc Check.

Non-blocking follow-up: add a real PostgreSQL overlap test for LockedResumePostBootCheck using the production CacheDescribe mode, covering both a stable capability refresh and withdrawal. The new handler tests model batch ordering, while the existing database overlap tests exercise LockedHostHasCapabilities rather than this new wrapper. I did not execute a new live-database overlap test locally.

Reviewed by QM using the authorized reviewer account.

PR #543, opened 2026-09-19
Raise template disk ceiling to 64 GiB

pavitrabhalla:
Raises the platform template disk ceiling from 30 GiB to 64 GiB (65,536 MiB). Updates API validation, the OpenAPI schema, and the database constraint via a new migration.

CPU/RAM limits, default sizes, and per-team quotas are unchanged. This does not resize existing templates or sandboxes.

Apply the migration before deploying the API change. Non-system teams still need an explicit disk quota increase to use larger templates.

Tested: API unit tests and template integration tests against local PostgreSQL 16, including 32/64 GiB acceptance, boundary rejection, and unchanged team quotas. No live VM provisioning performed; full CI remains to run.


pavitrabhalla:
@codex review

awille-ai (approved):
Reviewed all five changed files and the downstream builder sizing path. No blocking findings: API validation, OpenAPI and the database constraint agree on 65,536 MiB; the platform ceiling still applies before the system-team bypass, and customer defaults and explicit team limits remain enforced. Builder byte-size calculations widen to int64 before conversion.

All 11 current CI checks pass, including unit tests, PostgreSQL integration tests and migration ordering. The added tests cover boundary rejection, 32/64 GiB acceptance, persisted sizes and unchanged quotas. I relied on CI for execution; local tests were not completed because the review environment ran out of disk space. No live VM provisioning was performed.

Keep the documented rollout order: apply the migration before deploying the API change.

